PR #168, opened 2011-05-26
Support for non-Django template systems

nyergler:
Django 1.2 introduced support for non-Django template systems. Unfortunately the instrumented test renderer is still Django specific. This branch updates the Template panel monkey patching to wrap template lookup and decorate the Template class' render method to emit the correct signal. This allows DjDT to work with non-Django template systems (tested with Mako templates).


aaugustin:
I'm going to close this in favor of #693. Sorry it didn't get merged.

PR #323, opened 2012-09-17
Add EXPLAIN ANALYZE for Postgres

mattrobenolt:


matthewcornell:
Hi folks. Thanks for the great tool, which I've just started using. I'm developing locally with both sqlite3 and postgres, and while sqlite3 has `Sel` and `Expl` buttons next to all queries, postgres is missing some (see pics). Versions: Django 1.11, debug_toolbar 1.9.1, postgres 9.6.5, sqlite3 3.8.10.2. Sorry if I'm missing something basic here. Let me know if you need more info. Thanks!

postgres-not-all-buttons
sqlite3-yes-all-buttons

Issue #307, opened 2012-07-30
Incompatible with a Content Security Policy


graingert:
Because debug_toolbar uses in-line JS it cannot function with a Content Security Policy active.

While disabling CSP temporarily is a possible, it slows down development.

PR #330, opened 2012-10-09
Added support for Content Security Policy

gavinwahl:
In order to use debug toolbar with a content security policy enabled, no
inline scripts or styles can be used. This commit replaces all `style`
attributes with CSS or javascript. It uses `data` attributes to handle
the dynamic styles, like width and color.

In addition, this required updating jQuery to 1.8, because 1.7 is not
CSP compatible.

See also #307, which began implementing CSP compatibility.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/print.css
+++ b/debug_toolbar/static/debug_toolbar/css/print.css
@@ -1,0 +1,3 @@
+#djDebug {
+  display:none;
+}
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -646,3 +646,16 @@
         display: none;
     }
 }
+
+#djDebug .width-20 {
+    width: 20%;
+}
+#djDebug .width-60 {
+    width: 60%;
+}
+#djDebug .width-60 {
+    width: 60%;
+}
+#djDebug .highlighted {
+    background-color: lightgrey;
+}
--- a/debug_toolbar/static/debug_toolbar/js/jquery_existing.js
+++ b/debug_toolbar/static/debug_toolbar/js/jquery_existing.js
@@ -1,0 +1,1 @@
+var djdt = {jQuery: jQuery};
--- a/debug_toolbar/static/debug_toolbar/js/jquery_post.js
+++ b/debug_toolbar/static/debug_toolbar/js/jquery_post.js
@@ -1,0 +1,1 @@
+var djdt = {jQuery: jQuery.noConflict(true)}; window.define = _djdt_define_backup;
--- a/debug_toolbar/static/debug_toolbar/js/jquery_pre.js
+++ b/debug_toolbar/static/debug_toolbar/js/jquery_pre.js
@@ -1,0 +1,1 @@
+var _djdt_define_backup = window.define; window.define = undefined;
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -285,13 +285,21 @@
 
                 return value;
             }
+        },
+        applyStyle: function(name) {
+            $('#djDebug [data-' + name + ']').each(function() {
+                var css = {};
+                css[name] = $(this).data(name);
+                $(this).css(css);
+            });
         }
     };
     $.extend(publicAPI, {
         show_toolbar: djdt.show_toolbar,
         hide_toolbar: djdt.hide_toolbar,
         close: djdt.close,
-        cookie: djdt.cookie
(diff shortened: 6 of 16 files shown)


danpalmer:
Can someone take a look at this? It would be really useful for the project to be as conservative as possible in supported CSPs, so that development environment can differ as little as possible from production.

gavinwahl:
I started working on this again and updated the pull request. I have it working with RENDER_PANELS=True, but with RENDER_PANELS=False it's tricky.

When rendering panels lazily, debug toolbar fetches their content over xhr and then inserts them with $.html. When you do $.html('<script src="http://foo"></script>') in jquery, it doesn't actually insert a script tag into the dom -- it retrieves the script source with xhr and then evals it (see _evalUrl), which of course doesn't work with CSP. I'm not sure how to fix this.

graingert:
@gavinwahl nice one, thanks for taking another look at this 2yo PR!

What is the Google api Ajax URL for, and why is it not https?

gavinwahl:
It's for debug toolbar to load it's jquery (JQUERY_URL). It should actually be `//ajax.googleapis.com/` to allow https.

danpalmer:
Unless we want to bundle jQuery (which I'm pretty sure we don't!) then we will have to have that exception, so I suggest we document it.

With regards to the other parts, allowing `self` makes sense, and is pretty reasonable, but it would be nice to remove the dependency on `img-src data:;`. Is that possible to remove?

graingert:
@gavinwahl I'm right in thinking that JQUERY_URL can point to a self hosted jQuery JS.

gavinwahl:
data: is used for some icons: https://github.com/django-debug-toolbar/django-debug-toolbar/blob/master/debug_toolbar/static/debug_toolbar/css/toolbar.css#L150. I suppose they could be put into files instead. I'm not sure if there's any security problem with allow data urls for images.

Regarding the rest of the policy, this is not the purview of debug toolbar. It doesn't set the header itself, that's up to you and you can use whatever policy you like. If you set JQUERY_URL to something local, feel free to remove ajax.googleapis.com from script-src.

I only mentioned the policy I used so we could document a policy that works with the default toolbar config.

graingert:
@gavinwahl With data-uri images you can place a transparant gif over the top of your bank balance and trick a user into sending the incorrect amount to an attacker.

They can also be slower than multiple files in lots of places.

Please put these in separate files if you have time.

graingert:
Some of these might also look better as Unicode glyphs

aaugustin:
Since the debug toolbar is intended for local use, it shouldn't be optimized for speed, only for ease of maintenance and development.

gavinwahl:
I opened jquery/jquery#2012 for the script thing

gavinwahl:
Looks like the jquery thing isn't going to be fixed until jquery 3.0 is released. I'm not aware of a workaround, so we'll just have to document that you must either use RENDER_PANELS=False or jquery>=3.0. What section of the docs can these notes about CSP support go?

graingert:
@gavinwahl you can get this working in jQuery 2.x by triggering:

```javascript
if ( code.indexOf("use strict") === 1 ) {
    script = document.createElement("script");
    script.text = code;
    document.head.appendChild( script ).parentNode.removeChild( script );
} else {
```

gavinwahl:
No, I saw that too and tried it. Notice that it's putting the source code of the script in script.text -- this is an inline script, which CSP blocks.

graingert:
urgh yeah I didn't spot that.

graingert:
@gavinwahl can you use:

```javascript
$(selector)[0].innerHTML = htmlString;
```

gavinwahl:
Script tags inserted with innerHTML are not executed. That's the reason why jQuery is doing this stuff in the first place.

graingert:
Can we load the scripts statically and the content dynamically?

gavinwahl:
Yes, we could load the scripts for all the builtin panels in base.html, where toolbar.js is loaded. Third-party panels would be unable to load scripts then.

graingert:
How about getting a copy of the jQuery 3.0 .html function and including it
here?

gavinwahl:
That doesn't seem feasible, it's actually functions that are indirectly used by .html(). We'd have to monkey patch jquery's _evalUrl function to do it, but we reuse a jquery that's already on the page. It'd be rude to go about monkey patching some else's copy of jquery.

gavinwahl:
I don't think there's going to be any workaround for the jquery thing, so we'll just have to document that you must either use RENDER_PANELS=False or jquery>=3.0 for CSP compatibility. What section of the docs can these notes about CSP support go?

graingert:
I've rebased and fixed the tests in #704

tim-schilling:
Closing this since #704 was merged in.

PR #563, opened 2014-03-16
Disable Logging panel when DEBUG is set to False to prevent memory leaks

stanhu:
I upgraded django-debug-toolbar from 0.9.4. to 1.0.1, and I noticed my application was quickly leaking memory even when settings.DEBUG was set to False. Once I removed debug_toolbar from the list of applications, the problem disappeared.

Upon further investigation, I found that LoggingPanel was quietly collecting all my log messages and never releasing them.

This problem only occurs if the Django application is fully initialized (e.g. using runserver), since this causes the LoggingPanel to register its handlers. Here is a simple application that illustrates the problem:

``` python
from django.core.management import ManagementUtility
import logging

utility = ManagementUtility()
command = utility.fetch_command('runserver')
command.validate()

logger = logging.getLogger('test')

while True:
    logger.info('This logger will leak memory')
```

Run this application with the default debug_toolbar settings, and then use 'top' to watch the memory grow without bound. The problem appears that the LoggingPanel does not use the enable_instrumentation() function normally called by the middleware, which does the DEBUG check beforehand.

I think the simplest fix is to check if DEBUG is enabled before initializing the LogCollector. This pull request does just that, and it appears to fix the issue.

In addition, I think we should consider a few other things:

1) Warn the user that under DEBUG mode, memory used for log messages can grow without bounds (this might be true for other panels already).

2) Limit the number/age of messages stored (perhaps as a configuration option).

### DIFF
--- a/debug_toolbar/panels/logging.py
+++ b/debug_toolbar/panels/logging.py
@@ -16,12 +16,32 @@ MESSAGE_IF_STRING_REPRESENTATION_INVALID = '[Could not get log message]'
 
 class LogCollector(ThreadCollector):
 
+    def __init__(self):
+        ThreadCollector.__init__(self)
+        self.enabled = False
+
     def collect(self, item, thread=None):
         # Avoid logging SQL queries since they are already in the SQL panel
         # TODO: Make this check whether SQL panel is enabled
         if item.get('channel', '') == 'django.db.backends':
             return
         super(LogCollector, self).collect(item, thread)
+
+    def enable_logging(self):
+        """
+        Enable logging if it has not already been enabled and DEBUG is True.
+        """
+        if self.enabled:
+            return
+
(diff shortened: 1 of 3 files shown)


camilonova:
@aaugustin can you please review the changes?

debug_toolbar/panels/logging.py
```diff
@@ -15,13 +16,33 @@
 
 class LogCollector(ThreadCollector):
 
+    def __init__(self):
+        ThreadCollector.__init__(self)
```
aaugustin: Is there a particular reason for using `ThreadCollecter` instead of `super()`? If so, I think it should be documented.

matthiask:
Closing because of inactivity. Please resubmit if you still want to work on this.

PR #615, opened 2014-06-25
panels.sql.tracking.wrap_cursor: pass through arguments to original cursor.

keturn:
Fixes https://github.com/freelancersunion/django-salesforce/issues/23


aaugustin:
Let's just do it with a comment saying it's gross....

matthiask:
OK. I merged https://github.com/jazzband/django-debug-toolbar/pull/896 and added a short explanation.

PR #637, opened 2014-09-06
Adding Template Panel support for other template engines.

tim-schilling:
This is another take on the concept of supporting non-Django template engines. However, this will only support Django 1.6+ as it makes use of ```import_by_path``.

The solution makes use of decorators. I got the idea from @nyergler's PR [#168](https://github.com/django-debug-toolbar/django-debug-toolbar/pull/168).


aaugustin:
Related: #693.

tim-schilling:
@aaugustin Could you read my last comment and give me your opinion? Either way, I'm going to have to take a another look at this PR to make sure 1.8 will work. However I'd like to know what the standard is for building out features, do they have to work for every version, or can they fail gracefully on older versions?

aaugustin:
I find it acceptable to provide more features on more recent Django versions.

Note that the current patch uses `settings.TEMPLATE_LOADERS` which is deprecated in Django 1.8.

tim-schilling:
Cool, I wanted to be on the same page in term of Django versions and functionality.

I'm going to close this as it's not needed with #711 being merged in.

aaugustin:
Oops - I didn't notice the overlap. If you think I merged #711 too quickly let's talk! The patch was small and made sense, so I went ahead, but I didn't check the big picture.

tim-schilling:
No, not at all. I was going to merge it in myself. I think my solution with this PR is a little too clever monkey patching things on the fly and would get the same result as #711.

PR #667, opened 2014-11-23
Adjust Django debug toolbar to do non-ASCII-safe string operations

onceuponatimeforever:
Credit goes to Asheesh Laroia for writing this patch to Open Hatch code
base. I am doing the upstream part. Related commit is here: https://github.com/openhatch/oh-mainline/commit/e641408564fb9dde26083d510fb0a9ac591a1251

### DIFF
--- a/debug_toolbar/utils.py
+++ b/debug_toolbar/utils.py
@@ -35,7 +35,20 @@ def get_module_path(module_name):
         raise ImproperlyConfigured(
             'Error importing HIDE_IN_STACKTRACES: %s' % (e,))
     else:
-        source_path = inspect.getsourcefile(module)
+        # Note that source_path is a byte string in the default
+        # encoding for the filesystem, which means that
+        # calling .endswith() will attempt to convert it to
+        # ASCII.
+        #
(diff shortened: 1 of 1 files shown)


debug_toolbar/utils.py
```diff
@@ -35,7 +35,20 @@ def get_module_path(module_name):
...
+        # encoding for the filesystem, which means that
+        # calling .endswith() will attempt to convert it to
+        # ASCII.
+        #
+        # os.path.dirname() is OK dealing with byte strings
+        # in unknown encodings, but .endswith() is not, so
+        # here we cast it down to ASCII for the __init__.py
+        # check, since all we care about is if we need to
+        # call os.path.dirname() on it.
+        smashed_to_ascii = source_path.decode('ascii', 'replace').encode(
```
Zopieux: `source_path` is undefined, the line that defines it was removed by this commit (l. 38)!

tim-schilling:
Closing this PR as this change won't work with python 3 after the fix to the current build issue. It also doesn't account for the other places in the toolbar where unicode filepaths may cause problems.

PR #673, opened 2014-12-29
Improved the build process

thedrow:
The build will now test against the latest patch of each major python version.
I added Django 1.7 to the build.
Travis will use the new docker based workers since the build doesn't require sudo. They boot faster.


tim-schilling:
Thanks for the PR!

PR #674, opened 2014-12-29
Fixed Django1.8 support (template panel still requires a Django engine t...

apollo13:
...o be defined)

### DIFF
--- a/debug_toolbar/panels/templates/panel.py
+++ b/debug_toolbar/panels/templates/panel.py
@@ -13,7 +13,6 @@ from django.conf import settings
 from django.conf.urls import url
 from django.db.models.query import QuerySet, RawQuerySet
 from django.template import Context, RequestContext, Template
-from django.template.context import get_standard_processors
 from django.test.signals import template_rendered
 from django.test.utils import instrumented_test_render
 from django.utils.encoding import force_text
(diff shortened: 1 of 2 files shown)


orokusaki:
Will this pull request be merged when 1.8b1 is released?

hanleyhansen:
Anyone know when this will go in?

tim-schilling:
@aaugustin What do you think of creating a specific file to handle the different imports/functions based on the versions of Django? It should help restrict all the if/else logic into a single file and keep the rest of the project ignorant of it.

aaugustin:
@tim-schilling Yes, something like `debug_toolbar.compat` would be useful, especially when removing support for older Django versions.

aaugustin:
@tim-schilling Do you intend to work on this within the next few days?

tim-schilling:
I might not get to it until late this week or this weekend, but yes.

aaugustin:
I won't have time before the weekend either.

It would be good to upload a 1.8-compatible release as soon as possible. I suggest the first of us who has time writes the patch and the other reviews it :-)

aaugustin:
As of 22473d36346901b3320f1bf25f4f602d8ded6d04 tests pass on Django 1.8 beta 1.

PR #678, opened 2015-01-06
Fix jQuery events priority

### LINKED ISSUE #677: jQuery click events handlers priority conflicts with Angular.js
Angular.js listens for click events on $('html') element for navigation purpose. I don't know why they use such approach, but it results in conflict with dj-debug-toolbar.

toolbar.js adds delegated handlers to $(document), but angular.js handles this events in the first instance, thus every click in debug-toolbar leads to page reloading.

Handlers bound to $('html') are executed before $(document)

nailgun:
Fix #677

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -8,7 +8,7 @@
         init: function() {
             $('#djDebug').show();
             var current = null;
-            $(document).on('click', '#djDebugPanelList li a', function() {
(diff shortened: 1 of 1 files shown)


tim-schilling:
There looks to be a number of other event handlers bound to ```document``` in toolbar.js. Lines 138, 157, 171, and 233. Do these also need to be updated?

tim-schilling:
@nailgun can you let me know if those lines also need to be updated?

nailgun:
@tim-schilling No. These are global namespaced events, they don't cause any problems.

tim-schilling:
Thank you for the PR!

PR #684, opened 2015-01-29
Limit handle position to window visible area

marciomazza:
We should limit the handle top position property.
It is altered by user interaction, cookie saved and restored.
That can can make it too high for the current visible area.


marciomazza:
We have to subtract something a little greater than the height of the button (75px) for it to be fully visible. See:

https://github.com/django-debug-toolbar/django-debug-toolbar/blob/master/debug_toolbar/static/debug_toolbar/css/toolbar.css#L179

    #djDebug #djShowToolBarButton {
        ...
        height:75px;
        ...
    }

PR #689, opened 2015-02-20
Loading the `url` tag from the `future` library is deprecated (this will deprecate Django 1.4 support).

renstrom:
Using the `url` tag directly has been supported for quite a while.

```
django/templatetags/future.py:24: RemovedInDjango19Warning: Loading the`url` tag from the `future`
library is deprecated and will be removed in Django 1.9. Use the default `url` tag instead.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ env:
   - DJANGO_VERSION=1.5.5
   - DJANGO_VERSION=1.6.10
   - DJANGO_VERSION=1.7.4
+matrix:
+  exclude:
+    - python: "2.6"
+      env: DJANGO_VERSION=1.7.4
 install:
   - pip install -e .
   - pip install Django==$DJANGO_VERSION django-discover-runner sqlparse
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,9 @@ python:
   - "3.3"
   - "3.4"
 env:
-  - DJANGO_VERSION=1.4.10
   - DJANGO_VERSION=1.5.5
-  - DJANGO_VERSION=1.6.1
-matrix:
-  exclude:
-    - python: "3.2"
-      env: DJANGO_VERSION=1.4.10
-    - python: "3.3"
-      env: DJANGO_VERSION=1.4.10
-    - python: "3.4"
-      env: DJANGO_VERSION=1.4.10
+  - DJANGO_VERSION=1.6.10
+  - DJANGO_VERSION=1.7.4
 install:
   - pip install -e .
   - pip install Django==$DJANGO_VERSION django-discover-runner sqlparse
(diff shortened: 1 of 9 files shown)


aaugustin:
We can't do that as long as we support Django 1.4 LTS, as the build failure shows.

renstrom:
Do you have a development branch I can target instead of master? It seems extremely wasteful to stall all new development until Django 1.4 is deprecated sometime in October.

aaugustin:
Well, it's also quite wasteful to engage in hyperbolic arguments that discredit your position.

This doesn't stall *all* new development, plenty of tickets can move forwards.

kevinpostal:
Quick to jump on people who are trying to be supportive to the debug-toolbar. @aaugustin  Try working on your aspergers instead of your programing abilities.

tim-schilling:
Kevin, please refrain from personal attacks. The toolbar follows Django's version support. That means Django 1.4 will be supported by the toolbar until [October](https://docs.djangoproject.com/en/1.7/internals/release-process/#long-term-support-lts-releases).

renstrom:
I'm just trying to pave way for the upcoming Django 1.8 release. But it was apparently in vain. I'm closing this so that I'm not breaking your precious master branch, and you can fix yourself in 8 months.

renstrom:
Here's a couple of suggestions though on my way out the door:

- Create a LTS/1.4 branch, merge everything to master and backport bugs and stuff to the release branch.
- Do the opposite, create a develop (or something) branch which is allowed to break backward compatibility for forward compatibility. And merge it to master when appropriate.
- Just create a new release (`v2.0.0` or something) and tell everyone still on Django 1.4 that they have to install < `v2.0.0`. This is kind of what iPython did with Python 2.6 and their v2.0 release.

aaugustin:
Now I remember why I refrained from touching this repo in months... Anyway...

@renstrom Our problem is that the debug toolbar gets at best five hours of its maintainers' time per month (@tim-schilling and myself).

A handful of people have offered to help but none of them have gone as far as submitting a patch for a non-trivial issue, which isn't a good indicator of long term commitment.

There are plenty of things we could do, but we aren't keen to:

- add admistrative overhead (like managing backports): that could make us go from "make a release with some bugfixes from time to time" to "not making any releases"
- do anything that could result in a stream of support tickets (like requiring different versions of the toolbar depending on the Django version)

I'm not saying this is a great situation, but that's where we are. It used to be worse: at some point there was an **unmaintained!** disclaimer in the README.

renstrom:
Right, fair enough, it's not always easy to maintain a popular repo. I don't really have a great solution, just thought I'd offer some suggestions. Thanks for taking your time reply though.

aaugustin:
Cool. I understand you were trying to help and I'm sorry we don't have a good solution for accepting help in this area at this time.

aaugustin:
One way to fix the original issue without breaking Django 1.4 is to ship different sets of templates for different Django versions. Thanks @apollo13 for the suggestion.

PR #694, opened 2015-03-04
Compatibility refactor

### LINKED ISSUE #668: Not work with Django-current, "cannot import name get_standard_processors"
After "Encapsulated TEMPLATE_CONTEXT_PROCESSORS in Engine." https://github.com/django/django/commit/37505b6397058bcc3460f23d48a7de9641cd6ef0 debug panel not work with current version of Django.

django.core.exceptions.ImproperlyConfigured: Error importing debug panel debug_toolbar.panels.templates: "cannot import name get_standard_processors"

tim-schilling:
This PR pulls in all of the try/except ImportError logic for Django and Python version incompatibility issues. It's based off of #674 (thanks @apollo13 for getting us started) and would fix #668.

I left the threading, resource, selenium, and pygments try/except logic where it was since they are either platform or dependency issues.

I wanted to refactor the logic for the CacheHandler in debug_toolbar.panels.cache like I did for the Engine class, but it's too coupled at this point. If someone has an idea of how to do that relatively cleanly, I'm all ears.

### DIFF
--- a/debug_toolbar/compat.py
+++ b/debug_toolbar/compat.py
@@ -5,6 +5,13 @@ In order to avoid circular references, nothing should be imported from
 debug_toolbar.
 """
 
+
+try:
+    from django.core.cache import CacheHandler, caches
+except ImportError:  # Django < 1.7
+    CacheHandler = None
+    caches = None
+
 try:
     from django.template.engine import Engine
 except ImportError:  # < Django 1.8
--- a/debug_toolbar/compat.py
+++ b/debug_toolbar/compat.py
@@ -5,6 +5,7 @@ In order to avoid circular references, nothing should be imported from
 debug_toolbar.
 """
 
+from django.conf import settings
 
 try:
     from django.core.cache import CacheHandler, caches
@@ -16,7 +17,7 @@ try:
     from django.template.engine import Engine
 except ImportError:  # < Django 1.8
     Engine = None
-    from django.template.context import get_standard_processors  #NOQA
+    from django.template.context import get_standard_processors  # NOQA
     from django.template.loader import find_template_loader  # NOQA
 
 try:
@@ -45,3 +46,32 @@ try:
 except ImportError:  # >= Django 1.7
     import weakref
     WEAKREF_TYPES = weakref.ReferenceType,
+
+
+def get_template_dirs():
+    """Compatibility method to fetch the template directories."""
+    if Engine:
+        template_dirs = Engine.get_default().dirs
+    else:  # < Django 1.8
+        template_dirs = settings.TEMPLATE_DIRS
+    return template_dirs
+
+
+def get_template_loaders():
+    """Compatibility method to fetch the template loaders."""
+    if Engine:
+        loaders = Engine.get_default().template_loaders
+    else:  # < Django 1.8
+        loaders = [
+            find_template_loader(loader_name)
+            for loader_name in settings.TEMPLATE_LOADERS]
+    return loaders
+
+
+def get_template_context_processors():
+    """Compatibility method to fetch the template context processors."""
+    if Engine:
+        context_processors = Engine.get_default().template_context_processors
+    else:  # < Django 1.8
+        context_processors = get_standard_processors()
+    return context_processors
--- a/debug_toolbar/compat.py
+++ b/debug_toolbar/compat.py
@@ -1,0 +1,40 @@
+"""
+This file exists to contain all Django and Python compatibility issues.
+
+In order to avoid circular references, nothing should be imported from
+debug_toolbar.
+"""
+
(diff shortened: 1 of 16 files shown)


tox.ini
```diff
@@ -17,7 +17,7 @@ deps =
     django14: Django>=1.4,<1.5
     django15: Django>=1.5,<1.6
     django16: Django>=1.6,<1.7
-    django17: Django>=1.7,<1.8
+    django17: Django>=1.7,<1.8a
```
apollo13: Are you sure that is needed? By default pip ignores "pre" releases, so "<1.8" should be fine.

tim-schilling: When I run it locally, it installs 1.8b when it's specified as < 1.8

apollo13:
```
(tmp-d1fc6f38ef61116d)florian@apollo13$ pip install "Django<1.8"
Collecting Django<1.8
  Downloading Django-1.7.5-py2.py3-none-any.whl (7.4MB)
    100% |################################| 7.4MB 94kB/s 
Installing collected packages: Django

Successfully installed Django-1.7.5
(tmp-d1fc6f38ef61116d)florian@apollo13$ pip --version
pip 6.0.8 from /home/florian/.virtualenvs/tmp-d1fc6f38ef61116d/local/lib/python2.7/site-packages (python 2.7)
```

aaugustin:
@tim-schilling Perhaps you're using an "old" version of pip (pip changes quickly).

tim-schilling:
That's what I'm thinking. I'll check it out.

tim-schilling:
I'm a dummy, I had to upgrade tox. I'll get that commit removed, thanks for pointing that out!

debug_toolbar/compat.py
```diff
@@ -0,0 +1,77 @@
...
+
+In order to avoid circular references, nothing should be imported from
+debug_toolbar.
+"""
+
+from django.conf import settings
+
+try:
+    from django.core.cache import CacheHandler, caches
+except ImportError:  # Django < 1.7
```
aaugustin: You've used `< Django 1.7` everywhere else (but I prefer `Django < 1.7`).

Just normalize it one way or another.

debug_toolbar/compat.py
```diff
@@ -0,0 +1,77 @@
...
+try:
+    from collections import OrderedDict
+except ImportError:  # < python 2.7
+    from django.utils.datastructures import SortedDict as OrderedDict  # NOQA
+
+try:
+    from django.contrib.staticfiles.testing import (
+        StaticLiveServerTestCase as LiveServerTestCase)
+except ImportError:  # < Django 1.7
+    from django.test import LiveServerTestCase  # NOQA
```
aaugustin: I find it better to always import as the new name, in this case, `StaticLiveServerTestCase`. This is mostly a matter of looking towards the future, not the past :-)

debug_toolbar/compat.py
```diff
@@ -0,0 +1,77 @@
...
+
+try:
+    from django.db.backends import utils
+except ImportError:  # >= Django 1.7
+    from django.db.backends import util as utils  # NOQA
```
aaugustin: I would suggest to import `as db_backends_util` because `from .compat import as db_backends_util` is much more explicit than `from .compat import utils`.

debug_toolbar/compat.py
```diff
@@ -0,0 +1,77 @@
...
+    from django.dispatch.dispatcher import WEAKREF_TYPES
+except ImportError:  # >= Django 1.7
+    import weakref
+    WEAKREF_TYPES = weakref.ReferenceType,
+
+
+def get_template_dirs():
+    """Compatibility method to fetch the template directories."""
+    if Engine:
+        template_dirs = Engine.get_default().dirs
```
aaugustin: In this method, and all those that use `Engine.get_default()`, it would be more robust to catch `ImproperlyConfigured` and return a default value like the empty list. That will avoid a crash when there isn't exactly one Django templates backend configured.

aaugustin:
Great work. I made a few comments on the compat module but nothing major. Once you've addressed them (or decided to ignore them) and the tests pass on 1.8 beta please merge this PR.

You can add 1.8 beta to the tox configuration by editing it as follows:

    envlist =
        ...
        py{27,32,33,34}-django{17,18},

    deps =
        ...
        django18: https://www.djangoproject.com/download/1.8b1/tarball/

aaugustin:
@tim-schilling Do you want me to take it from there? I can complete the patch this week-end.

tim-schilling:
Yes, I think that'd be the better approach. Thanks.

aaugustin:
Merged (as well as a few additional commits).

PR #697, opened 2015-03-10
Resize example screenshot to not break pypi

adamchainz:
It irks me that when I open the pip page for django-debug-toolbar I get a broken design. It's pypi's fault really but hey you can't change that quickly.

screen shot 2015-03-10 at 15 17 22

This commit resizes the screenshot so it doesn't (plus I optimized it with `pngcrush`). Simulation screenshot of the future (?):

screen shot 2015-03-10 at 15 17 09

And hey, we don't all have retina screens, so let's save some pixels and bytes.


renstrom:
To save some space you could disable the screenshot border as well, it takes up a fair amount of space:

    $ defaults write com.apple.screencapture disable-shadow -bool true
    $ killall SystemUIServer

Or remove it manually using some photo editor.

adamchainz:
Removed it!

tim-schilling:
Thanks for the PR!

Issue #688, opened 2015-02-19
Introduce warnings of certain DB access patterns using SQL "explain" statement


mrkschan:
Sometimes, certain DB access patterns reported in "EXPLAIN" statement are better to be prevented for scalability reason (e.g. filesort in MySQL). These patterns can be caught during the development phase of a project before the project hitting the mass (especially missing index :). Why can't DjDT add warnings to those patterns?

Using "EXPLAIN" would introduce extra DB queries. Thus, the "auto EXPLAIN" queries are better turned off by default. And, some of the queries are repeating, the "EXPLAIN" results can be cached. When certain DB access patterns that are non-scalable is found, warning message of the SQL statement is inserted to the HTTP response.

Would this feature request be accepted to DjDT? I'm willing to implement this :)

aaugustin:
It's hard to commit to accepting a patch before seeing it ;-) but this looks interesting.

Ideally, the code should be sufficiently modular so warnings can be added for other databases.

mrkschan:
@aaugustin DjDT now does not use django messages framework. Shall I introduce something really new or just patching the SQLPanel?

aaugustin:
I think the logic should be contained within the SQLPanel or a new, standalone panel -- which can be developed externally at first.

mrkschan:
@aaugustin, do you think reusing the SQL tracking logic from SQLPanel is a good idea? https://github.com/mrkschan/django-debug-toolbar/commit/d1d8a422209838f2bdf58a41fa57916038de3449

PR #698, opened 2015-03-11
Start SQLWarningsPanel to warn non-scalable SQL statement

mrkschan:
Introduced a SQLWarningsPanel to detect any non-scalable SQL statement using EXPLAIN. Included some simple warnings for MySQL EXPLAIN.

#688

### DIFF
--- a/debug_toolbar/panels/sqlwarnings.py
+++ b/debug_toolbar/panels/sqlwarnings.py
@@ -29,6 +29,27 @@ def unwrap_cursor(conn):
         conn.cursor = conn._djdt_sqlwarnings_cursor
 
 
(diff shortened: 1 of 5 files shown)


tim-schilling:
I'm still on the fence about including a panel that only works with one vendor. It seems better suited to be a third party panel until there's support for others.

aaugustin:
Yes, I agree that this seems more appropriate as a third-party panel. We can link to it in the docs as soon as it's released.

mrkschan:
Let's close this until it has more vendors :)

mrkschan:
Let's close this until it has more vendors :)

PR #702, opened 2015-04-01
Fix incomplete html tag.

pandabuilder:


tim-schilling:
Good catch. Thank you. This was included with https://github.com/django-debug-toolbar/django-debug-toolbar/commit/51aa643c8856e862160c1897a3538af5a082910e

PR #704, opened 2015-04-10
 Added support for Content Security Policy, Rebase #330

graingert:


graingert:
This is a rebase of #330

debug_toolbar/templates/debug_toolbar/panels/templates.html
```diff
@@ -35,7 +35,7 @@
 	<dt><strong>{{ key|escape }}</strong></dt>
 	<dd>
 		<div class="djTemplateShowContextDiv"><a class="djTemplateShowContext"><span class="toggleArrow">&#x25B6;</span> {% trans "Toggle context" %}</a></div>
-		<div class="djTemplateHideContextDiv" style="display:none;"><code>{{ value|escape }}</code></div>
+		<div class="djTemplateHideContextDiv" hidden><code>{{ value|escape }}</code></div>
```
tim-schilling: Is there a reason why you didn't use ```hidden="hidden"``` here and on line 22?

tim-schilling:
This looks pretty good. Could you please add a ```djdt-``` prefix to the ```width-*``` classes? Developers may have stylesheets defined that would apply their styles to elements with that class name. Adding that prefix would help mitigate that possibility.

graingert:
Done, I've also prefixed the CSS class 'highlighted'

thedrow:
@graingert Can you please rebase again? The PR contains merge conflicts.

graingert:
re-based, also killed that other "hidden"

graingert:
@tim-schilling @thedrow poke

tim-schilling:
I tried setting up a test application with a content security policy enabled and I had to add ```'self' ajax.googleapis.com``` to the ```script-src``` directive and ```'self' data:``` to the ```img-src``` directive. Does that sound correct? If so, then I'm going to add some documentation that explains this.

graingert:
Yes that's correct. Although if you set a custom JQuery URL it will be whatever the origin of that URL is.

tim-schilling:
I was doing some more testing and whenever I click on the SQL tab, it generates the unsafe-eval error. Can you duplicate that issue?

graingert:
@tim-schilling Can you paste me the full error?

Do you have your test project on GitHub?

graingert:
I've repeated the issue. We need to allow `script-src "unsafe-eval"` for now.

For now can you merge this and I will work on a fix?

aaugustin:
Yes, I think we should merge this PR before it needs rebasing again.

@graingert Fortunately there isn't a lot of activity, only @tim-schilling and myself merge patches.

I think it would be helpful to add guidelines to the contributing documentation so future changes don't break CSP inadventently. I'm not sure if it could be tested automatically e.g. with Selenium?

@tim-schilling As far as I am concerned, I don't mind merging the patch as is and iterating.

tim-schilling:
I meant to respond earlier. I decided that I'd merge it but not document it
until the unsafe-eval directive isn't needed. I don't think we can say the
toolbar supports a content security policy until that directive is no
longer needed.

tim-schilling:
I rebased your changes on master, and merged them in with commit 8b29127ffb385db22fc2a3296b9904dd73180e56. Thank you for the PR @graingert!

PR #707, opened 2015-04-27
Only import url from future when using Django < 1.5.

### LINKED ISSUE #706: Removed In Django 1.9 Warning: url tag from future
The full warning:

/Users/JJZ/.virtualenvs/athletesunited/lib/python3.4/site-packages/django/templatetags/future.py:25: RemovedInDjango19Warning: Loading the `url` tag from the `future` library is deprecated and will be removed in Django 1.9. Use the default `url` tag instead.
  RemovedInDjango19Warning)

I've checked through all my code and I can't find a trace of where I do: {% load url from future %}

Now I've moved onto trying to help third party developers fix their code because I'm getting pretty tired of the warnings popping up. When I ran a search on your code base:

https://github.com/django-debug-toolbar/django-debug-toolbar/search?utf8=%E2%9C%93&q=%7B%25+load+url+from+future+%25%7D

I see that you have used this, this is from way back in Django 1.3 and it is being removed in 1.9 so if you all would, could you please go through clear this all up and let me know when the commit has been pushed so I can pull the updates?

Thanks a lot,

JJ

PS: https://docs.djangoproject.com/en/1.8/releases/1.7/#loading-ssi-and-url-template-tags-from-future-library

tim-schilling:
In order to hide the deprecation warnings, determine what version
django is being used, then either import the url template tag
from future or import the normal url template tag.

When support for Django 1.4 drops, this change can be removed.
Fixes #706.

### DIFF
--- a/debug_toolbar/compat.py
+++ b/debug_toolbar/compat.py
@@ -5,6 +5,7 @@ In order to avoid circular references, nothing should be imported from
 debug_toolbar.
 """
 
+import django
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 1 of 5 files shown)


tim-schilling:
Rebased on master.

PR #711, opened 2015-04-29
Don't ignore templates without an origin

maikhoepfel:
We're running Django 1.7 with django-jinja. For whatever reasons, the template origin isn't set, and the panel shows no template context whatsoever. Upon investigation, I discovered that the template rendered signal is indeed fired, but the template is skipped as no origin is set.

I deny any understanding of the existing code. I do not know why templates without an origin are skipped. But without this change, we aren't getting a template context in the panel, and with it, we are. So at first sight, it looks like it's safe to do. But somebody more qualified than me has to be the judge on that.

Furthermore, this code marks the 'No origin' string as translatable.

### DIFF
--- a/debug_toolbar/panels/templates/panel.py
+++ b/debug_toolbar/panels/templates/panel.py
@@ -201,12 +201,10 @@ class TemplatesPanel(Panel):
             info = {}
             # Clean up some info about templates
             template = template_data.get('template', None)
-            if not hasattr(template, 'origin'):
-                continue
-            if template.origin and template.origin.name:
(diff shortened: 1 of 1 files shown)


aaugustin:
Well, the origin API isn't very well specified, so it's hard to say what should or shouldn't be. Thankfully @prestontimmons is working hard to fix that :-)

Unless someone sees a problem with this patch, I think we should merge it as a stopgap solution until Django gets a better API.

prestontimmons:
Seems reasonable to me.

maikhoepfel:
Lovely, thanks!

Issue #570, opened 2014-03-31
NotImplementedError for cache clear


Gwildor:
I'm familiar with ticket #439, and I'm running DDT 1.0.1, which includes the fix created for that ticket. However, when I try to clear the cache with DDT enabled, I get a NotImplementedError, which originates directly from the `django.core.cache.backends.base.BaseCache.clear` method. As my backend, I'm running memcached. Disabling the DDT works; no exception is raised. Note that I'm trying to clear the cache from a signal receiver (not sure if that's relevant, so just throwing it out there).

When I do `from django.core.cache import cache; print cache` I get `<CacheStatTracker for <django.core.cache.backends.memcached.MemcachedCache object at 0x106b7a290>>`.

PR #712, opened 2015-05-01
Fix #670 and #570. Add 'clear()' to CacheStatTracker

### LINKED ISSUE #670: cache.clear() not available inside a view function
cache.clear() raises NotImplementedError during request/view processing between debug toolbar middleware request processing and debug toolbar middleware response processing.

This is a copy of my message in closed issue #570:

It took me all of 10 minutes to find out why this was happening in my project, inside a view function:
- the cache panel replaces the default cache (and the get_cache) function during the request middleware processing with CacheStatTracker/a function returning CacheStatTracker
- processing continues until the actual view function is called
- cache.clear() is called inside the view function, on CacheStatTracker
- CacheStatTracker has no clear() method, but has inherited the clear() method from BaseCache
- clear() is called on BaseCache
- BaseCache raises NotImplementedError

You won't find this occurring outside a view function, because on the way back out, the processing passes through the debug toolbar middleware again, this time calling disable_instrumentation on the cache panel, which replaces the cache/get_cache with the original values again.

blopker:
This PR adds the `clear()` method to CacheStatTracker. Without this Django will raise `subclasses of BaseCache must provide a clear() method` whenever `cache.clear()` is called.

Cheers! :beers:

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -82,6 +82,10 @@ class CacheStatTracker(BaseCache):
         return self.cache.delete(*args, **kwargs)
 
     @send_signal
+    def clear(self, *args, **kwargs):
+        return self.cache.clear(*args, **kwargs)
+
+    @send_signal
     def has_key(self, *args, **kwargs):
         # Ignore flake8 rules for has_key since we need to support caches
         # that may be using has_key.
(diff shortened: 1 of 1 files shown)


tim-schilling:
I added a test to verify the ```clear``` method can be called as well. Thank you for the PR!

PR #715, opened 2015-05-24
Package name column added to Versions panel.

idlesign:
Hi,

Since in Django 1.7+ `ugettext` is applied to `app.verbose_name` it might be hard to identify the very package name. Here it's solved by a new `Package` column: so we know both the package name and a title it is represented in Admin contrib.

### DIFF
--- a/debug_toolbar/panels/versions.py
+++ b/debug_toolbar/panels/versions.py
@@ -6,7 +6,7 @@ import django
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 
-from debug_toolbar.compat import import_module, OrderedDict
+from debug_toolbar.compat import import_module
 from debug_toolbar.panels import Panel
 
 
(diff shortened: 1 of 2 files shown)


tim-schilling:
Do you have an example of a package that would benefit from this change?

idlesign:
Those two are from my list, yet virtually it could be any app. The same way we localize verbose names in fields and model meta we can give our apps user-friendly names.

https://github.com/idlesign/django-sitetree/blob/master/sitetree/config.py#L9
https://github.com/idlesign/django-sitegate/blob/master/sitegate/config.py#L9

aaugustin:
I'm sorry, I failed to merge this PR for more than a year and it no longer applies.

But the underlying ideas sounds interesting and the implementation looks OK.

If you rebase it on master and ping me, I'll merge it.

idlesign:
Ended up with #859

PR #719, opened 2015-06-05
Display the count of SQL duplicates (fixes #27)

### LINKED ISSUE #27: SQL: Show duplicate queries as in SQL Log Middleware
SQL Log Middleware is here: http://www.djangosnippets.org/snippets/344/

Working solution proposition:

http://dumpz.org/12573/
http://dumpz.org/12575/

ewjoachim:
This if my most successful approach at fixing #27

If we stick to just displaying the number of duplicates for each database connection and each query, we get some information, and it will help debugging.

I've been trying to push it a bit further and trying to display only the unique queries at first, with a simple way of showing everything and clear way to understand the groups but I couldn't get something easily understandable without changing the whole UI (which may be interesting, but is definitely out of the scope of this ticket).

That being said, I have some ideas and I'll try to implement something at some point.

### DIFF
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -139,6 +139,7 @@ class SQLPanel(Panel):
     def process_response(self, request, response):
         colors = contrasting_color_generator()
         trace_colors = defaultdict(lambda: next(colors))
+        query_duplicates = defaultdict(lambda: defaultdict(int))
         if self._queries:
             width_ratio_tally = 0
             factor = int(256.0 / (len(self._databases) * 2.5))
@@ -161,6 +162,8 @@ class SQLPanel(Panel):
             trans_id = None
             i = 0
             for alias, query in self._queries:
+                query_duplicates[alias][query["raw_sql"]] += 1
+
                 trans_id = query.get('trans_id')
                 last_trans_id = trans_ids.get(alias)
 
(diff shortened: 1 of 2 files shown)


debug_toolbar/panels/sql/panel.py
```diff
@@ -204,6 +207,30 @@ def process_response(self, request, response):
...
+        # Queries are duplicates only if there's as least 2 of them.
+        # Also, to hide queries, we need to give all the duplicate groups an id
+        query_duplicates = {
+            alias: {
+                query: duplicate_count
+                for query, duplicate_count in queries.items()
+                if duplicate_count >= 2
+            }
+            for alias, queries in query_duplicates.items()
+        }
```
aaugustin: Until now we haven't used dict comprehensions because we're supporting Django 1.4 and Python 2.5. 1.4 is still supported until October 1st.

ewjoachim: Done :) (used dict((a,b) for a, b in ...), that should be better...)

tim-schilling:
As an example here here's what the output will now look like: 
image

It appears that a query will be counted as a duplicate depending on the query alone rather than the actual values of the parameters. I think that this will work for most cases and is better than checking the parameter values.

One thing that I'm not thrilled about is that there's no distinction between which queries are duplicates of which. It's easy to which query is being duplicated in the example I posted, but I could see this being difficult with more complex pages. To mitigate this, I can see this being a separate tab on the SQL Panel while the main panel would continue to be what exists currently. That way a developer could see what the flow of queries is and then see a barebones list of the duplicated ones.

All that said, I'm leaning towards merging this because it does improve the toolbar. What are your thoughts @aaugustin?

ewjoachim:
Yeah, I thought it wasn't a good idea to radically change the SQL panel, but, indeed, creating a new tab that would display the information without the timeline but with a cleaner way to show how duplicates are grouped could help a lot when trying to find which loop create a cascade of queries.

Regarding the "Which queries are duplicates" question, would you like me to try to add a color information (like a thin rectangle filled with a solid color on the left of the ``<tr>`` ?)

aaugustin:
@tim-schilling I discussed this with @ewjoachim before he implemented it at the DjangoCon Europe sprints.

His work goes in the right direction. Adding a visual cue that queries are duplicated would be nice. I've thought of "folding" duplicated queries in a single line but that'll make it impossible to display the parameters of each query, which may be a problem.

tim-schilling:
Okay cool. I'm going to merge this in then and we can iterate on it.

aaugustin:
Thanks Tim!

ewjoachim:
Thank you @tim-schilling , and @aaugustin , I'll try to continue in this direction :)

PR #720, opened 2015-06-06
Update the cache panel to support the cache_page decorator.

### LINKED ISSUE #718: Cache panel does not update for per-view level caching
I'm using the toolbar 1.3 and Django 1.8.2.  The toolbar displays fine and the SQL, Time, Headers, Request, Static Files, Templates panels all update but the Cache panel stays fixed at 0 calls even though it's activated.  I've tried it using the LocMemCache and Memcached with same result.  I'm using the explicit setup with DEBUG_TOOLBAR_PATCH_SETTINGS = False.

Here are my Middleware classes:

MIDDLEWARE_CLASSES = (
    'debug_toolbar.middleware.DebugToolbarMiddleware',
    'django.middleware.cache.UpdateCacheMiddleware',
    'django.contrib.sessions.middleware.SessionMiddleware',
    'django.middleware.common.CommonMiddleware',
    'django.middleware.csrf.CsrfViewMiddleware',
    'django.contrib.auth.middleware.AuthenticationMiddleware',
    'django.contrib.auth.middleware.RemoteUserMiddleware',
    'django.contrib.messages.middleware.MessageMiddleware',
    'django.middleware.clickjacking.XFrameOptionsMiddleware',
    'django.middleware.cache.FetchFromCacheMiddleware',
)

In my urls I have:

if settings.DEBUG:
    import debug_toolbar
    urlpatterns += patterns('',
        url(r'^__debug__/', include(debug_toolbar.urls)),
    )

Is there anything else I'm missing to get the panel to update?

tim-schilling:
Unfortunately, once the CacheMiddleware's caches/get_cache has been
monkey patched, it can't be disabled. This results in a bit of overhead.

Fixes #718

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -11,6 +11,7 @@ from django.core.cache import (
     get_cache as original_get_cache)
 from django.core.cache.backends.base import BaseCache
 from django.dispatch import Signal
+from django.middleware import cache as middleware_cache
 from django.utils.translation import ugettext_lazy as _, ungettext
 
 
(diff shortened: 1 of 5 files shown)


aaugustin:
Looks reasonable.

PR #721, opened 2015-06-08
Adding a color information to distinguish duplicate queries

ewjoachim:
This pull request simply adds a color element to distinguish between each duplicated request, as a follow-up to the conversation on #719 

This is what it would look like
There's a colored rectangle next to each "Duplicated x times" line

I would simply like to add a note : the SQL panel uses several techniques to generate different colors, and I feel that it would be quite simpler if it was done using ``import colorsys`` (from the standard lib) which provides conversion functions from HSV (which allows selecting a color with full saturation and full value, varying only the hue from 0 to 1 instead of making convoluted operations with RGB) (As a side I'd even add that CSS supports HSL (HSL is most like HSV) input directly for most browser except IE <= 8.)

I'd be happy to redo the code for that part too, but I'm waiting for your input on that.

### DIFF
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -209,9 +209,10 @@ class SQLPanel(Panel):
 
         # Queries are duplicates only if there's as least 2 of them.
         # Also, to hide queries, we need to give all the duplicate groups an id
+        query_colors = contrasting_color_generator()
         query_duplicates = dict(
             (alias, dict(
-                (query, duplicate_count)
+                (query, (duplicate_count, next(query_colors)))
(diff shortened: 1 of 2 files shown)


aaugustin:
That commit looks reasonable to me.

If you want to refactor color handling, please open another pull request. I don't feel a strong urge to fix something that isn't broken, but if it allows removing some code, why not.

You could have a look at existing algorithms for generating contrasting colorschemes. I don't know if the current version implements a common algorithm or reinvented that wheel.

ewjoachim:
It's been a few months, now, and I'm not sure what we're waiting for... Is there something I can do to help ?

PR #722, opened 2015-06-15
Really hide the toolbar when printing

aumo:
Use `!important` to hide the toolbar handle when printing
so it does not get overridden by inline CSS.


tim-schilling:
Nice catch. Thank you.

PR #733, opened 2015-07-14
Check User IP from HTTP_X_FORWARDED_FOR http header if exist

appsolute:
Check User IP from HTTP_X_FORWARDED_FOR http header if exist

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -24,7 +24,12 @@ def show_toolbar(request):
     """
     Default function to determine whether to show the toolbar on a given page.
     """
-    if request.META.get('REMOTE_ADDR', None) not in settings.INTERNAL_IPS:
+    x_forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
+    if x_forwarded_for:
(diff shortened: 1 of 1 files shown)


aaugustin:
Django used to contain logic along these lines, which was removed for security reasons, in 1.4 or perhaps earlier. I don't want to introduce a similar vulnerability in the debug toolbar, all the more since it can leak sensitive information.

appsolute:
Ok if this's about security reason I do agree with you.
So, Do you know any solutions for the Django application that run behind proxy (reverse proxy) or run with in docker (in my case) to allow only admin to access Debug Toolbar?

tim-schilling:
You can use your own function and specify it as the ```SHOW_TOOLBAR_CALLBACK``` in the [toolbar configuration](http://django-debug-toolbar.readthedocs.org/en/1.3/configuration.html#debug-toolbar-config).

aaugustin:
Or write a middleware to extract the actual IP and put it in `REMOTE_ADDR`.

appsolute:
Thank you @aaugustin, I'll try this solution.
Also thank you @tim-schilling.

Issue #705, opened 2015-04-17
Error with MySQL/Postgres database using % in direct SQL


vaal-:
TypeError: not enough arguments for format string

I am using SQL formatting dates in view.
DATE_FORMAT(payment_date, '%Y-%m-%d') as date
There are no parameters.

This is probably connected with this
https://code.djangoproject.com/ticket/9055
https://github.com/django/django/commit/76aecfbc4b49f5ab0613cccff1df6fab03253fab

https://github.com/django-debug-toolbar/django-debug-toolbar/blob/master/debug_toolbar/panels/sql/tracking.py#L158

tim-schilling:
Have you tried using ```DATE_FORMAT(payment_date, '%%Y-%%m-%%d') as date```?

vaal-:
This variant is work only with debug panel.

If disable panel sql request remains in the same form - DATE_FORMAT(payment_date, '%%Y-%%m-%%d') as date . Which leads to a wrong response from database.

tim-schilling:
I'm having difficulty reproducing your problem. What version of Django are you using? What's the query you're using? How exactly is it breaking?

vaal-:
https://www.dropbox.com/s/t6gsoe63y0xox5z/django_debug_toolbar_bug.zip?dl=0
Here is an example project with reproduction problems

I used: Django 1.8, Python 2.7, MySQL-python 1.2.4, django-debug-toolbar 1.3.0
need to run manage.py syncdb and create 1 user

then open URLs
http://127.0.0.1:8000/bug1/
http://127.0.0.1:8000/bug2/
with enabled debug_toolbar
and then without debug_toolbar

Sorry for the long answer, there was no time.

tim-schilling:
I'm having difficulty building the environment for your test example. Before I invest more time in it, can you please try upgrading MySQL-python to 1.2.5?

vaal-:
I've updated, the problem still exist.
Describe your typical environment - I check it.

IlianIliev:
I can confirm that the issue exists only if Debug toolbar is ON. Otherwise it works fine.

tim-schilling:
This is a problem with postgres as well as per #745

tim-schilling:
This was fixed in #735

PR #735, opened 2015-08-01
fix for issue #705

vaal-:


aaugustin:
PEP249 doesn't specify a default value of `params` to signify that it's empty / not provided.

I'm slightly afraid this will fix one database backend but break another...

vaal-:
We can make these changes in Django mysql backend?
https://github.com/farcepest/MySQLdb1/blob/master/MySQLdb/cursors.py#L182
https://github.com/PyMySQL/mysqlclient-python/blob/master/MySQLdb/cursors.py#L195
https://github.com/mysql/mysql-connector-python/blob/master/lib/mysql/connector/django/base.py#L197

If I understand correctly, all three connectors for mysql expected None as defaults.

aaugustin:
Actually Django itself defaults to `params=None` and calls the underlying function with one arg if `params is None` and two args if `params is not None`. So this change is correct.

PR #743, opened 2015-08-31
Replace hidden="hidden" attribute with style="display: none;".

### LINKED ISSUE #742: Use of hidden="hidden" breaks DjDT when used with bootstrap4 and pureCSS
`django-debug-toolbar` uses `hidden="hidden"` attributes on its HTML elements which are left regardless of element being displayed or not. This causes `django-debug-toolbar` UI (both handle and main interface) to remain hidden when used together with Bootstrap 4 or PureCSS.

Both Boostrap 4 and PureCSS use:
```
[hidden] {
  display: none !important;
}
```
https://github.com/twbs/bootstrap/blob/v4.0.0-alpha/scss/_utilities.scss#L35

The problem is that jQuery `.show()` does not remove `hidden="hidden"` attribute but instead sets `style="display: block"`, which is overriden by the above CSS block (See note about `!important` in http://api.jquery.com/show/)

One approach would be to follow jQuery docs and use `.css('style', 'display: block !important')`. A better approach would be to use `style="display: none"` instead of `hidden="hidden"` which I think is better, since `style` attribute is also used for setting DjDT handle position.

frgtn:
Replaces `hidden="hidden"` attrs with `display: none` to resolve Bootstrap 4 compatibility issues described in #742.

Closes #742

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/base.html
+++ b/debug_toolbar/templates/debug_toolbar/base.html
@@ -10,10 +10,10 @@
 <script src="{% static 'debug_toolbar/js/jquery_existing.js' %}"></script>
 {% endif %}
 <script src="{% static 'debug_toolbar/js/toolbar.js' %}"></script>
-<div id="djDebug" hidden="hidden" dir="ltr"
+<div id="djDebug" style="display: none;" dir="ltr"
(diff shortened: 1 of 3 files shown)


tim-schilling:
Merging this will re-open #330.

frgtn:
Thanks for taking a look, gonna try a different approach.

PR #744, opened 2015-08-31
Remove hidden attr when showing DjDT items and readd it when hiding.

frgtn:
Another take at #742.

Removes `hidden` attribute when showing items and adds it when hiding. It does add the attr in a lot of places.

This approach does seem to break fade-in animation though. Fade-out animation is unaffected.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -6,7 +6,7 @@
         },
         isReady: false,
         init: function() {
-            $('#djDebug').removeAttr('hidden').show();
+            $('#djDebug').djdtShow();
             var current = null;
             $('#djDebugPanelList').on('click', 'li a', function() {
                 if (!this.className) {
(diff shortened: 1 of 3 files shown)


debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -207,13 +207,15 @@
...
             // close all panels
-            $('.djdt-panelContent').hide();
+            $('.djdt-panelContent').attr('hidden', 'hidden').hide();
             $('#djDebugToolbar li').removeClass('djdt-active');
             // finally close toolbar
-            $('#djDebugToolbar').hide('fast');
-            $('#djDebugToolbarHandle').show();
+            $('#djDebugToolbar').hide('fast', function() {
+                $(this).attr('hidden', 'hidden')
+            });
```
tim-schilling: How come you decided to use the callback function to remove the hidden attribute here rather than following the same method for all the other changes?

frgtn: Callback function is used to allow fade-out animation to finish.

tim-schilling: I didn't realize that there wasn't an animation if there weren't any arguments to the call.

frgtn:
I checked and it seems `function show_toolbar(animate)` is never called with `true`, hence fade-in animation code is never run. Removing the functionality if it's not used is probably out of scope of this PR, though I can do it in a new PR if people think it's a good idea.

tim-schilling:
It's unlikely that we'd remove ```show_toolbar``` since it's a part of the public API for the toolbar. An improvement I can see with this PR would be to create a jQuery plugin that would handle removing of the hidden attribute and handling the ```hide``` call at once to make things a bit cleaner. 

If you have the time to implement it, great. Otherwise I'll test this out when I have some time and get it merged.

frgtn:
I found some time this morning and implemented a jQuery plugin for show/hide functionality. It may be rough around the edges as I'm bad at Jabbascript. Feel free to comment if something looks wrong.

PR #746, opened 2015-09-02
Undefine "define.amd" instead of "define", to allow other RequireJS modules to be loaded

peap:
By undefining the entire `window.define`, issue #591 successfully prevents DDT's jQuery from registering itself with RequireJS under the "jquery" name. However, it also randomly prevents other RequireJS modules that are loading asynchronously from registering themselves via `window.define()`, since it's temporarily `undefined`.

This patch prevents jQuery from registering itself by unsetting `window.define.amd`, rather than `window.define`. As shown in #591, jQuery checks for a truthy `amd` property of `window.define` before registering itself, so this patch still accomplishes the goal of #591. Since `window.define` is still a function, however, other modules will still be able to register with `define()`.


aaugustin:
I don't want to have anything to do with this hack again, but this looks like an improvement.

peap:
Thanks, @aaugustin.

Is there someone else who could look at it, and/or anything I can do to help this along?

tim-schilling:
I'll take a look. I should have some time this weekend.

aaugustin:
Oh well let's merge it, we'll see what happens...

Issue #758, opened 2015-09-29
AttributeError: 'dict' object has no attribute 'dicts'


seperman:
I keep getting this error: `AttributeError: 'dict' object has no attribute 'dicts'`

aaugustin:
Based on the discussion in the PR, it looks like this is a symptom of the debug toolbar not being totally compatible with the support for multiple template engines in Django 1.8. Bumping severity to release blocker.

PR #759, opened 2015-09-29
Fixing AttributeError: 'dict' object has no attribute 'dicts'

seperman:
I kept getting `AttributeError: 'dict' object has no attribute 'dicts'` on django 1.8.3: AttributeError: #758

### DIFF
--- a/debug_toolbar/panels/templates/panel.py
+++ b/debug_toolbar/panels/templates/panel.py
@@ -121,46 +121,47 @@ class TemplatesPanel(Panel):
             return
 
         context_list = []
-        for context_layer in context.dicts:
-            temp_layer = {}
-            if hasattr(context_layer, 'items'):
-                for key, value in context_layer.items():
-                    # Replace any request elements - they have a large
-                    # unicode representation and the request data is
-                    # already made available from the Request panel.
-                    if isinstance(value, http.HttpRequest):
-                        temp_layer[key] = '<<request>>'
-                    # Replace the debugging sql_queries element. The SQL
-                    # data is already made available from the SQL panel.
-                    elif key == 'sql_queries' and isinstance(value, list):
-                        temp_layer[key] = '<<sql_queries>>'
-                    # Replace LANGUAGES, which is available in i18n context processor
-                    elif key == 'LANGUAGES' and isinstance(value, tuple):
-                        temp_layer[key] = '<<languages>>'
-                    # QuerySet would trigger the database: user can run the query from SQL Panel
-                    elif isinstance(value, (QuerySet, RawQuerySet)):
-                        model_name = "%s.%s" % (
-                            value.model._meta.app_label, value.model.__name__)
-                        temp_layer[key] = '<<%s of %s>>' % (
-                            value.__class__.__name__.lower(), model_name)
-                    else:
-                        try:
-                            recording(False)
-                            pformat(value)  # this MAY trigger a db query
-                        except SQLQueryTriggered:
-                            temp_layer[key] = '<<triggers database query>>'
-                        except UnicodeEncodeError:
-                            temp_layer[key] = '<<unicode encode error>>'
-                        except Exception:
-                            temp_layer[key] = '<<unhandled exception>>'
+        if 'dicts' in context:
+            for context_layer in context.dicts:
+                temp_layer = {}
+                if hasattr(context_layer, 'items'):
+                    for key, value in context_layer.items():
+                        # Replace any request elements - they have a large
+                        # unicode representation and the request data is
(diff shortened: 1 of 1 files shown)


aaugustin:
If we take this route `for context_layer in getattr(context, 'dicts', [])` would result in fewer changes and less indentation.

It would be nice to know why the attribute isn't there.

seperman:
Good point!

seperman:
I will dig into why the attribute doesn't exist as soon as I can.

seperman:
@aaugustin I changed the code based on your recommendation. Have not had a chance to figure out why the attribute doesn't exist. Do you need me to dig into that before you merge?

tschundler:
@aaugustin,

I work with @seperman and I know why it isn't there - it's because we're using Jinja2 templates, which doesn't use the magical Django context objects when rendering templates. Rather `context` is just a plain dict.
I suspect this will impact others using Jinja2 templates as of Django 1.8 (or was it 1.7?).

seperman:
@tschundler  Thanks is interesting. But I have used django debug toolbar with many projects that used jinja2.

debug_toolbar/panels/templates/panel.py
```diff
@@ -121,7 +121,7 @@ def _store_template_info(self, sender, **kwargs):
             return
 
         context_list = []
-        for context_layer in context.dicts:
+        for context_layer in getattr(context, 'dicts', []):
```
tschundler: I think better might be something like
```
getattr(context, 'dicts') or [context]
```

aaugustin:
@tschundler If I understand correctly, this bug only manifests when using Django's built-in support for Jinja 2 (which I happen to be responsible for)?

tschundler:
I was going to say yes, but figured I should double check and make my own test app. And debug toolbar works.

The site Sep & I are working on is not using quite vanilia Django templates for Jinja. Rather it's using something slightly modified so that we have context processors. So I took a look, and I see I also added this:
```
        test_signals.template_rendered.send(
            sender=self,
            template=self,
            context=context
        )
```

in my `JinjaTemplate.render` method - it was necessary to keep unit tests working, since like context processors, that functionality doesn't work with Django 1.8's template renderer. We had the same thing before with switched to Django 1.8 style backends, where it was just a special template loader. But then it was getting Django style contexts that it passed to the signal.

so... @aaugustin, maybe this is better as a question of how you're planning to support checking for template rendered and such in unit tests with Jinja2 in Django. (Or is that doable in newer versions?) We can adjust our code to match that. Or if you want to adjust your code to match ours:

```
def make_jinja2_context(context=None, request=None):
    if context is None:
        context = {}

    if request is not None:
        fresh_context = {'request': request}
        for conext_processor in env.django_context_processors:
            data = conext_processor(request)
            if data:
                fresh_context.update(**data)

        fresh_context.update(**context)
        context = fresh_context

    return context

class JinjaTemplate(jinja2.Template):
    def render(self, context=None, request=None):

        jinja_context = make_jinja2_context(context, request)

        test_signals.template_rendered.send(
            sender=self,
            template=self,
            context=context
        )

        return super(JinjaTemplate, self).render(jinja_context)
```

For this to work, we also set
```
env.template_class = JinjaTemplate
env.django_context_processors = map(import_string,
        settings.TEMPLATES[0]['OPTIONS']['context_processors'])
```
So we get what we need directly from the Jinja2 template loader. We also have code to do something like `APP_DIRS` using ChoiceLoader + PackageLoader. It's pretty full-featured. Even supports `signals.setting_changed`, and a way to register filters.

I had been meaning to clean it up to submit a pull request to Django project for never got around to it. (Maybe it should be that time?)

tschundler:
Globals work for constants, but not everything from context processors are constant. Several of ours rely on properties of the request. In the last two years, we've been moving to class based views, and most of those are now view mixins, and some could surely be middleware.

But it's a 5+ year old code base, and that was the most straight-forward way to move forward while minimizing what needs to be rebuilt.

aaugustin:
I recommend putting callables in Jinja2 globals and use ``{{ callable_global(request) }}` when you need the resulting value. This allows making a difference between context processors that return constants (a common case) and those that actually depend on the request (meaning you can't use them outside of the request processing e.g. when sending email from a scheduled task).

debug_toolbar/panels/templates/panel.py
matthiask: Indeed. Respectively `getattr(context, 'dicts', [context])`, otherwise the AttributeError will still happen.

matthiask:
I think this isn't an issue anymore. Also, review comments should be addressed (and a merge conflict fixed) if it still is. Thanks anyway!

PR #762, opened 2015-10-06
Support Django 1.9 & drop Django < 1.7

aaugustin:


aaugustin:
This is almost ready. There's just one test failure on Django 1.9 which I haven't had the time to investigate.

aaugustin:
The remaining test failure was caused by changes in the template loader API. I fixed it.

PR #770, opened 2015-11-09
Not empty context_processors from any template

kwist-sgr:
Fetch assigned context_processors from any template


matthiask:
Can you please explain? Why is this necessary, under what circumstances would the result be different?

PR #776, opened 2015-11-19
[WIP] Added wildcard support for INTERNAL_IPS

ZuluPro:
Added wildcard support with [`fnmatch.fnmatch`](https://docs.python.org/2/library/fnmatch.html) in `debug_toolbar.middleware.show_toolbar`.

User can now add easily a range of IP, by example:
- `192.168.*.*` (or `192.168.*`) for match local network
- `*` for crappy dev server

If this feature is accepted, let me update documentation before merge it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.pyc
 *.DS_Store
 *~
+*.sw[po]
 build
 .coverage
 dist
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import, unicode_literals
 
 import re
 import threading
+import fnmatch
 
 from django.conf import settings
 from django.utils import six
@@ -24,13 +25,17 @@ def show_toolbar(request):
     """
     Default function to determine whether to show the toolbar on a given page.
     """
-    if request.META.get('REMOTE_ADDR', None) not in settings.INTERNAL_IPS:
-        return False
-
+    show = bool(settings.DEBUG)
+    if request.META.get('REMOTE_ADDR', None) and show:
+        for ip in settings.INTERNAL_IPS:
+            if fnmatch.fnmatch(request.META['REMOTE_ADDR'], ip):
+                show = True
+                break
+        else:
+            show = False
     if request.is_ajax():
-        return False
-
-    return bool(settings.DEBUG)
+        show = False
+    return show
 
 
 class DebugToolbarMiddleware(object):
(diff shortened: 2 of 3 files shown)


ZuluPro:
@aaugustin Any news ?

aaugustin:
@ZuluPro I'm not sure what's right here. Besides I'm not working the toolbar anymore, except when I fancy.

ZuluPro:
Ok thanks !
@dcramer, @jezdez or @robhudson  ?

dcramer:
I would use CIDR not wildcards as its safer and more precise.

The Python ipaddr library makes this possible, and then its as simply as doing subset matching.

That said, unless Django says "INTERNAL_IPS can be things beyond a strict IPv4 address" then this should be a new setting that just defaults to INTERNAL_IPS.

dcramer:
Oh of note, I probably wouldn't deal with the merging here as I dont really have time and haven't been active in dev on DjDt lately.

ZuluPro:
Ok !

I'll use `ipaddr`, for check network, function will be really bigger than before for check IPv4, IPv6, IPs, and ranges, but I'll try to make as efficient as possible.

I suggest to add `settings.DEBUG_TOOLBAR_CONFIG['INTERNAL_IPS']` (or better `settings.DEBUG_TOOLBAR_CONFIG['ALLOWED_IPS']`) and use it with `settings.INTERNAL_IPS` for detection ?

@tim-schilling, What do you think about ?

ZuluPro:
@dcramer I tested to do it with `ipaddress` here: https://github.com/ZuluPro/django-debug-toolbar/tree/ipaddress

There's a performance issue with big networks (0.0.0.0 for example), for `in` operation it may have a tons of loop before find the `REMOTE_ADDR`. If IP is not authorized user must wait until the end.

I think wildcard is a better solution.

robhudson:
Looking at the django source code, `INTERNAL_IPS` is never used in this way and is assumed to be a list of IPs as a string:
https://github.com/django/django/search?utf8=%E2%9C%93&q=INTERNAL_IPS

My preference would be to treat it the same as Django does. If Django moves to allowing CIDR notation for `INTERNAL_IPS` then so should the toolbar.

Currently anyone can override this by making their own function and setting it with `SHOW_TOOLBAR_CALLBACK`. As such, my vote (if there were one) would be to not add this functionality.

All that said, I'm not maintaining the toolbar currently so I leave it up to whoever is. (I'm only here b/c I was at-mentioned)

tim-schilling:
I'm more inclined to leave this out and force devs to use `SHOW_TOOLBAR_CALLBACK`.

PR #786, opened 2015-12-05
Fix table display

virtosubogdan:
With small text content, the columns in the settings and request panels do not fill the whole screen.
Also the colors for the view row did not look consistent.


tim-schilling:
Can you show examples of what you're trying to fix?

virtosubogdan:
The display:table will turn this:
table_before
into this:
table_after
The row colors for the request panel are just for consistency.
From this:
request_panel_before
to this:
request_panel_after

PR #788, opened 2015-12-11
Add VCS info panel as a third-party panel

giginet:
I released third-party panel `django-rebug-toolbar-vcs-info` which enables to displays VCS information on the debug toolbar.

https://github.com/giginet/django-debug-toolbar-vcs-info

So I added it to the official document.

Please check it if you would like.


aaugustin:
Thanks!

PR #789, opened 2015-12-14
Update installation.rst

wjurkowlaniec:
Django admin isn't working if debug toolbar isn't at the end of INSTALLED_APS. I suggest to put it in docs.


aaugustin:
If that's true, it's a bug and we should fix it, not document it.

tim-schilling:
Closing this PR. @wjurkowlaniec Can you please open up an issue with the details of your environment so we can investigate this bug?

PR #791, opened 2015-12-18
Make ddt works with django-jinja 2.0.0

douglas:
This seems to fix django-debug-toolbar to work with django-jinja 2.0.0, it is also compatible with the django-jinja 1.4.2 =)


douglas:
Not needed anymore as the problem was fixed in django-jinja =)

PR #797, opened 2016-01-05
Update settings.py

dankram:
It's fairly standard that in Django urlconf's might be a tuple instead of a list. I think this function should take that into account.


tim-schilling:
You've reversed the order in which the `debug_urls` are added here. While I don't think the impact is that significant since the likelihood of someone using `__debug__/` to prefix a url pattern is low, this PR shouldn't be changing that.

Issue #799, opened 2016-01-11
Cache panel does not handle imported "cache" on module level


blueyed:
When using `from django.core.cache import caches` in a module, that is loaded only before a view is processed, this `caches.caches` instance will not be monkey-patched via `CacheHandlerPatch`.

One possible workaround would be to not use `enable_instrumentation` / `disable_instrumentation` for this, but do it in `debug_toolbar.panels.cache` on import time.

`enable_instrumentation` / `disable_instrumentation` could then be used still to enable/disable sending the signals, but `caches` would be wrapped always.

Brobin:
Any update on #800? I'm having this issue, which causes my cache clearing on save to break while using the toolbar.

```python
    from django.core.cache import cache
    ...

    def save(self, *args, **kwargs):
        self.slug = slugify(self.name)
        cache.clear('faq_sidebar')
        super(Topic, self).save(*args, **kwargs)
```

    File "./venv/lib/python3.5/site-packages/django/contrib/admin/options.py" in save_model
      991.         obj.save()
    File "./project/faq/models.py" in save
      35.         cache.clear('faq_sidebar')
    File "./venv/lib/python3.5/site-packages/debug_toolbar/panels/cache.py" in wrapped
      33.         value = method(self, *args, **kwargs)
    File "./venv/lib/python3.5/site-packages/debug_toolbar/panels/cache.py" in clear
      87.         return self.cache.clear(*args, **kwargs)
    Exception Type: TypeError at /admin/faq/topic/add/
    Exception Value: clear() takes 1 positional argument but 2 were given

minusf:
i am not sure if my problem is related to this, but i have noticed that using redis cache with the sitewide cache framework alwasys shows "0 calls in 0.0ms" on the cache panel.  django 3.1.1, django-redis 4.12.1, django-debug-toolbar 3.1.


Total calls | Total time | Cache hits | Cache misses
-- | -- | -- | --
0 | 0 ms | 0 | 0

mypseudo:
I have noticed that the "cache" panel does not work, it shows always 0 calls to cache whereas some are done.

billyquith:
> I have noticed that the "cache" panel does not work, it shows always 0 calls to cache whereas some are done. (django==3.2) But don't know if the problem is related to that.

I am noticing the same issue.

PR #800, opened 2016-01-11
Cache panel: monkey-patch caches at module level

blueyed:
Fixes https://github.com/django-debug-toolbar/django-debug-toolbar/issues/799.

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -8,7 +8,7 @@ from collections import OrderedDict
 import django
 from django.conf import settings
 from django.core import cache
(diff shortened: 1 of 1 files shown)


blueyed:
It's not early enough for celery though: it will have a different `caches` instance.

tim-schilling:
Can you add a test to verify the fix for #799 to avoid regressions in the future?

blueyed:
@tim-schilling 
Won't be easy probably, and I'm not sure if it's the right thing to do here.

matthiask:
This should probably be closed in favor of #982?

PR #804, opened 2016-01-27
Update documentation surrounding django.contrib.postgres.

tbeadle:


aaugustin:
Ideally, we'd fix this bug rather than document it...

camilonova:
@aaugustin you know if this is currently fixed?

aaugustin:
I'm not aware of a fix.

matthiask:
If that's still the case it would be better to add a check to `check_middleware`: https://github.com/jazzband/django-debug-toolbar/blob/master/debug_toolbar/apps.py

tim-schilling:
Closing this since I could not reproduce the message with Django 2.2, the lowest version the toolbar supports currently.

PR #805, opened 2016-01-29
add the 'CHECK_FOR_INTERNAL_IPS' configuration option

flo-dhalluin:
to disable internal ips check in the middleware. 

As more and more people (including me :-) ) are using virtualisation solution for dev (Docker, vagrant .. ) you might not know what to put in INTERNAL_IPS, and the  SHOW_TOOLBAR_CALLBACK option is difficult to understand for such a trivial problem. 
The new option is false by default, and if true will disable the IP address checks (so it has to be a conscious choice). The rationale is that if you are using docker to develop, the container is not public anyway.

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -24,8 +24,9 @@ def show_toolbar(request):
     """
     Default function to determine whether to show the toolbar on a given page.
     """
-    if request.META.get('REMOTE_ADDR', None) not in settings.INTERNAL_IPS:
-        return False
+    if dt_settings.CONFIG['CHECK_FOR_INTERNAL_IPS']:
+        if request.META.get('REMOTE_ADDR', None) not in settings.INTERNAL_IPS:
+            return False
 
     if request.is_ajax():
         return False
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -23,6 +23,7 @@ CONFIG_DEFAULTS = {
     'RESULTS_CACHE_SIZE': 10,
     'ROOT_TAG_EXTRA_ATTRS': '',
     'SHOW_COLLAPSED': False,
+    'CHECK_FOR_INTERNAL_IPS': True,
     'SHOW_TOOLBAR_CALLBACK': 'debug_toolbar.middleware.show_toolbar',
     # Panel options
     'EXTRA_SIGNALS': [],
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -124,6 +124,13 @@ Toolbar options
   request must not be an AJAX request. You can provide your own function
   ``callback(request)`` which returns ``True`` or ``False``.
 
+* ``CHECK_FOR_INTERNAL_IPS``
+
+  Default: ``True``
+
+  Checks that the IP of the client in the request is in ``INTERNAL_IPS`` in the default
+  toolbar callback. (INTERNAL_IPS default to ``("127.0.0.1", "::1")`` )
+  
 Panel options
 ~~~~~~~~~~~~~
 


tim-schilling:
Are you aware of `SHOW_TOOLBAR_CALLBACK`? It exists for developers to override the default check of whether or not to enable the toolbar.

flo-dhalluin:
Yes I am aware of SHOW_TOOLBAR_CALLBACK, but it feels tedious and confusing to have to write a callback function just to disable the IP address checks. Also then you need to dig in the code if you want to retain the "no ajax" behavior for injection. 

Maybe there could be an alternative callback implementation supplied something like "always_show_toolbar" ? And maybe make it more explicit in the "Installation" that toolbar will only work for local ips by defaut (I know it's mentionned, but at the bottom ... maybe in the "note" in the quick setup section ? 
the Debug toolbar is a magnificent tools, but this is a caveat that may discourage new comers, especially as more and more people turns to solution such as docker for development.

This is a question often asked on StackOverflow.. 
http://stackoverflow.com/questions/10517765/django-debug-toolbar-not-showing-up
http://stackoverflow.com/questions/28984239/cannot-get-django-debug-toolbar-to-appear?lq=1
... 

Just my 2cts, thanks for the awesomness anyway !

nwaxiomatic:
In case anyone else is having this issue, the fastest way to solve this if running Django on a remote server is to follow this answer:
http://stackoverflow.com/a/10518040/4237114
But edit the code slightly to be
```
def show_toolbar(request):
    print request.META.get('REMOTE_ADDR')
    return True
DEBUG_TOOLBAR_CONFIG = {
    "SHOW_TOOLBAR_CALLBACK" : show_toolbar,
}
```
Then restart server (gunicorn, runserver, w/e) go to some page on your check your log file for the IP address you need to add printed out nicely.

matthiask:
We shouldn't encourage DDT usage on IP addresses not in INTERNAL_IPS. My opinion is that exposing DDT *should* be harder than changing a single boolean.

PS1. (If you want a dirty trick which works for development in your LAN, add `INTERNAL_IPS = type(str('c'), (), {'__contains__': lambda *a: True})()` to your settings.)

PS2. Reopen if you disagree strongly.

edmorley:
What is the risk that the current implementation is trying to avoid?

I presume it's for users who are using `DEBUG=True` and exposing that machine either:
a) intentionally (eg shared office devbox, but not realising the risk)
b) accidentally (eg not knowing the difference between binding to 0.0.0.0 rather than 127.0.0.1)

For group (a), I would posit that they are going to just use a hack like the one given above even if there isn't a setting, unless it's easy for them to eg use wildcards to whitelist a LAN.

For group (b), if the django-debug-toolbar docs explain using INTERNAL_IPS and warn very strongly against disabling the checks, then I don't think adding a setting will increase the risk.

Ultimately I agree it's important to have a safe out of the box config, plus also protect users who make simple mistakes (like using `DEBUG=True` in production), but there's a point after which I think that legitimate users (those using Vagrant/docker etc for development) should take priority over users that are likely going to do the wrong thing regardless.

It's also worth noting that the INTERNAL_IPS feature offers no protection if the user has bungled their apache/nginx/varnish reverse proxy settings, since all IPs will appear as 127.0.0.1 to the backend anyway. In fact this is the only reason I've come across this PR, since I'm removing the reverse proxy from our Vagrant development config, which was previously making all requests appear to have come from 127.0.0.1.

ricardodani:
@edmorley I strongly agree

I don't see a real meaning of testing the INTERNAL_IPS either. This looks just paranoid to me.

ricardodani:
+1 to this PR

matthiask:
I don't see the point in adding an untested configuration option to a three-line method, which can be easily replaced as a whole.

Also, if you think `INTERNAL_IPS` does not make much sense please take this to the Django developers' mailing list. If Django drops the `INTERNAL_IPS` check for its `{% debug %}` template tag and context variable, I'm of course happy to drop the check in django-debug-toolbar as well.

PR #808, opened 2016-02-21
Make debug_toolbar.settings import safe.

### LINKED ISSUE #801: Make defaults importable without side-effects
I am trying to use / add to `debug_toolbar.settings.PANELS_DEFAULTS` with django-configurations, but it cannot be imported during setting up the settings:

    @classmethod
    def setup(cls):
        super().setup()
        from debug_toolbar import settings as dt_settings
        cls.DEBUG_TOOLBAR_PANELS = (
            ['debug_toolbar.panels.profiling.ProfilingPanel'] +
            dt_settings.PANELS_DEFAULTS)

The problem seems to be with the following code:

    USER_CONFIG = getattr(settings, 'DEBUG_TOOLBAR_CONFIG', {})

    (from https://github.com/django-debug-toolbar/django-debug-toolbar/blob/8f75a7b980d150fc554053d76e0438edd6e40f02/debug_toolbar/settings.py#L42)

It triggers the following exception:

> django.core.exceptions.ImproperlyConfigured: The SECRET_KEY setting must not be empty.

It would be good if `PANELS_DEFAULTS` (and possibly other defaults) could be imported without side-effects.

aaugustin:
Fix #801.

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -11,6 +11,6 @@ class DebugToolbarConfig(AppConfig):
     verbose_name = _("Debug Toolbar")
 
     def ready(self):
-        if dt_settings.PATCH_SETTINGS:
+        if dt_settings.get_patch_settings():
             dt_settings.patch_all()
         dt_settings.check_middleware()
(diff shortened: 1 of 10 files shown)


aaugustin:
I'm waiting for testing or feedback on issue #801 to merge this PR.

blueyed:
:+1: 
Works for me: https://github.com/django-debug-toolbar/django-debug-toolbar/issues/801#issuecomment-222962370

PR #809, opened 2016-02-21
Django 1.9 is not "upcoming" anymore

svetlyak40wt:
This pull fixes a changelog, which states that 1.4 debug toolbar is compatible with "upcoming" django 1.9.


aaugustin:
Thanks!

PR #828, opened 2016-05-20
Added test target for Django 1.10

claudep:


claudep:
I'm not there yet, Django <1.11 tells: Requirement already satisfied. And Django >=1.10 doesn't take 1.10.a1 into account...

claudep:
Seems "Django<1.11,>=1.10a" works. Is it the right expression?

MarkusH:
You want to set `pip_pre=True` in tox: http://testrun.org/tox/latest/config.html#confval-pip_pre=True|False(default)

claudep:
I think I'm satisfied with the PR now!

PR #833, opened 2016-05-27
Monkey patch Django _create_cache method

lovemyliwu:
cacheback used this method to create cache instance


aaugustin:
What in cacheback?

lovemyliwu:
_create_cache  used in Django cache back https://github.com/DistPub/django-cacheback/blob/39108778fccbc353af52a430c5ba7f7fd0edcade/cacheback/utils.py#L40

lovemyliwu:
There are some failure in UT, anybody help?

aaugustin:
I don't think the debug toolbar should encourage using private APIs of Django.

lovemyliwu:
alright, Thanks.

PR #837, opened 2016-06-07
:bug: fix repeat record on cache panel

lovemyliwu:


aaugustin:
How can this bug be triggered?

lovemyliwu:
from some template tags logic

aaugustin:
I'm afraid it's going to be hard to reproduce the bug, let alone write a test case, based on the information you provided.

lovemyliwu:
sorry, I found the issue from django cms tags, hard to say the root cause why proxy class wrapper on a proxy class.

Issue #844, opened 2016-06-24
Raw query execution issue with Django 1.9


valentjedi:
Hi! Don't sure if it's in Django 1.9 only.
So let's have a code with custom query execution:
```python 
with connection.cursor() as c:
    c.execute(query) 
```
Then we've got an exception:
```
/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
IndexError: tuple index out of range
```

The problem is that in django code execution signature is:
```python
def execute(self, sql, params=None):
    self.db.validate_no_broken_transaction()
    with self.db.wrap_database_errors:
        if params is None:
            return self.cursor.execute(sql)
        else:
            return self.cursor.execute(sql, params)
```
And in debug-toolbar is:
```python
 def execute(self, sql, params=()):
     return self._record(self.cursor.execute, sql, params)
```

Also same issue goes with `callproc` method

I'll also make PR to fix this if you don't mind :)

PR #845, opened 2016-06-24
#844 fix tuple index out of range error

valentjedi:
See #844


aaugustin:
That's how Django wraps execute and callproc as well, so let's make this change.

PR #848, opened 2016-07-11
Add JAVASCRIPT_CUSTOM_ATTRIBUTE configuration

sebatyler:
When I use django-debug-toolbar with custom JQUERY_URL, I need to add custom attribute to script element.
I want to use defer or async attribute.

So I added JAVASCRIPT_CUSTOM_ATTRIBUTE configuration.

### DIFF
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -37,7 +37,8 @@ CONFIG_DEFAULTS = {
     ),
     'PROFILER_MAX_DEPTH': 10,
     'SHOW_TEMPLATE_CONTEXT': True,
-    'SQL_WARNING_THRESHOLD': 500,   # milliseconds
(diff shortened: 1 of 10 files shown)


aaugustin:
Why do you need this?

sebatyler:
Now I am enhancing client-side performance of my site. So I need to include jquery with defer loading.

aaugustin:
This doesn't make sense to me.

The debug toolbar isn't for end users; it's just for developers.

The overhead of a development environment — just running with DEBUG enabled and with the toolbar installed, I'm not even talking about frontend — will dwarf the cost of loading jQuery on localhost.

aaugustin:
Closing because I'm strongly opposed to anything that gives the impression that it's feasible to run with the toolbar enabled in production, and this level of optimization clearly does.

PR #855, opened 2016-07-20
Fix #852, anyway this parameter never used

### LINKED ISSUE #852: Appears to be incompatible with sqlparse version 0.2.0.
I get the following exception and stack trace on every page I try to view:

```
Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  223.                 response = middleware_method(request, response)
File "/usr/local/lib/python2.7/dist-packages/debug_toolbar/middleware.py" in process_response
  129.                 panel.generate_stats(request, response)
File "/usr/local/lib/python2.7/dist-packages/debug_toolbar/panels/sql/panel.py" in generate_stats
  192.                     query['sql'] = reformat_sql(query['sql'])
File "/usr/local/lib/python2.7/dist-packages/debug_toolbar/panels/sql/utils.py" in reformat_sql
  27.     return swap_fields(''.join(stack.run(sql)))
File "/usr/local/lib/python2.7/dist-packages/sqlparse/engine/filter_stack.py" in run
  29.             stream = filter_.process(stream)

Exception Type: TypeError at /admin/mediasources/document/
Exception Value: process() takes exactly 3 arguments (2 given)

```

I can workaround by forcing the version to 0.1.19 via 
sqlparse==0.1.19
in the requirements file.

sqlparse 0.2.0 was released today

valentjedi:
Fix for new process api of sqlparse.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ setup(
     packages=find_packages(exclude=('tests.*', 'tests', 'example')),
     install_requires=[
         'Django>=1.8',
-        'sqlparse',
+        'sqlparse>=0.2.0',
     ],
     include_package_data=True,
     zip_safe=False,                 # because we're including static files
--- a/debug_toolbar/panels/sql/utils.py
+++ b/debug_toolbar/panels/sql/utils.py
@@ -9,7 +9,7 @@ from sqlparse import tokens as T
 
 class BoldKeywordFilter:
     """sqlparse filter to bold SQL keywords"""
-    def process(self, stack, stream):
+    def process(self, stream):
         """Process the token stream"""
         for token_type, value in stream:
             is_keyword = token_type in T.Keyword


setup.py
```diff
@@ -16,7 +16,7 @@
     packages=find_packages(exclude=('tests.*', 'tests', 'example')),
     install_requires=[
         'Django>=1.8',
-        'sqlparse',
+        'sqlparse>=0.2.0',
```
webknjaz: I'd add upper limit as well:
```python
        'sqlparse >=0.2.0,<0.3',
```
This will help to avoid similar issues with unexpected API changes.

valentjedi:
@webknjaz I'd rather not. Why lock ourselves to old version of other packages? We can also add upper boundary for Django that way. Let's say, <1.10.

webknjaz:
@valentjedi it will still allow to patch version upgrades. It looks they don't follow semver, introducing backwards incompatible changes with minor version upgrade. So only upgrading to patch version change is safe.
On the other hand, there are some services (like require.io), which help keep track on outdated dependencies.
It's better to maintain dependencies manually, preventing users from having unexpected and unpredictable app failures like this one.

setup.py
aaugustin: I don't think sqlparse strictly adheres to semantic versioning and I'm afraid we'll forget to update this when 0.3 gets release, so I prefer the current version.

webknjaz: @aaugustin 
https://requires.io/ could be used to track dependencies status. They provide a badge, which can be included in `README.md`, so if some package gets outdated it changes background from green to yellow or red alerting the issue.
They also list all packages with their versions like this: https://requires.io/github/celery/celery/requirements/?branch=5.0-devel
This way, it's easy to mention one requiring update.
There's also an option to be notified about outdated packages by email and/or pull request.

webknjaz: I'll submit a couple PRs suggesting to add useful badges if you don't mind

aaugustin: Since we don't set upper limits, how could our requirements get out of date?

(Also I'm not a fan of cargo-culting badges.)

aaugustin: (And I'm especially not a fan of shaming badges that become red if I don't look at the repo for a month or two.)

webknjaz: Well it makes sense to consider changing personal preferences given that there's a tool to help handle this. The badge motivates to keep it green, but it's not a must. I believe, automatic notifier does not require adding badge explicitly to work.
Still it's up to you to decide whether to agree with this point.

PR #861, opened 2016-07-24
changes view function to render

auvipy:


aaugustin:
The reason why this line doesn't use `render` is explained in a comment on the previous line.

auvipy:
since render_to_response is now depricated what should be done now?

PR #868, opened 2016-08-12
Add support django 1.10 middleware

digitalashes:
Hook for support django 1.10 middleware


matthiask:
Thanks! The fix is mostly the same as in #864, but still with support for the automatic setup. If automatic setup were removed we should move forward with the simpler PR.

(I sent a message to django-users@, maybe chime in if you have an opinion regarding the automatic setup.)

matthiask:
I've merged #864 instead, therefore this PR is now obsolete. 

Still, thanks for the contribution!

PR #876, opened 2016-09-12
resolve error with django-jiinja templates

### LINKED ISSUE #867: Conflict with django_jinja
We upgraded from django-debug-toolbar 1.4 to 1.5 and ended up with an error that seems to break compatibility with django_jinja.

We received the same error as this Stack Overflow question:
http://stackoverflow.com/questions/38569760/django-debug-toolbar-template-object-has-no-attribute-engine

Line 175 in `panels.templates.panel.py` raises:

``` python
Exception Value: 'Template' object has no attribute 'engine'
```

If you look at the asker's written solution, you'll notice that they patched line 175 by checking for the `engine` attribute on `self.templates[0]['template']` or to alternately look for a `backend` attribute (which exists and fulfills the same function in Jinja2).

This seems to have it's origin in the way that template dirs are grabbed, by pulling the engine info from an individual template:

``` python
        # Fetch context_processors/template_dirs from any template
        if self.templates:
            context_processors = self.templates[0]['context_processors']
            template_dirs = self.templates[0]['template'].engine.dirs
        else:
            context_processors = None
            template_dirs = []
```

In the following linked commit, you can see that debug_toolbar used to have a `get_template_dirs()` function that used `django.template.engine.Engine.get_default()` to get an engine and `engine.dirs` to pull in the template dirs. I haven't tested based on this commit, but I suspect that removing the call to `django.template.engine.Engine.get_default()` and pulling template dirs from an individual template caused the breaking change:
https://github.com/jazzband/django-debug-toolbar/commit/6390279a5eba1dc47fc12627204484fda67d3825#diff-f6bb86ee9199645f22f9b23d002ed707L24

The following seems to work and I would assume has a lower chance of breaking compatibility with other template engines.

``` python
from django.template.engine import Engine
engine = Engine.get_default()
engine.dirs # returns [] in my case
```

We encountered this using:

```
Django==1.10 (also tried with 1.9.9)
django_jinja==2.2.0
Jinja2==2.8
django-debug-toolbar==1.5
```

samuelcolvin:
fix #867

Let me know if you want tests and if so how you want them configured.

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -6,7 +6,7 @@
 
 Django
 sqlparse
-Jinja2
+django_jinja
 
 # Testing
 
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -6,6 +6,7 @@
 
 Django
 sqlparse
+Jinja2
 
 # Testing
 
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -1,16 +1,66 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.apps import AppConfig
+from django.conf import settings
+from django.core.checks import Error, register
+from django.middleware.gzip import GZipMiddleware
+from django.utils.module_loading import import_string
 from django.utils.translation import ugettext_lazy as _
 
-from debug_toolbar import settings as dt_settings
+from debug_toolbar.middleware import DebugToolbarMiddleware
 
 
 class DebugToolbarConfig(AppConfig):
     name = 'debug_toolbar'
     verbose_name = _("Debug Toolbar")
 
-    def ready(self):
-        if dt_settings.get_patch_settings():
-            dt_settings.patch_all()
-        dt_settings.check_middleware()
+
+@register
+def check_middleware(app_configs, **kwargs):
+    errors = []
+    gzip_index = None
+    debug_toolbar_index = None
+
+    setting = getattr(settings, 'MIDDLEWARE', None)
+    setting_name = 'MIDDLEWARE'
+    if setting is None:
+        setting = settings.MIDDLEWARE_CLASSES
+        setting_name = 'MIDDLEWARE_CLASSES'
+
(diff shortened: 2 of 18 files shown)


matthiask:
Since I don't use django-jinja there is a (big) chance it'll break again without tests. I'm wondering whether this problem also happens with Django's bundled jinja2 template backend, and if it doesn't, why not?

pbaranay:
I've proposed some tests in @samuelcolvin's fork, [here](https://github.com/samuelcolvin/django-debug-toolbar/pull/1).

samuelcolvin:
Ok, i've checked with selenium tests enabled and that's now working, however travis isn't running selenium tests so it could easily fail again. For example 94d91de passed on travis but failed with selenium enabled.

Can I suggest selenium tests are enabled for travis, it might be slightly slow but I think it would be worth it.

samuelcolvin:
bump @matthiask, would be great to fix this and have a new release.

requirements_dev.txt
```diff
@@ -6,6 +6,7 @@
 
 Django
 sqlparse
+Jinja2
```
matthiask: Shouldn't this be `django_jinja` as well?

samuelcolvin: sorry, fixed.

matthiask:
This looks mostly good to me.

That being said, we're not using Jinja, and I therefore would very much appreciate a review by @aaugustin if you have the time? Or someone else who knows more about template engines than I do :)

I'll gladly make a release afterwards.

samuelcolvin:
of course.

Perhaps the best person to ask would be @niwinz if he had a chance to look at it as he built django-jinja and therefore understands it best.

aaugustin:
Indeed Django-Jinja predates official support for Jinja2 in Django and operates quite differently, so I don't expect my knowledge of the latter to be helpful.

The patch looks fairly reasonable. I wish the debug toolbar used standard APIs exposed by Django to access debug information in templates but we're far from that.

PR #883, opened 2016-09-27
use HMAC and timing safe comparison

zt-initech:
Instead of naive code vulnerable to length-extension attack and instead of short-circuiting string comparison.

https://en.wikipedia.org/wiki/Hash-based_message_authentication_code#Design_principles

https://docs.python.org/2/library/hmac.html#hmac.compare_digest

### DIFF
--- a/debug_toolbar/panels/sql/forms.py
+++ b/debug_toolbar/panels/sql/forms.py
@@ -8,8 +8,8 @@ from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import connections
-from django.utils.encoding import force_text
 from django.utils.crypto import constant_time_compare
+from django.utils.encoding import force_text
 from django.utils.functional import cached_property
 
 from debug_toolbar.panels.sql.utils import reformat_sql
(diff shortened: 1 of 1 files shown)


aaugustin:
I'm not a fan of changes that suggest that it might be feasible, let alone a workable idea, to run the debug toolbar in untrusted environments. But people will do it anyway, and defense in depth is a good thing. For this reason, I'm in favor of making these changes.

Travis CI says the PR doesn't work on Python 3.

zt-initech:
@aaugustin I think there are two possible states that make sense: either an authentication tag is not produced and is not checked, or it is produced securely and checked securely. The current situation in which an authentication tag is produced insecurely and then checked insecurely does not make sense. The code needs to change to be either in state 1 or 2. I prefer state 2, but state 1 is also good, because it emphasizes that debug toolbar is only for trusted environments.

matthiask:
Thanks!

aaugustin:
Yep. It's best to follow best-practices regardless of the context, since the context may change or not be what we expect. Thanks for the patch.

PR #896, opened 2016-11-23
Pass through extra cursor arguments.

furious-luke:
Hi there. We have a server-side cursor implementation for PostgreSQL and Django and it requires us to pass in the name of the cursor to the `.cursor` method on connections. The `wrap_cursor` wrapper doesn't pass through arguments by default, so I made a very minor change to do so.


matthiask:
That's the same change as https://github.com/jazzband/django-debug-toolbar/pull/615. Maybe we can continue the discussion there? Thanks!

PR #897, opened 2016-11-25
Log a warning in process_response when skipping for gzip

blueyed:
TODO:
 - [ ] test
 - [ ] only log once?

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -4,6 +4,7 @@ Debug Toolbar middleware
 
 from __future__ import absolute_import, unicode_literals
 
+import logging
 import re
 import threading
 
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks!

We already have a system check that shows an error if the DDT middleware comes before the GZipMiddleware[1]. I don't see any runtime logging for possible runtime configuration issues in DDT yet, and am leaning towards rejecting this change because I doubt that the gain is worth the complexity (not much complexity, but no gain).

[1]: https://github.com/jazzband/django-debug-toolbar/blob/master/debug_toolbar/apps.py#L52

blueyed:
Fair point.
I have not checked if the system check works in the same way, but assuming that let's close this.

PR #904, opened 2017-01-04
Prefix timeline and query css classes with "djdt" to avoid conflicting class names

### LINKED ISSUE #903: Debug toolbar table CSS classnames missing djdt prefix
I had my own css class called `.timeline` and it conflicted with the SQL query timeline such that it overflowed the page. It took me quite a bit to find that the root cause was conflicting class names. Perhaps it should have a debug toolbar prefix like `.djdt-timeline`?

I noticed this is in place for some, but not all of the `th` classes. Adding it to `.query` and `.timeline` would probably be best for consistency.

```html
<tr>
    <th class="djdt-color">&nbsp;</th>
    <th class="query" colspan="2">Query</th>
    <th class="timeline">Timeline</th>
    <th class="djdt-time">Time (ms)</th>
    <th class="djdt-actions">Action</th>
</tr>
```

Brobin:
Resolves #903

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -503,7 +503,7 @@
 #djDebug .highlight .s2 { color:#333 } /* Literal.String.Double */
 #djDebug .highlight .cp { color:#333 } /* Comment.Preproc */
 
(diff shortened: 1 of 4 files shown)


matthiask:
Makes sense, thank you!

PR #913, opened 2017-02-02
Use Django's import_string so as to avoid re-implementing it

jdufresne:


matthiask:
Thanks! I just checked whether `import_string` is documented, and indeed it is.

PR #916, opened 2017-02-24
Bullet-proof the test for window.define in jquery_pre.js and ..._post.js

KimSJ:
The existing test `if (window.define)` could fail in obscure circumstances where someone is using `window.define` for another purpose. The suggested fix reduces the chance of that significantly.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/jquery_post.js
+++ b/debug_toolbar/static/debug_toolbar/js/jquery_post.js
@@ -1,4 +1,4 @@
 var djdt = {jQuery: jQuery.noConflict(true)};
-if (window.define) {
+if (typeof window.define is "function" && window.define.amd) {
     window.define.amd = _djdt_define_amd_backup;
 }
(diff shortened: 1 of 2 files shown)


matthiask:
Sounds like a debugging session from hell. Unfortunately I don't really have any experience with AMD/define and can't judge whether this pull request is correct (I went from inline everything directly to es6 modules...)

Are any other projects using a similar pattern? Or is there maybe even a best practice for this sort of thing? I get the impression that DDT had many problems in this area from browsing the issues list...

aaugustin:
These aren't good memories :-|

You can merge and pray...

matthiask:
I don't intend to cause anyone more work, and I'm not exactly bored myself either. Still waiting for additional feedback, but I'm leaning towards rejecting this change.

camilonova:
I'll reject this change since is safer for the general use of this package.

@KimSJ Thank you

PR #921, opened 2017-03-09
Fixed #920 Version representation

### LINKED ISSUE #920: Incorrectly parse versions of some apps
I saw that for `django-celery-beat` and `django-celery-results` it shows extra dots at the end of the version:
```
django_celery_beat	Django_Celery_Beat	1.0.1..
django_celery_results	Django_Celery_Results	1.0.1..
```

This is because both apps use custom `namedtuple`:
```
version_info_t = namedtuple('version_info_t', (
    'major', 'minor', 'micro', 'releaselevel', 'serial',
))
```

When `releaselevel` and `serial` are not set they are empty strings. This will lead to joining dots in the string representation.

I'm preparing a fix and will create a PR.

vstoykov:
Fixed #920

### DIFF
--- a/debug_toolbar/panels/versions.py
+++ b/debug_toolbar/panels/versions.py
@@ -54,5 +54,7 @@ class VersionsPanel(Panel):
         else:
             return
         if isinstance(version, (list, tuple)):
-            version = '.'.join(str(o) for o in version)
+            # We strip dots from the right because we do not want to show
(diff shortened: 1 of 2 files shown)


matthiask:
Thanks, this is almost obviously correct, but a testcase would make this PR even better!

vstoykov:
Ok test case added.

matthiask:
Thanks!

PR #924, opened 2017-03-26
Consistently use HTML5 self closing tag syntax

jdufresne:
Was previously a mix of "`/>`", "` />`", and "`>`".

As a result of using HTML5 syntax, removed XML test.


aaugustin:
I think it would be more useful to normalise towards closing tags, so the toolbar remains usable in websites that render XHTML with a strict doctype. This will still be valid HTML5.

jdufresne:
Makes sense. I've updated the PR to take this approach.

PR #927, opened 2017-03-27
Fix #902: [SQL panel] display of included/inherited template source

### LINKED ISSUE #902: SQL panel points a query in an included template to the same line of its parent template
Tested with:

views.py
````python

def index(request):
````

index.html
````html
{% extends 'base.html' %}

{% block content %}

{% include 'inner.html' %}

{% endblock content %}
````

inner.html
````html
<p>Here's inner.html:</p>


<p>All questions:</p>
{{ questions }}
````

image

lingxiaoyang:
Changes proposed:

1. [Documentation] Clarify the Django version support of `get_template_source_from_source` and `get_template_source_from_exception_info` functions.
2. [Fix] Port the bugfix in Django 1.11 branch regarding the retrieval of template source. (Ref: [Django ticket #27584](https://code.djangoproject.com/ticket/27584) and [Django ticket 27956](https://code.djangoproject.com/ticket/27956) - incorrect retrieval of `{% extends %}` and `{% include %}` tags). Ported code: https://github.com/django/django/blob/stable/1.11.x/django/template/base.py#L960

### DIFF
--- a/debug_toolbar/utils.py
+++ b/debug_toolbar/utils.py
@@ -116,7 +116,7 @@ def get_template_info():
 
 def get_template_context(node, context, context_lines=3):
     source = getattr(node, 'source', None)
-    # In Django 1.9 template Node does not have source property, Origin does
+    # From Django 1.9 template Node does not have source property, Origin does
     # not reload method, so we extract contextual information from exception
     # info.
     if source:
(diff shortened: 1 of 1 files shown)


matthiask:
Would constructing a testcase for this be too hard? As we're doing more than Django does by fixing 1.9 and 1.10 I'd sleep better with a simple test case (see also the discussion on #902 )

camilonova:
@lingxiaoyang Can you please take a look?

matthiask:
This might be obsolete since we dropped support for Django<1.11.

Please reopen if you disagree. Thanks!

PR #942, opened 2017-05-04
Refs #910: Skip template-based widget templates by default

matthiask:


debug_toolbar/settings.py
```diff
@@ -35,6 +35,11 @@
     ),
     'PROFILER_MAX_DEPTH': 10,
     'SHOW_TEMPLATE_CONTEXT': True,
+    'SKIP_TEMPLATE_PREFIXES': (
+        'debug_toolbar/',
+        'django/forms/widgets/',
+        'admin/widgets',
```
kezabelle: Missing trailing slash. If it's intentionally omitted, the docs change is wrong instead :)

debug_toolbar/panels/templates/panel.py
```diff
@@ -85,7 +85,8 @@ def _store_template_info(self, sender, **kwargs):
 
         # Skip templates that we are generating through the debug toolbar.
         if (isinstance(template.name, six.string_types) and
-                template.name.startswith('debug_toolbar/')):
```
kezabelle: By making this configurable, it becomes possible to unsilence (accidentally by deleting the wrong line in your own `SKIP_TEMPLATE_PREFIXES`, or intentionally) the djdt templates themselves. 
Presumably, it's never been requested to be able to do so (or has been rejected for introducing too much configuration when it wasn't previously potentially necessary) so is it worth keeping the hard-coded extant functionality and making the `SKIP_TEMPLATE_PREFIXES` an *additional* set of skips?

goblinJoel:
This looks like a simple, effective fix, if not as fancy as de-duplicating or lazy loading (which would allow users to still see skipped templates and contexts if they needed to).

debug_toolbar/panels/templates/panel.py
goblinJoel: I think I agree with @kezabelle . It's probably going to cause fewer headaches for people if we keep the 'debug_toolbar/' skip hard-coded instead of requiring users who customize the list to include it.

On the other hand, if it makes developing and debugging the toolbar itself easier, I don't think it's a big deal, as long as the default value is clearly documented.

debug_toolbar/settings.py
matthiask: Thanks! I meant to include the trailing slash.

debug_toolbar/panels/templates/panel.py
matthiask: Good point. We can do that later should the need arise. I changed this.

PR #945, opened 2017-05-12
Escape stats output in profiling panel

jdufresne:
A time consuming list comprehension is rendered as "`<listcomp>`". As this
contains angular brackets, it should be escaped to be valid HTML.


matthiask:
Thanks!

PR #953, opened 2017-06-08
tip for toolbar in docker

DevRo5000:
Added tip for toolbar not being enabled when django is run in a docker container


SalahAdDin:
@DevRo5000 Is not correctly, each container have a different gateway, instead this put it like and example, i did:
```
INTERNAL_IPS = ['::1', '172.18.0.1']
# Modify second value accord your Docker configuration:
#    docker inspect uzman.web | grep -e '"Gateway"'
```

Issue #954, opened 2017-06-08
SHOW_TOOLBAR_CALLBACK change in django-debug-toolbar 1.8 breaks existing code and it is not documented


glasslion:
In django-debug-toolbar 1.7 and below, the `SHOW_TOOLBAR_CALLBACK` callback is used to check both whether a request  to  debug-toolbar views should be authorized and whether a request is ajax.

In django-debug-toolbar 1.8, the ajax check is moved into the middleware, and it raises a 404 if a request can not pass the SHOW_TOOLBAR_CALLBACK check.

So a callback written for django-debug-toolbar<=1.7 will broken debug-toolbar 1.8's ajax request. :-(

Debug-toolbar's own doc still indicates that the callback should check if the request is AJAX, it's confusing：

> The default checks are that DEBUG must be set to True, the IP of the request must be in INTERNAL_IPS, and the request must not be an AJAX request. 

Because those 404s do not have explicit error messages, it is even harder to find the root cause.

matthiask:
Cc @jdufresne you might want to take a look at this? Does the PR make sense to you?

awbacker:
In my case I don't want to do the `internal_ips` check since we work in ways where thats really not possible.  I've overridden the callback to check debug + !is_ajax(), and that was broken now (not sure why).  This changed fixed it for me:

```
-            return bool(base_settings.DEBUG) and not request.is_ajax()
+            return bool(base_settings.DEBUG)
``

As for changes to message, I would stick with the 404 from the resource, however... The calling panel obviously can render html, so it could:

- detect that you are using a custom callback and ask you to check that
- say `404` but give a list of possible reasons

Just a thought.  This was quite frustrating, as @glasslion says.  Since there is nothing that indicates the possible causes of the error *and* if you copy the failing url into your browser directly it works!  So somehow the

jdufresne:
> and that was broken now (not sure why)

The behavior changed so that `SHOW_TOOLBAR_CALLBACK` could be reused to guard against unauthorized access of panel views that are requested with AJAX. So the AJAX check is not always applicable. The details of this are outline in PR #926.

It was **my mistake** not documenting this breaking change, but the net effect is that panel views now check access is allowed to avoid leaking private data, which I think should remain.

I think modifying the documentation to alert users to the breaking changes is the best way to handle it without reintroducing unauthorized access of panel views.

PR #955, opened 2017-06-08
raise Http404 with error messages in the require_show_toolbar and fix…

glasslion:
This PR provides an hint messages when raises 404 , and fix the outdated/misleading doc. See #954

### DIFF
--- a/debug_toolbar/decorators.py
+++ b/debug_toolbar/decorators.py
@@ -10,7 +10,11 @@ def require_show_toolbar(view):
 
         show_toolbar = get_show_toolbar()
         if not show_toolbar(request):
-            raise Http404
+            raise Http404(
+                'You do not have the permission to access debug-toolbar'
+                ' urls. Please check your INTERNAL_IPS and'
+                ' SHOW_TOOLBAR_CALLBACK configurations'
+            )
 
         return view(request, *args, **kwargs)
     return inner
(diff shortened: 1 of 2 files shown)


debug_toolbar/decorators.py
```diff
@@ -10,7 +10,11 @@ def inner(request, *args, **kwargs):
 
         show_toolbar = get_show_toolbar()
         if not show_toolbar(request):
-            raise Http404
+            raise Http404(
+                'You do not have the permission to access debug-toolbar'
+                ' urls. Please check your INTERNAL_IPS and'
+                ' SHOW_TOOLBAR_CALLBACK configurations'
```
jdufresne: The idea behind raising a 404 instead of a 403 is that unprivileged users shouldn't even be aware of Django debug toolbar's existence. From the user's perspective, the resource "doesn't exist" instead of "the resource exists, but you aren't authorized to access it". This message looks to be leaking this information now. I think it shouldn't.

jdufresne:
@glasslion, I'm not sure if you're still working on this. I've proposed an alternative PR in #961. If you prefer to continue with this, I can close mine, no problem. If you want to reuse the wording I've added to the change log, also no problem.

matthiask:
Closing in favour of the merged #961.

PR #961, opened 2017-06-29
Documented backwards incompatible change in 1.8

### LINKED ISSUE #954: SHOW_TOOLBAR_CALLBACK change in django-debug-toolbar 1.8 breaks existing code and it is not documented

jdufresne:
SHOW_TOOLBAR_CALLBACK no longer returns False for AJAX requests. Custom values shouldn't either.

Fixes #954

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -15,6 +15,16 @@ Bugfixes
 This version is compatible with Django 1.11 and requires Django 1.8 or
 later.
 
+**Backwards incompatible changes**
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* ``debug_toolbar.middleware.show_toolbar`` (the default value of setting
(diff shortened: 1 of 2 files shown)


matthiask:
Thanks!

PR #993, opened 2017-09-23
Allow tests to pass with Django 2.0

jdufresne:
Django 2.0 has removed the deprecated setting `MIDDLEWARE_CLASSES`. Use `MIDDLEWARE` instead.

As a result, adjust the the expected warnings during test as system check warning `1_10.W001` was removed.

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -30,7 +30,7 @@ INSTALLED_APPS = [
 
 MEDIA_URL = '/media/'   # Avoids https://code.djangoproject.com/ticket/21451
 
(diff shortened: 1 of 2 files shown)


jdufresne:
Given the [Django support schedule](https://www.djangoproject.com/download/#supported-versions) and the [deprecation policy](https://docs.djangoproject.com/en/dev/internals/release-process/#deprecation-policy), would it make sense to drop support for Django < 1.11 when adding support for 2.0? I can update this PR to drop all deprecated support if you agree it is a good idea.

PR #1013, opened 2017-11-16
Add dates to all releases in changes.rst

jdufresne:
Updated `docs/contributing.rst` to date the release as part of the release process.

Useful when scanning history for the evolution of the project over time.

Follows practices recommended by http://keepachangelog.com/

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,13 +1,13 @@
 Change log
 ==========
 
-UNRELEASED
-----------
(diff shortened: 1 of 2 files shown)


matthiask:
Thanks! I'm generally leaning towards not documenting patch releases too much because of the additional workload it causes, but release dates are definitely a good idea! ([example](https://github.com/matthiask/django-content-editor/blob/master/CHANGELOG.rst))

PR #1018, opened 2017-12-17
Drop support for Python 3.3

michael-k:
It reached its EOL in September 2017
https://www.python.org/dev/peps/pep-0398/#x-end-of-life


matthiask:
Thank you! Does this change enable any cleanups in the code, or allow us to make parts of the code more idiomatic? 

If there are no additional benefits to this change I'm -0 on it.

michael-k:
> Does this change enable any cleanups in the code, or allow us to make parts of the code more idiomatic?

None I'm aware of.

matthiask:
Thank you! 

This has been made obsolete by #1020 (dropping compatibility with Django 1.8, and therefore also with older Python versions such as 3.3).

PR #1019, opened 2017-12-22
Fix missing query-actions buttons for queries with date/datetime params.

### LINKED ISSUE #882: support params in sql panel that cannot be JSON encoded by default JSONEncoder
If sql params are values that cannot be encoded by the built-in `json.JSONEncoder`, then the sql panel doesn't get the params and subsequently can't perform the query or `EXPLAIN` the query.

In my case, I often have params that are instance of `datetime.datetime`.

I currently use the following patch: https://github.com/zt-initech/django-debug-toolbar/commit/47e766434a81622d94432069257f7f4bc79e5dde

It is not very nice. I don't like using both JSON and pickle, and there are probably other problems.

How do you think this problem should be fixed?
For example, should params always be pickled?
Can you help me fix this patch to get it merged upstream?

jieter:
The param `strings_only=True` caused params with type `time`, `date` and
`datetime` to not be converted to string by `NormalCursorWrapper._decode()`,
and in turn causes the json dumping to fail.

This change keeps `strings_only=True` for every type but datetime, date and
time, enabling conversion to string in those cases.

Fixes #882

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, unicode_literals
 
+import datetime
 import json
 from threading import local
 from time import time
@@ -95,8 +96,10 @@ class NormalCursorWrapper(object):
         return [self._quote_expr(p) for p in params]
 
     def _decode(self, param):
+        # make sure datetime, date and time are converted to string by force_text
+        CONVERT_TYPES = (datetime.datetime, datetime.date, datetime.time)
         try:
-            return force_text(param, strings_only=True)
+            return force_text(param, strings_only=not isinstance(param, CONVERT_TYPES))
         except UnicodeDecodeError:
             return '(encoded string)'
 
@@ -114,7 +117,7 @@ class NormalCursorWrapper(object):
             _params = ''
             try:
                 _params = json.dumps([self._decode(p) for p in params])
-            except Exception:
+            except TypeError as e:
                 pass  # object not JSON serializable
 
             template_info = get_template_info()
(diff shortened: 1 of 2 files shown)


jieter:
Apparently I'm a bit too optimistic on the support of those date/time lookups in django, let me fix that.

jieter:
I changed the test so it passed on all test environments. Unfortunately, this means I cannot test `time` and `date`, as no fields on user are of that time an the `__time` and `__date` lookups are not widely supported yet.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -114,7 +117,7 @@ def _record(self, method, sql, params):
             _params = ''
             try:
                 _params = json.dumps([self._decode(p) for p in params])
-            except Exception:
+            except TypeError as e:
```
matthiask: I'm a bit surprised that this pull request passes the build, since this is an unused variable right here.

jieter: Strange indeed. Just pushed a commit removing the variable.

jieter:
@matthiask ran into this bug again today, it would be nice to have this into a released version. Is there a plan to do a new release?

matthiask:
Thanks for the reminder. It's on a long list of things I want to do. I'll hopefully get around to it in the next few days.

jieter:
@matthiask I'm a bit hesitant to bug you again, but can I do anything to help to get this released? (ran into another query with date-params I wanted to debug...

matthiask:
Because it took such a long time to do this: 1.10.1 has been released this week.

jieter:
Yes, I noticed. Thanks!

PR #1020, opened 2017-12-23
Drop support for Django < 1.11 and remove workarounds

jdufresne:
Per Django's recommendation, now that django-debug-toolbar supports 2.0,
it should drop support for Django < 1.11 to ease maintenance and
compatibility across Django versions. So long as no deprecation warnings
are produced, django-debug-toolbar should remain forward compatible with
the next Django version.

For more details, see the Django docs:

https://docs.djangoproject.com/en/2.0/releases/2.0/#third-party-library-support-for-older-version-of-django

> Third-party library support for older version of Django
>
> Following the release of Django 2.0, we suggest that third-party app authors
> drop support for all versions of Django prior to 1.11. At that time, you
> should be able to run your package’s tests using python -Wd so that
> deprecation warnings do appear. After making the deprecation warning fixes,
> your app should be compatible with Django 2.0.

Those wishing to continue using an older Django can pin the version
django-debug-toolbar.


matthiask:
Thank you! I agree with this change. It also makes #1018 obsolete (drop Python 3.3 support) while making maintenance easier.

This uglyness can also be removed now? https://github.com/jazzband/django-debug-toolbar/blob/master/debug_toolbar/__init__.py#L17

jdufresne:
Thanks for the review and identifying that cleanup. I've scanned the code or all use of `django.VERSION` and have made additional cleanups. Ready for review.

matthiask:
Thanks!

PR #1024, opened 2018-01-08
Fix Index Error: list index out of range

jackton1:
```python 
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/core/handlers/exception.py", line 41, in inner
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 217, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/core/handlers/base.py", line 215, in _get_response
    response = response.render()
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/template/response.py", line 107, in render
    self.content = self.rendered_content
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/template/response.py", line 84, in rendered_content
    content = template.render(context, self._request)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/template/backends/django.py", line 66, in render
    return self.template.render(context)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/template/base.py", line 207, in render
    return self._render(context)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/test/utils.py", line 106, in instrumented_test_render
    template_rendered.send(sender=self, template=self, context=context)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/dispatch/dispatcher.py", line 193, in send
    for receiver in self._live_receivers(sender)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/django/dispatch/dispatcher.py", line 193, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "/Users/user/.virtualenvs/venv/lib/python3.6/site-packages/debug_toolbar/panels/templates/panel.py", line 103, in _store_template_info
    pformatted = self.pformat_layers[index]
IndexError: list index out of range
```

### DIFF
--- a/debug_toolbar/panels/templates/panel.py
+++ b/debug_toolbar/panels/templates/panel.py
@@ -101,7 +101,7 @@ class TemplatesPanel(Panel):
                 if key_values in self.seen_layers:
                     index = self.seen_layers.index(key_values)
                     pformat_layers = self.pformat_layers
-                    if pformat_layers and index in zip(*enumerate(pformat_layers))[0]:
+                    if pformat_layers and index in list(zip(*enumerate(pformat_layers)))[0]:
                         pformatted = self.pformat_layers[index]
                         context_list.append(pformatted)
                 else:
--- a/debug_toolbar/panels/templates/panel.py
+++ b/debug_toolbar/panels/templates/panel.py
@@ -100,11 +100,9 @@ class TemplatesPanel(Panel):
                 key_values = sorted(context_layer.items())
                 if key_values in self.seen_layers:
                     index = self.seen_layers.index(key_values)
-                    try:
+                    pformat_layers = self.pformat_layers
+                    if pformat_layers and index in zip(*enumerate(pformat_layers))[0]:
                         pformatted = self.pformat_layers[index]
-                    except IndexError:
-                        continue
-                    else:
                         context_list.append(pformatted)
                 else:
                     temp_layer = {}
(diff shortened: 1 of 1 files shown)


matthiask:
@jackton1 When would this happen? 

@kezabelle Would be great to have your eyes on this if you find the time. Thanks in advance :)

kezabelle:
If I'm re-reading the pseudo-cache stashing correctly, I think for this to have happened, this has to fail: `force_text(pformat(temp_layer))` and the context layer has to appear twice?
In which case there's 2 options I think:
- handling it here makes sense, as it means we can skip looking at the layer again, and it'll silently skip this every time it sees it **OR**;
- change so that when `UnicodeEncodeError` happens, store the layer as `<<unicode encode error>>` ... though I think it must be a `UnicodeEncodeError` on the *key* to avoid hitting that repr-replacement?

jedie:
I see this error, too.
It's with:

Django==1.11.14
django-debug-toolbar==1.9.1

It happens only at the first request, after the developing server restarts.

Any work-a-round for this?

camilonova:
@jackton1 could you please add a test?

jedie:
This issues still exists in django-debug-toolbar==1.10.1 :(

Moulde:
I have this issue as well.

jackton1:
@camilonova been a while I added this change, I'll include the test.

tim-schilling:
Closing this PR due to inactivity and the lack of a test case to reproduce this. If anyone is still having issues with this, please let me know and we can work towards fixing the issue.

PR #1026, opened 2018-01-11
Pass python_requires argument to setuptools

jdufresne:
Helps pip decide what version of the library to install.

https://packaging.python.org/tutorials/distributing-packages/#python-requires

> If your project only runs on certain Python versions, setting the
> python_requires argument to the appropriate PEP 440 version specifier
> string will prevent pip from installing the project on other Python
> versions.

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> python_requires
>
> A string corresponding to a version specifier (as defined in PEP 440)
> for the Python version, used to specify the Requires-Python defined in
> PEP 345.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,6 +16,7 @@ setup(
     download_url='https://pypi.python.org/pypi/django-debug-toolbar',
     license='BSD',
     packages=find_packages(exclude=('tests.*', 'tests', 'example')),
+    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
     install_requires=[
         'Django>=1.11',
         'sqlparse>=0.2.0',


matthiask:
I hope we will keep this up to date :-/ 

It definitely makes sense, but still.

jdufresne:
@matthiask Do you think it would help if I added a blurb to the [making a release](https://django-debug-toolbar.readthedocs.io/en/stable/contributing.html#making-a-release) documentation about updating `python_requires` (and identified versions)? Something similar was done for django-rest-framework: https://github.com/encode/django-rest-framework/commit/b2ec681d8d4c5a4a5ff3febb43449be5eea07883

matthiask:
Oh yes, this most likely would help! Thanks in advance.

jdufresne:
I added some docs in #1037 to try to make this easier to remember.

PR #1029, opened 2018-01-20
handling the case if return_value is None in get_many

lokeshatbigbasket:
Hi,

I ran into a case where I am trying `get_many` with empty list, in that case, debug toolbar is throwing exception and the toolbar is not appearing. Here I came up with a fix. I think that if `return_value` is `None` it's neither a cache miss nor a cache hit right??? Please review and merge if the code is correct.

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -163,7 +163,7 @@ class CachePanel(Panel):
                 self.misses += 1
             else:
                 self.hits += 1
-        elif name == 'get_many':
+        elif name == 'get_many' and return_value:  # what if return_value is None?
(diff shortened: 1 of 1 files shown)


lokeshatbigbasket:
Hey @robhudson,

can you review the above PR ???

jdufresne:
Could you add a test?

camilonova:
@lokeshatbigbasket please add a test so is more clear what do you want to fix. Thank you.

matthiask:
I'm closing this pull request because of inactivity, not rejecting the feature. Please resubmit if you still want to work on this, after resolving the conflicts and adding a unit test.

PR #1034, opened 2018-01-21
Fix example.urls to work with Django 2.0

jdufresne:
Fixes unhandled exception:

```
django.core.exceptions.ImproperlyConfigured: Passing a 3-tuple to include() is not supported. Pass a 2-tuple containing the list of patterns and app_name, and provide the namespace argument to include() instead.
```


matthiask:
Thanks!

PR #1036, opened 2018-01-23
In templates, load "static" templatetag instead of "staticfiles"

jdufresne:
Since Django 1.10, it is preferable to load the static templatetag. If in `INSTALLED_APPS`, the templatetag will use `django.contrib.staticfiles` to determine static paths. For additional details, see:

https://docs.djangoproject.com/en/dev/releases/1.10/#django-contrib-staticfiles

> The static template tag now uses `django.contrib.staticfiles` if it’s in `INSTALLED_APPS`. This is especially useful for third-party apps which can now always use `{% load static %}` (instead of `{% load staticfiles %}` or `{% load static from staticfiles %}`) and not worry about whether or not the `staticfiles` app is installed.

The staticfiles templatetag will be deprecated and removed in a future Django version.


matthiask:
Nice. Now dropping compatibility with older Django versions finally pays off.

PR #1039, opened 2018-01-27
Add testing for Django master branch

jdufresne:
Will help ensure that django-debug-toolbar stays compatible as Django is
developed.


jdufresne:
Good idea. I've made the change in the latest revision.

matthiask:
Thank you!

PR #1040, opened 2018-01-30
Handle of Composed object from psycopg2

stephane:
http://initd.org/psycopg/docs/sql.html#psycopg2.sql.Composed


jdufresne:
Would it be possible to add a test?

camilonova:
@stephane can you add a test?

stephane:
Added but now running tests requires a PostgreSQL server.

camilonova:
@aaugustin any ideas how to support this?

aaugustin:
Provide a separate settings file that uses PostgreSQL and skip the test if not running with PostgreSQL. There are examples of doing this in Django itself.

Issue #986, opened 2017-08-20
SQL panel in django 1.11


ad-m:
Hello,

SQL panel doesn't show any queries on Django 1.11+.  It works after downgrade to ``django <1.11``.

Greetings,

tclancy:
I'm seeing this as well. Are you on Postgres? If so, it appears the new use of [cursors in Postgres in Django 1.11](https://docs.djangoproject.com/en/1.11/ref/databases/#transaction-pooling-server-side-cursors) is the culprit. You will need to add [`'DISABLE_SERVER_SIDE_CURSORS': True`](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-DATABASE-DISABLE_SERVER_SIDE_CURSORS) to each entry in your `DATABASES` setting to get the queries working again.

ad-m:
@tclancy:

One application:
``django==1.11.4`` + ``django-debug-toolbar==1.8`` + MySQL => 0 queries
``django==1.10.7`` + ``django-debug-toolbar==1.8`` + MySQL => 0 queries
``django==1.9.13`` + ``django-debug-toolbar==1.8`` + MySQL => 0 queries

Second application:
``Django==1.11.4`` + ``django-debug-toolbar==1.8`` + PostgreSQL (default) = 1 query
``django==1.10.7`` + ``django-debug-toolbar==1.8`` + PostgreSQL (default) => 1 query

Third application:
``django==1.11.4`` + ``django-debug-toolbar==1.8`` + MySQL => 0 queries
``django==1.10.7`` + ``django-debug-toolbar==1.8`` + MySQL => 0 queries
``django==1.9.13`` + ``django-debug-toolbar==1.8`` + MySQL => 0 queries

It's look that I have common misconfiguration in my application using MySQL or MySQL support is broken.

ad-m:
It's due duplicate ```debug_toolbar.middleware.DebugToolbarMiddleware``` middleware.

blueyed:
See https://github.com/jazzband/django-debug-toolbar/pull/1106 for handling duplicate entries (which can be a valid use case).

PR #1052, opened 2018-04-05
Validate duplicate middleware load

ad-m:
Protection for #986.


kezabelle:
Looks sensible enough to me.

matthiask:
To me as well, but the build fails with a flake8 error:
https://travis-ci.org/jazzband/django-debug-toolbar/jobs/362824956

Can you please fix this @ad-m and ping me after this?

matthiask:
Thank you!

blueyed:
I think this should be a warning only.
See https://github.com/jazzband/django-debug-toolbar/issues/1102.

PR #1056, opened 2018-04-19
Fix crash in profiling when no directory name found

yanjost:


matthiask:
Thanks! I think this looks good, apart from the `flake8` failure.

matthiask:
Thank you!

PR #1062, opened 2018-06-03
Add sorting for sql panel query time.

onemadgeek:
Sorting added for SQL query time to quickly identify which query costing more. Any suggestions?


matthiask:
Hmm, I don't like this too much. 

If you have so many queries on a page that it's not visible at a glance which queries took longest you should probably spend some time optimising the query count before spending developer time on individual queries.

Also, I fear the maintenance burden of introducing sorting here, and later everywhere.

onemadgeek:
Thanks for the response. I understand there will be maintenance and improvement burden. I'll look into some other improvement. If you have any suggestion let me know.

PR #1064, opened 2018-06-10
Fix invalid/non-representative tests

jdufresne:
Passing bytes objects to a model CharField (a string field) is not
supported and incorrect use of the API. When Python is started with -b,
it can result in the warnings:

```
.../django/db/models/fields/__init__.py:1078: BytesWarning: str() on a bytes instance
  return str(value)
.../django/db/models/fields/__init__.py:1078: BytesWarning: str() on a bytes instance
  return str(value)
```

Avoid this warning by passing str values, as expected.

For details on the Python -b flag, see:

https://docs.python.org/3/using/cmdline.html#miscellaneous-options

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -61,15 +61,11 @@ class SQLPanelTestCase(BaseTestCase):
         list(User.objects.filter(username='thé'))
         self.assertEqual(len(self.panel._queries), 2)
 
-        # non-ASCII bytes parameters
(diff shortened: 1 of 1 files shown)


matthiask:
Hmm. It *is* incorrect use of the queryset API, but doesn't it still make sense to test that django-debug-toolbar doesn't crash in this case?

jdufresne:
A use case for binary values is querying by a `BinaryField`. I'll see if there is a better way to handle this.

PR #1070, opened 2018-07-26
Declare local JS variable to avoid leaking to global scope

jdufresne:
Without the use of `var`, `handle` becomes a global variable. Avoid that by declaring it local.


matthiask:
Thanks!

PR #1071, opened 2018-07-26
Fix "make example" command to run the dev server

jdufresne:
Previously failed with:

```
  ModuleNotFoundError: No module named 'example'
```


matthiask:
Thanks!

PR #1072, opened 2018-07-26
stacktrace: include number of hidden frames

blueyed:

PR #1073, opened 2018-07-26
settings: HIDE_IN_STACKTRACES: more finegrained hiding of django

blueyed:
Hiding all of Django in stacktraces is too much and not necessary usually.

I do not remember the details, but had this in a local patch since a
while already.

This goes good together with https://github.com/jazzband/django-debug-toolbar/pull/1072, but can be seen independent from it - therefore a separate PR.


matthiask:
Makes sense, thanks!

PR #1074, opened 2018-07-31
Make JS varaible "current" local to the event handler

jdufresne:
Save a tiny amount of memory by not keeping the value around. Only used by the event handler. The value does not need to be remembered across different invocations of the event handler.


camilonova:
Thank you

Issue #1053, opened 2018-04-12
Bugfix for issue #910 raises error with pandas DataFrame objects


anotherzo:
We are using debug-toolbar in a django project where we have [DataFrame](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.html) objects of [pandas](https://pandas.pydata.org/) as items in the context_layer.

The pseudo cache code added to panel.py (e.g., around lines 97ff) breaks with DataFrame objects in context_layer.items(), because the 'if key_values in self.seen_layers' requires a comparison of key_values with all items in seen_layers, but DataFrame objects are not always so easy to compare without [throwing an error](https://stackoverflow.com/questions/18548370/pandas-can-only-compare-identically-labeled-dataframe-objects-error).

Right now, we simply put a try-except statement in our local panel.py file(s), but it would be better to have a real patch in the django-debug-toolbar code for rollouts. I realize you could argue this is a pandas problem and not a debug-toolbar problem, but it still would be great for us to have some exception handling so Pandas does not simply kill the toolbar code.

Besides, thanks for this tool. We love it :) .

anotherzo:
To clarify:
In file debug_toolbar/panels/templates/panel.py, around line 97, we use now
<pre><code>key_values = sorted(context_layer.items())
try:
    is_seen = key_values in self.seen_layers
except:
  is_seen = False
if is_seen:
   ...
</code></pre>
Not pretty, but works for us.

flinz:
Just ran into the same issue, and I am not using a local version of toolbar which makes patching this a little harder.. (didn't look into monkeypatching it though)

👍 for me on this issue.

iamjonmiller:
Running into the same issue on a pandas heavy project. 

`Value Error: Can only compare identically-labeled Series objects`

iamjonmiller:
While I no longer have to edit the source with the latest update, I still periodically encounter this error.

tim-schilling:
Closing as stale. If someone runs into this again, please provide steps to reproduce.

Issue #989, opened 2017-09-04
1.8 seems to block form field rendering with Django 1.11. 1.7 works fine (though slow)


eskhool:
Hi, 
On rendering a form field using a simple {{ my_form_name.my_field_name }} which works fine without debug toolbar makes the field disappear in 1.8. It works fine with 1.7 (both tested using Django 1.11) although 1.7 is very slow (2-3 seconds for page load)

eskhool:
Have narrowed the issue down to the Templates Panel. Now investigating why and how can the panel influence the rendering itself (which shouldn't ever happen)

eskhool:
@matthiask, since you were pretty active with release 1.8 and this issue seems to have started with 1.8,  do you have any quick ideas what might be causing this?

matthiask:
We had massive slowdowns with Django 1.11 that were reported in #910. I thought those were fixed but your report suggests that this isn't the case. Do you have a chance of profiling the panel to find out where the problem might be? No quick ideas unfortunately :-(

eskhool:
@matthiask, I think I didn't make myself clear. The slowdowns are indeed fixed with DjDT 1.8 however the template rendering seems to be broken in places as the rendered page is missing content specifically with the Templates panel enabled. That shouldn't happen at all! 

I am using a simple template variable from a form field which works fine without djdt1.8 but now with and it has to do with the template panel

{{ my_form_name.my_field_name }}

sspross:
we experienced the same problem and the problem was a field queryset, which is somehow evaluated by `debug_toolbar.panels.templates.TemplatesPanel`.

e.g.

```
class MyForm(forms.Form):
    user = forms.ModelChoiceField(queryset=User.objects.all())
```

kezabelle:
I'm fairly certain the problem, at least as @sspross has shown it, is my doing, as part of #933, because `pformat` was so slow, we (I) moved to using `force_text` to evaluate the variable (which is intrinsically known to perhaps cause an SQL query) - forgetting that the `__str__` for anything `Form` related is ... the HTML for the form. Probably using `_saferepr` is more correct, whilst still avoiding some of the inherent slowdown using `pformat` multiple times causes.

timthelion:
I'm suffering from a problem with django-debug-toolbar version: 1.9.1 in which the `required` tag is being removed from form inputs, iff django-debug toolbar is enabled. Is my problem a duplicate of this bug, or should I be working on a minimum test case myself?

[...]

Issue #725, opened 2015-06-24
Issues with pformat for contexts


bradshjg:
When contexts are extremely large, the pformat calls in panels/templates/panel.py can take >15 seconds.

Would you be amenable to a pull request that allows for a setting that would replace pformat with a call to repr (or something similar)?

tim-schilling:
Yes.

jbbqqf:
Hi! Triaging older issues — I think this one has shipped (in 2017) and can be closed.

The `TemplatesPanel` now memoizes `pformat` output across context layers, which is the workaround the issue asked about:

- Commit [`e0b6f888`](https://github.com/django-commons/django-debug-toolbar/commit/e0b6f888) ("Refs #910 - After Django introduced template based widget rendering, the number of context values which need prettifying has increased drastically...") introduced the `pformat_layers` cache on 2017-04-24.
- The current implementation lives at [`debug_toolbar/panels/templates/panel.py:79`](https://github.com/django-commons/django-debug-toolbar/blob/main/debug_toolbar/panels/templates/panel.py#L79) (`self.pformat_layers = []`) and the cache lookup at [`debug_toolbar/panels/templates/panel.py:136-185`](https://github.com/django-commons/django-debug-toolbar/blob/main/debug_toolbar/panels/templates/panel.py#L136-L185) — identical context layers reuse a previously-formatted `pformat` result rather than re-running it.
- Additionally, [`docs/tips.rst`](https://github.com/django-commons/django-debug-toolbar/blob/main/docs/tips.rst) ("Performance considerations" section) now documents `SHOW_TEMPLATE_CONTEXT` and `SKIP_TEMPLATE_PREFIXES` as escape hatches when the panel is still too slow on a project.

If you still hit pathological pformat slowness on a current version, a fresh issue with a reproducer would be easier to act on than this one. Otherwise, would you mind closing this out?

---

*Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the references above were verified manually against the current `main` branch.*

PR #1076, opened 2018-08-06
WIP - template panel context issues

kezabelle:
As I've now encountered an issue myself where the #910 changes caused problems, possibly related to some of the others, probably I should try and fix them all up.

As a short reminder, after template based widgets came along, and before #942 landed, I tried to ensure the number of calls to `pprint.pformat` was kept to a minimum, as it is an expensive calculation (#933).

In doing so, the call changed from: `try: ... pformat(value); except: ...` to use `force_text` instead of `pformat`. This was, in hindsight, a mistake, mostly around usages of `Form` instances.

Because `__str__` on a Form will by default **render** the form, it has a number of side-effects:
- The repr won't get used if the rendering causes an SQL query to be triggered (ie: most forms!) (one I've now encountered)
- The form instance can have `full_clean` called on it because the `__str__` includes error printing (one I've now encountered)
- Possibly subsequently prevents forms from rendering correctly at all? (#989)

------------

In an ideal world, we could just change to `repr` (#725) instead of `force_text`, as this is similarly lightning fast. Instead, to err on the side of caution, I've opted for calling `pprint.saferepr` instead, which is *protected against recursive data structures.* 
This has the benefit of still being faster than pformat (as far as I can tell) and avoids a certain class of problems the other options bring. But it's much slower than calling `repr` or `force_text` 
directly.

------------
Marking as `WIP` because there's a number of other issues (#989, #1024, #1053, #725) are related, and one option is just to **undo** the changes I made to avoid too many `pformat` calls. Also there's no tests ;)

### DIFF
--- a/debug_toolbar/panels/templates/panel.py
+++ b/debug_toolbar/panels/templates/panel.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, unicode_literals
 from collections import OrderedDict
 from contextlib import contextmanager
 from os.path import normpath
-from pprint import pformat
+from pprint import pformat, saferepr
 
 from django import http
 from django.conf.urls import url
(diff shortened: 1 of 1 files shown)


matthiask:
Makes sense to me. Thanks!

PR #1077, opened 2018-08-07
Refactor js

xi:
I would like to help refactoring the javascript code and potentially removing the jquery dependency. I know this has been discussed in #562 already. But since then, 4 years have passed and some new DOM APIs have become available. 

As a first step I did some small changes to bring the code up to modern jquery standards.

Please let me know if you are interested in this work or if there is anything else frontend-related I could help with.

PR #1083, opened 2018-08-31
Reactor js part 2

xi:
Continues my work from #1077.

I think I found some code that was not really used. I am not completely sure though.

Note that the `<details>` element I used to replace some functionality is currently not supported by edge.

I plan to continue working on this code with the aim to remove jquery. This is a lot of work though and I am not sure how long it will take. So I prefer to do small steps and get your feedback often. If you would like me to prepare a single large pull request I could probably also do that, but it would take some time.


xi:
Replaced by #1085

PR #1084, opened 2018-09-03
duplicate_key: Handle raw_params being None sometimes

pjdelport:
This is a small bug-fix update to #995 that I missed in my local testing. (Whoops.)

### DIFF
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -149,7 +149,8 @@ class SQLPanel(Panel):
             return query['raw_sql']
 
         def duplicate_key(query):
-            return (query['raw_sql'], tuple(query['raw_params']))
+            raw_params = () if query['raw_params'] is None else tuple(query['raw_params'])
+            return (query['raw_sql'], raw_params)
 
         if self._queries:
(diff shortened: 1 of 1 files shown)


pjdelport:
@camilonova: I hope you don't mind a quick review. :)

debug_toolbar/panels/sql/panel.py
```diff
@@ -149,7 +149,8 @@ def similar_key(query):
             return query['raw_sql']
 
         def duplicate_key(query):
-            return (query['raw_sql'], tuple(query['raw_params']))
+            raw_params = () if query['raw_params'] is None else tuple(query['raw_params'])
```
jieter: Just ran into this as well using master, nice to get this fixed! 

I'd use though `tuple(query['raw_params'] or ())`

pjdelport: @jieter: I prefer the more explicit `None` check, since `or` can hide unexpected type errors that should be raised rather than silenced.

jieter: fine of course! currently working on this branch, works like charm!

matthiask:
Thank you!

Issue #605, opened 2014-06-04
Toolbar introduces race condition on requireJS sites


spectras:
On version 1.2.1, do not know for `master` nor older releases. Opening an issue so other people with the same problem can find this. It turns out google was not my friend on that one.

**Root cause:**
In file `debug_toolbar/templates/debug_toolbar/base.html` (the one that gets injected into site's pages), lines 8 to 10, the code hijacks `window.define` and restores it later.

However, as it is a global and requireJS does aynchronous loading, a whole bunch of funny stuff can happen randomly.

**Some effects:**
Here is what I experienced until I eventually found out debug_toolbar was the culprit:
- some other module gets ready and tries to call `define` at this very time, breaking with a `define is not defined`, aborting module load.
- more fun: the main jQuery completing its loading asynchronously, registering itself as window.jQuery only to have debug toolbar call noConflict on it, before requireJS has time to grab the reference. Therefore, everything appears to load fine, except module that use jQuery will get `undefined` instead of jQuery.

**Workaround:**
As a workaround, it is possible to set `JQUERY_URL` to None in the settings. This forces debug_toolbar to use the main jquery and not try to load its own. However, doing so requires that the main jQuery be loaded before the toolbar kicks in so either:
- load jQuery synchronously
- or wrap the whole scripting part of debug_toolbar in a `require(['jquery'], function(jQuery){   })` block, then wrap toolbar's script in `define` blocks. That's messy though, I would not recommend it.

matthiask:
Probably fixed by #1085.

PR #1085, opened 2018-09-05
Remove jquery

### LINKED ISSUE #562: Should jQuery be a dependency ?
Related a few issues, but mainly #535 

Here is [the list of `$` occurences in the source code](http://sprunge.us/fQdh) [1].
- The selectors can be replaced by `.querySelectorAll` and `.querySelector`
- `on()` can be replaced by `addEventListener`
- `cookie()` can be somewhat easily reimplemented
- `removeClass()` can be replaced by `.style.cssText = .style.cssText.replace(/\bClass\b/)`
- `hide()` can be replaced by `.style.display = 'none'`
- `show()` can be replaced by `.style.display = 'initial'`
- `parent()` can be replaced by `.parentNode`
- `.html()` can be replaced by `.innerHTML`
- `.find()` can be replaced by `.querySelectorAll`
- `.text()` can be replaced by `.innerText`
- `.addClass()` can be replaced by `+=  ' Class'`
- `.trigger()` can be replaced by `dispatchEvent()`
- `toggle()` can be replaced by a `.style.display = .style.display === 'none' ? 'initial' : 'none'`
- `is()` is a bit harder to replace, but the selector is always `:visible`, so `.style.display !== 'none'`

etc, etc

Removing jQuery from the dependencies would have multiple benefits :
- No conflict with users environments (which would be great, since it would make the Debug Toolbar a lot less obtrusive)
- A lot less code, and better performances overall (since there is relatively few jQuery use cases in the current codebase)
- No vendors to ship

And the following disadvantages :
- We would have to maintain the very small library which would reimplement jQuery functionalities
- It could break on some browser versions

Overall, I think it would be for the best : the Debug Toolbar is a debug component, and the developers can use a modern browser to use it. Since no client will ever see it, the compatibility issue isn't very relevant.

<hr />

[1]

```
grep -R '\$' /usr/local/lib/python2.7/dist-packages/debug_toolbar/static/debug_toolbar/js/ | grep -Po '\.[a-zA-Z]+\(' | sort | uniq -c | sort -gr
```

xi:
This is a full port to vanilla javascript, ditching jquery. Fixes #562.

It uses many modern DOM features such as [fetch](https://caniuse.com/#feat=fetch), [classList](https://caniuse.com/#search=classList), or [NodeList.forEach](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach). All of them are supported in all major browsers though.

This includes some refactoring which might also fix some bugs (e.g. #605).

I removed the open/close animation. It is probably possible to bring it back, but I did not focus on that yet.


matthiask:
Awesome!

xi:
That was quick. Thanks!

WhyNotHugo:
The removal of the `JQUERY_URL` setting doesn't seem to be on the CHANGELOG.

Had a bit of a hard time tracking down what had happened here. Maybe it's still worth updating?

matthiask:
@WhyNotHugo Sure! Thanks. I just proposed a documentation change here: https://github.com/jazzband/django-debug-toolbar/pull/1266

PR #1089, opened 2018-09-06
Remove unnecessary CSS in jQuery example

jdufresne:
Simply use a `<strong>` tag instead. One fewer CSS rule.


matthiask:
Thanks!

PR #1096, opened 2018-09-08
adding support to view file from stacktraces

miki725:
Added a way to view files from stacktrace. Useful when looking at SQL panel which allows to view code directly in debugging without having a need to jump to editor.

image

image

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -645,6 +645,15 @@
 #djDebug .djdt-file span.djdt-highlight-line {
 	font-weight: bold;
 }
+#djDebug .djdt-file .djdt-line-no {
+	-webkit-touch-callout: none; /* iOS Safari */
+	-webkit-user-select: none; /* Safari */
+	-khtml-user-select: none; /* Konqueror HTML */
+	-moz-user-select: none; /* Firefox */
+	-ms-user-select: none; /* Internet Explorer/Edge */
+	user-select: none; /* Non-prefixed version, currently supported by Chrome and Opera */
+	margin-right: 10px;
+}
 
 #djDebug .djdt-width-20 {
     width: 20%;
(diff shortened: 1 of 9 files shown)


debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -615,16 +615,35 @@
     margin-bottom: 3px;
     font-family:Consolas, Monaco, "Bitstream Vera Sans Mono", "Lucida Console", monospace;
 }
+#djDebug .djdt-stack {
+	white-space: normal;
+}
 #djDebug .djdt-stack span {
-    color: #000;
```
miki725: this file is inconsistent with indents. some are tabs and some are spaces. let me know if I should make it consistent in whole file.

debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -111,20 +111,16 @@
                 var name = this.tagName.toLowerCase();
                 var ajax_data = {};
 
-                if (name == 'button') {
+                if (name === 'button' || name === 'a') {
                     var form = this.closest('form');
-                    ajax_data.url = this.getAttribute('formaction');
+                    ajax_data.url = name === 'button' ? this.getAttribute('formaction') : this.getAttribute('href');
```
miki725: not sure if this has side-effects anywhere else. I could switch to using a button to view file however that did not looked well so switched back to using a link

matthiask:
The templates panel already does a similar thing, without a form though but with django.core.signing:
https://github.com/jazzband/django-debug-toolbar/blob/master/debug_toolbar/panels/templates/views.py#L14

It would be great to reuse the implementation we already have if possible.

miki725:
sure. I can take a look. didnt realize template already did something similar. I see it also uses pygments so Ill try to include that as well

matthiask:
I'm closing this pull request because of inactivity, not rejecting the feature. Please resubmit if you still want to work on this, after resolving conflicts and adapting the code for current versions of Django etc.

Issue #816, opened 2016-04-14
MultiValueDictKeyError: "u'store_id'"


vkaracic:
Hello! I'm trying to enable the DDT but it doesn't get displayed. If I go to ``__debug__/render_panel/`` I get this exception (so I'm thinking this might be the key to my problem):

```
Traceback (most recent call last):
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/debug_toolbar/views.py", line 12, in render_panel
    toolbar = DebugToolbar.fetch(request.GET['store_id'])
  File "/edx/app/ecommerce/venvs/ecommerce/local/lib/python2.7/site-packages/django/utils/datastructures.py", line 322, in __getitem__
    raise MultiValueDictKeyError(repr(key))
MultiValueDictKeyError: "u'store_id'"
```

Does anyone know how to fix this?

sasha0:
Did you use explicit setup to enable DDT on your project?

aaugustin:
`/ __debug__/render_panel/` by itself doesn't render anything, it needs some query string parameters, so this is unrelated to the issue you're seeing.

vkaracic:
Sorry guys, we went on from this and I can't remember the details. I'm closing this issues and if it comes up again I'll reopen with some more details.

scuml:
Got this error all of a sudden.  Not sure what triggered it, but noticed the javascript was now requesting: ` "GET /__debug__/render_panel/?%5Bobject+Object%5D= HTTP/1.1" 500`

Replicated by opening a console in Safari and Chrome and typing:
`new URLSearchParams({"a":1}).toString();`
which returns: `"%5Bobject+Object%5D="`

Was able to patch in `toolbar.js` with: 

```
var url = djDebug.getAttribute('data-render-panel-url');
var url_params = new URLSearchParams();
url_params.append('store_id', store_id);
url_params.append('panel_id', this.className);
url += '?' + url_params.toString();
```

matthiask:
[...]

PR #1098, opened 2018-09-10
Patch for store_id: MultiValueDictKeyError

scuml:
Fixes issue found in https://github.com/jazzband/django-debug-toolbar/issues/816

Mozilla's documentation shows that strings are meant to be parsed in the constructor and the support of parsing a dictionary is not guaranteed to work across browsers.  
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

This updates toolbar.js to be compliant with the current spec.


scuml:
Thanks for the quick turnaround @xi & @matthiask

Issue #1102, opened 2018-09-13
Provide ids for checks, turn duplicate-middleware into warning?


blueyed:
1. there should be ids for the checks, so that they can be ignored easily

2. I have not checked if I am affected by #986, but I have a valid use case for having the middleware in there twice: I want it as early as possible, but then also after AuthenticationMiddleware (for custom `request.user` checks).  I would argue that this should be a warning instead, so that it does not fail hard to start runserver for example.

Code: https://github.com/jazzband/django-debug-toolbar/blob/27f8a70953aebbd4949dc34ef552ce441d463a40/debug_toolbar/apps.py#L49

matthiask:
Fair enough, but does it even work if it's included twice?

And sure: Adding IDs to the errors sounds good!

blueyed:
It used to work at least.. ;)

I am using this:
```python
def show_debug_toolbar(request):
    from debug_toolbar.middleware import DebugToolbarMiddleware
    import threading

    if (DebugToolbarMiddleware.debug_toolbars.get(
            threading.current_thread().ident)):
        return False
    if not int(request.GET.get('my_enable_ddt', not request.path.startswith(
            reverse('some-path')))):
        return False
    if hasattr(request, 'user'):
        return (request.user and
                request.user.groups.filter(name='Debug').exists())
    from debug_toolbar.middleware import show_toolbar as ddt_show_toolbar
    return ddt_show_toolbar(request)
```

So it checks if it has been enabled already:
```python
    if (DebugToolbarMiddleware.debug_toolbars.get(
            threading.current_thread().ident)):
        return False
```

PR #1103, opened 2018-09-13
checks: use warnings instead of errors, add ids

blueyed:
Fixes https://github.com/jazzband/django-debug-toolbar/issues/1102.


matthiask:
The tests now of course fail because they are very specific :)

Just a small input: How about making the ID start with `debug_toolbar` instead of [ddt](https://en.wikipedia.org/wiki/DDT)?

blueyed:
Cool, I agree with your changes!
Thanks for picking it up.

PR #1107, opened 2018-10-18
Fix unicode error when trying to escape binary data

dbowd:


matthiask:
`isinstance(element, six.binary_type)` and handling binary data explicitly would be nice. Also, can you please add a test for this so that it doesn't break in the future?

matthiask:
Hmm, reading this again it seems to have some relation to https://github.com/jazzband/django-debug-toolbar/pull/1064

@jdufresne Care to chime in? Maybe you some advice to offer when it comes to handling binary data in django-debug-toolbar?

dbowd:
@matthiask I encountered this bug when creating an object with a `models.BinaryField()` field.

The test is using the `username` field because there is no `BinaryField` in the django User model. I am making a raw query using `cursor.execute()` because django does complain about binary data as mentioned by @jdufresne.

The exact same exception is encountered in both cases.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -84,7 +84,12 @@ def __init__(self, cursor, db, logger):
 
     def _quote_expr(self, element):
         if isinstance(element, six.string_types):
-            return "'%s'" % force_text(element).replace("'", "''")
+            try:
+                return "'%s'" % force_text(element).replace("'", "''")
+            except DjangoUnicodeDecodeError:
+                return repr(element)
+        elif isinstance(element, six.binary_type):
```
jdufresne: If you put this check first, do you still need the `except DjangoUnicodeDecodeError`? I guess it is required for Python 2.7?

tests/panels/test_sql.py
```diff
@@ -122,6 +122,18 @@ def test_param_conversion(self):
...
         ))
 
+    @unittest.skipUnless(connection.vendor not in ('sqlite', 'postgresql'), '')
+    def test_binary_param_force_text(self):
+        self.assertEqual(len(self.panel._queries), 0)
+
+        with connection.cursor() as cursor:
+            cursor.execute("SELECT * FROM auth_user WHERE username = %s", [b'\xff'])
+
+        self.assertEqual(len(self.panel._queries), 1)
```
jdufresne: Is it possible to assert the contents of the query? I think that would make the test more interesting.

tests/panels/test_sql.py
```diff
@@ -122,6 +122,18 @@ def test_param_conversion(self):
             '["2017-12-22 16:07:01"]'
         ))
 
+    @unittest.skipUnless(connection.vendor not in ('sqlite', 'postgresql'), '')
```
jdufresne: I think avoiding the double negative would be more readble.

I'm a little confused as to why these DBs are excluded. How do they handle `BinaryField`?

dbowd: ### SQLite:

```
  File "tests/panels/test_sql.py", line 130, in test_binary_param_force_text
    cursor.execute("SELECT * FROM auth_user WHERE username = %s", [b'\xff'])

ProgrammingError: You must not use 8-bit bytestrings unless you use a text_factory that can interpret 8-bit bytestrings (like text_factory = str). It is highly recommended that you instead just switch your application to Unicode strings.
```
### Postgresql:

```
  File "/home/travis/build/jazzband/django-debug-toolbar/.tox/postgresql/lib/python3.6/site-packages/django/db/backends/utils.py", line 85, in _execute
    return self.cursor.execute(sql, params)
psycopg2.ProgrammingError: operator does not exist: character varying = bytea
LINE 1: SELECT * FROM auth_user WHERE username = '\xff'::bytea
                                               ^
HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
```

jdufresne: I'm suggesting that you write this as:

```py
@unittest.skipIf(connection.vendor in ('sqlite', 'postgresql'), '')
```

You should also fill in the reason string instead of using the empty string.

dbowd: Yes, I understand your suggestion, I will make the change when I get a chance.
I was just answering the question you had asked about why they were being excluded.

dbowd: Can you suggest a reason I could place in the `skipIf`?

jdufresne: Maybe something like:

```
'Mixing bytestrings and text is not allowed on PostgreSQL and SQLite'
```

WDYT? Feel free to tweak it.

dbowd:
Hi @jdufresne, can you review this PR.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -88,7 +88,12 @@ def __init__(self, cursor, db, logger):
...
+            return "(binary data)"
```
jdufresne: Should this use `repr(element)` as well? I guess I'm confused why sometimes binary data is represented but other times it is not. Or alternatively, should above be:

```py
            except DjangoUnicodeDecodeError:
                return "(binary data)"
```

Is this branch tested? I don't see `(binary data)` added to the tests.

tests/panels/test_sql.py
```diff
@@ -113,6 +113,25 @@ def test_param_conversion(self):
...
+            cursor.execute("SELECT * FROM auth_user WHERE username = %s", [b"\xff"])
+
+        self.assertEqual(len(self.panel._queries), 1)
+        self.assertEqual(
+            self.panel._queries[0][1]["sql"],
+            "SELECT * FROM auth_user WHERE username = '\ufffd'",
+        )
+
+        self.panel.process_response(self.request, self.response)
+        self.panel.generate_stats(self.request, self.response)
```
jdufresne: What is the purpose of calling these _after_ the assert? Did you intend to assert something after this line?

debug_toolbar/panels/sql/tracking.py
dbowd: I agree, it should be `return repr(element)` but in that case there is no point specifically testing for `elif isinstance(element, six.binary_type):` because the `else:` clause also returns `return repr(element)`.

jdufresne:
Should we add MySQL/MariaDB to Travis so this gets covered? We could use the PostgreSQL target as a template.

matthiask:
@dbowd Thanks!

@jdufresne That's a good idea. I added a basic PR here: https://github.com/jazzband/django-debug-toolbar/pull/1120

PR #1108, opened 2018-10-25
Use defer attribute on all script tags, remove inline JS

adamchainz:
I'm deploying the experimental `Feature-Policy` on my site. I tried disabling the `sync-script` policy, which blocks HTML-parse-blocking JavaScript, and django-debug-toolbar broke because its script tags aren't async or deferred.

There's a bit more information about this feature, and a demo at https://feature-policy-demos.appspot.com/sync-script.html?on

I've added `defer` to all the `<script>` tags, which makes them execute after HTML parsing has finished, in order. There was also one inline script for redirects which I've extracted - this would break for `sync-script` but also for use of `Content-Security-Policy` blocking inline JavaScript, which has been supported since #704 but it seems this was missed.

Tested by using the `example` app, redirect interception worked and every panel loaded and looked right.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/redirect.js
+++ b/debug_toolbar/static/debug_toolbar/js/redirect.js
@@ -1,3 +1,1 @@
-document.addEventListener('DOMContentLoaded', function () {
-  document.getElementById('redirect_to').focus();
-});
+document.getElementById('redirect_to').focus();
--- a/debug_toolbar/static/debug_toolbar/js/redirect.js
+++ b/debug_toolbar/static/debug_toolbar/js/redirect.js
@@ -1,0 +1,3 @@
+document.addEventListener('DOMContentLoaded', function () {
+  document.getElementById('redirect_to').focus();
+});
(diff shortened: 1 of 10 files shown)


adamchainz:
@graingert you added CSP and like security features, perhaps you'd like to check this?

debug_toolbar/static/debug_toolbar/js/redirect.js
```diff
@@ -0,0 +1,3 @@
+document.addEventListener('DOMContentLoaded', function () {
```
graingert: This should be redundant with defer

debug_toolbar/static/debug_toolbar/js/redirect.js
```diff
@@ -0,0 +1,3 @@
...
+  document.getElementById('redirect_to').focus();
```
graingert: https://caniuse.com/#search=autofocus

adamchainz: This is focussing a link not a form element :(

debug_toolbar/static/debug_toolbar/js/redirect.js
adamchainz: Idk how far back django-debug-toolbar tries to support, but ancient browsers don't support `defer`, so best to leave it here?

matthiask:
Thanks!

adamchainz:
No worries, thanks for active maintenance!

PR #1109, opened 2018-10-25
Remove py35-djmaster test environment

adamchainz:
Failing on Travis ([example](https://travis-ci.org/jazzband/django-debug-toolbar/jobs/446114199)) with:

> Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.5.tar.bz2
> $ curl -sSf -o python-3.5.tar.bz2 ${archive_url}
> curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
> Unable to download 3.5 archive. The archive may not exist. Please consider a different version.

It looks like they removed the package so it can't be used any more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,6 @@ matrix:
       env: TOXENV=py37-dj21
       dist: xenial
       sudo: true
-    - python: 3.5
-      env: TOXENV=py35-djmaster
     - python: 3.6
       env: TOXENV=py36-djmaster
     - python: 3.7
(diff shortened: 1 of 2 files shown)


jdufresne:
Are you sure this wasn't due to a transient issue? The py35-dj21 envrionment seems to be building fine. Unfortunately, I've found Travis to have the intermittent failure you posted. Can you try rebuilding it?

adamchainz:
You're right, it was transient, the rebuild on #1108 passed for the environment.

PR #1110, opened 2018-10-25
Ignore flake8 warning W503

adamchainz:
Currently breaking tests with the latest flake8, which added a pair of competing rules, W503 and W504, for whether line breaks should be before or after a binary operator. `ignore` needs to ignore at least one of them, otherwise `flake8` can never be pleased.

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -79,6 +79,6 @@ def is_middleware_class(middleware_class, middleware_path):
     except ImportError:
         return
     return (
-        inspect.isclass(middleware_cls) and
-        issubclass(middleware_cls, middleware_class)
+        inspect.isclass(middleware_cls)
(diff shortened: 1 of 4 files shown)


matthiask:
I don't care very much which warning is ignored but black recommends ignoring W503 https://github.com/ambv/black#line-breaks--binary-operators -- do you have a preference?

adamchainz:
I did this to avoid having to change any code, will switch to W503 to copy black/pep8.

matthiask:
Don't worry, it's fine for me as it is -- I was just asking :)

(Maybe we should just switch to black completely instead?)

jdufresne:
> Maybe we should just switch to black completely instead?

I'm in favor of this. :+1: I've found it very helpful to just let black handle formatting.

adamchainz:
Already changed

matthiask:
Thanks!

PR #1112, opened 2018-10-26
Always deterministically close files in setup.py

jdufresne:
When running Python with warnings enabled, fixes warning:

```
setup.py:12: ResourceWarning: unclosed file <_io.TextIOWrapper name='README.rst' mode='r' encoding='utf-8'>
  long_description=open("README.rst", encoding="utf-8").read(),
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,12 +4,18 @@ from io import open
 
 from setuptools import find_packages, setup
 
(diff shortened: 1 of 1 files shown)


matthiask:
Nice, thanks!

PR #1113, opened 2018-11-07
Add tip about browser caching.

FlipperPA:
More than once, I've spent time tweaking settings, when all I really needed to do was a hard refresh after an upgrade to force the JavaScript to reload. Mentioning this in the documentation.


matthiask:
Thank you! That is definitely a good idea.

PR #1115, opened 2018-11-14
Stop inlining images in CSS

adamchainz:
These are done on top of including the raw assets, doubles the work. There's no need for this optimization since DDT is for debug mode only, and with a restrictive Content Security Policy, the `data:` URL's trigger an unnecessary error before the real asset gets loaded.

These were last touched in #1097 / b3561007aa2f75f7b373681fdaaacd2d428f569d.


matthiask:
Thanks!

adamchainz:
Can we get a release? I'd like to use this and #1108.

PR #1116, opened 2018-11-23
Display indented SQL queries in expanded views.

thomas-riccardi:
This reuses the query details toggle to display the indented version: the default view does not change.

This uses `&nbsp;` and `<br/>` to convert `sqlparse.format` from text to html visual indentation, it has a small visual glitch with the `+` button to toggle details on the `SELECT` first line, but hopefully it's OK for everyone.

Also, use expanded form in SQL select/explain/profile views to profit from the indented formatting.

Result:
screenshot_20181123_194110
screenshot_20181123_194130

### DIFF
--- a/debug_toolbar/panels/sql/forms.py
+++ b/debug_toolbar/panels/sql/forms.py
@@ -77,7 +77,7 @@ class SQLSelectForm(forms.Form):
         return hash
 
     def reformat_sql(self):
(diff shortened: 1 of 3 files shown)


thomas-riccardi:
I have *not* executed the selenium tests yet (simply following the contributing doc resulted in an error (`WebDriverException: Message: 'geckodriver' executable needs to be in PATH.` )); they may thus be broken...

PR #1117, opened 2018-11-25
Add missing import in installation doc URLconf section.

hramezani:


matthiask:
Thanks. The import seems to be redundant though since `debug_toolbar` is also imported at the first line in the `if DEBUG` block:
image

hramezani:
Oh, my bad. I didn't see it. so sorry.

matthiask:
Don't worry!

PR #1120, opened 2018-11-28
Run tests with MariaDB too on Travis CI

matthiask:


matthiask:
Reverted, I pushed the wrong remote :-(

PR #1121, opened 2018-11-28
Run tests with MariaDB

matthiask:


matthiask:
@dbowd I added a "true" `BinaryField` in this pull request because I couldn't get the test to work otherwise (I always got collation-related errors). While doing this I noticed that it was relatively straightforward to make the test pass in PostgreSQL and sqlite3 too. The formatting of the parameter is a bit different so I left it out though. 

I'm a bit surprised though that the test also seems to pass *without* the `DjangoUnicodeDecodeError` thingy you added, so maybe the test tests something which it shouldn't now.

PR #1125, opened 2019-01-06
Update Travis CI matrix

kevin-brown:
- [x] Test Python 3.7 and Django 1.11
- [x] Switch to using tox-travis


kevin-brown:
Tests appear to be failing on Python 3.7 + MariaDB, and that appears to be unrelated to this pull request (which does not affect that configuration).

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist =
-    py{27,34,35,36}-dj111
+    py{27,34,35,36,37}-dj111
```
matthiask: Yes, please.

.travis.yml
```diff
@@ -46,13 +23,9 @@ matrix:
...
     - python: 3.7
       env: TOXENV=readme
-  allow_failures:
-    - env: TOXENV=py35-djmaster
-    - env: TOXENV=py36-djmaster
-    - env: TOXENV=py37-djmaster
 
 install:
-  - pip install tox codecov
+  - pip install tox tox-travis codecov
```
matthiask: Please don't. tox-travis hides jobs for individual Django versions in the Travis CI build overview.

kevin-brown:
> tox-travis bundles all Django versions into one build making it a bit more cumbersome to understand what's happening.

We both have differing ideas of how useful knowing the Django version that broke a build is on first glance.

----

The alternative that I generally recommend for larger projects where combining parts of the matrix would cause absurdly long individual build times (think 20 mins vs 5 mins) is to set an environment variable such as `DJANGO` with the Django version number that is being tested. That way Travis CI clearly shows the versions of Django + Python that are broken and tox-travis will still be able to detect it within the matrix.

You can see an example of a Travis matrix using that here: https://travis-ci.org/encode/django-rest-framework

matthiask:
OK, I didn't know that it is possible to keep separate builds for Django versions even when using tox-travis.

However, if you're doing this you're back to defining version combinations twice again, once in [tox.ini](https://github.com/encode/django-rest-framework/blob/master/tox.ini) and once in [.travis.yml](https://github.com/encode/django-rest-framework/blob/master/.travis.yml) so I'm unclear how this is an improvement and/or a simplification when compared to what we're currently doing?

matthiask:
Also regarding

> We both have differing ideas of how useful knowing the Django version that broke a build is on first glance.

In my* experience stuff breaks more often because of differences between Django versions than between Python versions (excluding 2.x vs. 3.x maybe, but not that much either anymore). I'm definitely not of the opinion that the Django version is less important.

* And also with my experience.

jdufresne:
FWIW, I too prefer avoiding tox-travis. But that is just my opinion and I understand this may come down to one's personal preference.

matthiask:
I misread what @jdufresne wrote (I didn't see "avoiding"). 

#1126 adds py37-dj111 to the build without adding tox-travis. Let's move forward with this.

Thanks @kevin-brown for the idea!

PR #1128, opened 2019-01-18
Handle removed staticfiles.templatetags in Django 3.0

jdufresne:
django.contrib.staticfiles.templatetags.staticfiles was deprecated in Django 2.1:

https://docs.djangoproject.com/en/dev/releases/2.1/#id2

> `{% load staticfiles %}` and `{% load admin_static %}` are deprecated in favor of `{% load static %}`, which works the same.

And will be removed in Django 3.0:

https://docs.djangoproject.com/en/dev/releases/3.0/#features-removed-in-3-0

> The `staticfiles` and `admin_static` template tag libraries are removed.

Catch and ignore an `ImportError` as the module may not exist. Fixes tests against the Django master branch.

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -5,7 +5,6 @@ from os.path import join, normpath
 
 from django.conf import settings
 from django.contrib.staticfiles import finders, storage
-from django.contrib.staticfiles.templatetags import staticfiles
 from django.core.files.storage import get_storage_class
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import LazyObject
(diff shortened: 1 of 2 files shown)


matthiask:
Thank you! 

I suspect that changing `d.c.s.templatetags.staticfiles` isn't even necessary. The `staticfiles_storage` import in this module is gone since cf546e11ac76c8dec527e39ff8ce8249a195ab42 in 2015, and Django 1.11 was released in 2017.

jdufresne:
Thanks, excellent point! That makes the cleanup even easier. In the latest revision I have completely removed `django.contrib.staticfiles.templatetags.staticfiles.staticfiles_storage`.

matthiask:
Thanks!

PR #1132, opened 2019-01-23
Make ProfilingPanal enabled but inactive by default

jdufresne:
I believe the ProfilingPanal is disabled by default to avoid interfering
with other middleware and panels during profiling. The same can be
achieved by making the panel inactive. This makes it easier for users to
access the panel without additional configuration but continues to avoid
profiling interference.


matthiask:
Thanks! 

`DISABLE_PANELS` is a bad name if it does not disable panels but just deactivate them, but that's a different issue.

PR #1137, opened 2019-02-15
Replace MIDDLEWARE_CLASSES references to MIDDLEWARE

jdufresne:
`MIDDLEWARE_CLASSES` is deprecated and removed in Django 2.0, so refer to
`MIDDLEWARE` by default.


matthiask:
Good find, thanks!

PR #1139, opened 2019-02-16
Move common, repetitive test setup to BaseTestCase

jdufresne:
The subclass now only needs to define the class attribute "panel_id" and
BaseTestCase handles the finding it and enabling it.


matthiask:
Thank you! Great cleanup. I'm impressed by the count and quality of your pull requests :)

PR #1140, opened 2019-02-24
Restructure the Panel class to execute more like the MIDDLEWARE

### LINKED ISSUE #1135: profiling panel causes custom middleware to skip "process_exception" call
I detailed my experiments with this here: https://stackoverflow.com/questions/54346767/django-2-0-custom-middleware-process-exception-hook-not-firing?noredirect=1#comment95511527_54346767

Basically I have a custom middleware class that has a `process_exception` method, but if I also have debug toolbar enabled with the profiling panel added in `DEBUG_TOOLBAR_PANELS` than the `process_exception` method in my middleware class will never fire, I can see the traceback for the error occuring but the returned response is a standard django 500 debug page.

Middleware:

    class MyMiddleware(MiddlewareMixin):
        def process_exception(self, request, exception):
            print("Exception!")

Activated middleware

    DJANGO_MIDDLEWARE = [
        'django.middleware.security.SecurityMiddleware',
        'django.contrib.sessions.middleware.SessionMiddleware',
        'django.middleware.common.CommonMiddleware',
        'django.middleware.csrf.CsrfViewMiddleware',
        'django.contrib.auth.middleware.AuthenticationMiddleware',
        'django.contrib.messages.middleware.MessageMiddleware',
        'django.middleware.clickjacking.XFrameOptionsMiddleware',
        'debug_toolbar.middleware.DebugToolbarMiddleware',
        'path.to.MyMiddleware',
    ]

Panels:

    DEBUG_TOOLBAR_PANELS = (
        'debug_toolbar.panels.versions.VersionsPanel',
        'debug_toolbar.panels.timer.TimerPanel',
        'debug_toolbar.panels.settings.SettingsPanel',
        'debug_toolbar.panels.headers.HeadersPanel',
        'debug_toolbar.panels.request.RequestPanel',
        'debug_toolbar.panels.sql.SQLPanel',
        'debug_toolbar.panels.templates.TemplatesPanel',
        'debug_toolbar.panels.cache.CachePanel',
        'debug_toolbar.panels.signals.SignalsPanel',
        'debug_toolbar.panels.logging.LoggingPanel',
        'debug_toolbar.panels.redirects.RedirectsPanel',
        'debug_toolbar.panels.profiling.ProfilingPanel',  # commenting this out works
    )

jdufresne:
This allows the `ProfilingPanel` to no longer skip remaining panels or middlewares.

This change is backwards incompatible. Third party panels will need to adjust to use this new architecture. It removes support for Django 1.11's deprecated `MIDDLEWARE_CLASSES` setting.

The `Panel.__init__()` method is now passed a get_response argument. The `Panel.process_request()` method must now always return a response. Usually this is the response returned by `get_response()` but may also return a different response as is the case in the `RedirectsPanel`.

`Panel.process_response()` and `Panel.process_view()` have been removed.

Fixes #1135

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -23,14 +23,8 @@ def check_middleware(app_configs, **kwargs):
     gzip_index = None
     debug_toolbar_indexes = []
 
-    setting = getattr(settings, "MIDDLEWARE", None)
-    setting_name = "MIDDLEWARE"
-    if setting is None:
-        setting = settings.MIDDLEWARE_CLASSES
-        setting_name = "MIDDLEWARE_CLASSES"
-
     # Determine the indexes which gzip and/or the toolbar are installed at
-    for i, middleware in enumerate(setting):
+    for i, middleware in enumerate(settings.MIDDLEWARE):
         if is_middleware_class(GZipMiddleware, middleware):
             gzip_index = i
         elif is_middleware_class(DebugToolbarMiddleware, middleware):
@@ -41,9 +35,9 @@ def check_middleware(app_configs, **kwargs):
         errors.append(
             Warning(
                 "debug_toolbar.middleware.DebugToolbarMiddleware is missing "
-                "from %s." % setting_name,
+                "from MIDDLEWARE.",
                 hint="Add debug_toolbar.middleware.DebugToolbarMiddleware to "
-                "%s." % setting_name,
+                "MIDDLEWARE.",
                 id="debug_toolbar.W001",
             )
         )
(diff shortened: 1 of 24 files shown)


matthiask:
This is an excellent idea. I should probably polish the state of current things and publish a release first.

auvipy:
@matthiask that would be a better approach.

jdufresne:
> This is an excellent idea. I should probably polish the state of current things and publish a release first.

That sounds good to me. Ping me if you need me to rebase or modify anything after that goes through.

jdufresne:
@matthiask @auvipy If we're going to introduce backwards incompatibility, maybe we get some other cases out of the way at the same time. WDYT? Some that come to mind:

- Drop Python 2 support
- All the deprecation warnings in `debug_toolbar/settings.py` `get_config()`
- `debug_toolbar.panels.DebugPanel`

If this is agreeable, I can submit separate PRs for those points.

auvipy:
IMHO, it's time to drop python 2 from next major release which breaks BC.

PR #1141, opened 2019-02-24
Fix toggling of table rows in the profiling panel UI

jdufresne:
Fixes JS error. The error in Firefox was:

    TypeError: container is null

        init http://localhost:8000/static/debug_toolbar/js/toolbar.js:147
        on http://localhost:8000/static/debug_toolbar/js/toolbar.js:7


matthiask:
Thanks!

PR #1142, opened 2019-02-24
Remove py35-djmaster from the test matrix

jdufresne:
Django removed support for Python 3.5 in commit:
https://github.com/django/django/commit/7e6b214ed34f5562dbd83cf54924a5b589a29715


matthiask:
👍

PR #1145, opened 2019-03-02
Remove support for Python 2

jdufresne:
- Removed use of `six`
- Removed `__future__` imports
- Removed coding cookie comment (Python 3 defaults to utf-8)
- Removed some `force_text()` calls where the value is known to be str
- Replaced deprecated `ugettext` with `gettext`
- Removed object as parent class
- Simplified use of `super()`
- Updated `setup.cfg` file to reflect that without Python 2 support, the wheel is no longer "universal"
- Updated trove classifiers


auvipy:
please rebase!

jdufresne:
Rebased :slightly_smiling_face:

PR #1147, opened 2019-03-26
WIP: Add proper failing testcase for binary data

Naddiseo:
I noticed there was a hole in the tests for actual binary data. I came across this because I was trying to save a pickled pandas model in a `BinaryField`, but everytime I did, djdt would die. I've tracked it down to the binary containing the proposed change in this patch. 
It's currently a failing test case in mariadb/mysql and I haven't looked into how to fix it, any guidance on that would help and I'll also submit a fix

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -108,7 +108,7 @@ class SQLPanelTestCase(BaseTestCase):
         with connection.cursor() as cursor:
             cursor.execute(
                 "SELECT * FROM tests_binary WHERE field = %s",
-                [connection.Database.Binary(b"\xff")],
+                [connection.Database.Binary(b"\udc80\x04\xff")],
             )
 
         response = self.panel.process_request(self.request)


Naddiseo:
Hmm, It didn't fail.. strange, on my machine I get:

```
  File "/tmp/django-debug-toolbar/tests/panels/test_sql.py", line 119, in test_binary_param_force_text
  File "/tmp/django-debug-toolbar/debug_toolbar/panels/sql/panel.py", line 211, in generate_stats
    query["form"] = SQLSelectForm(auto_id=None, initial=copy(query))
  File "/tmp/django-debug-toolbar/debug_toolbar/panels/sql/forms.py", line 38, in __init__
    initial["hash"] = self.make_hash(initial)
  File "/tmp/django-debug-toolbar/debug_toolbar/panels/sql/forms.py", line 83, in make_hash
    m.update(force_bytes(item))
  File "/tmp/django-debug-toolbar/env/lib/python3.7/site-packages/django/utils/encoding.py", line 105, in force_bytes
    return s.encode(encoding, errors)
UnicodeEncodeError: 'utf-8' codec can't encode character '\udcff' in position 50: surrogates not allowed
```

But, the only other change I made was to use pymysqldb instead of mysqlclient. so I'm not sure where the error is, any help tracking this down would be helpful!

tim-schilling:
I think this can be closed as per #1148.

PR #1148, opened 2019-03-26
added pymysql as a testing environment

Naddiseo:
As mentioned in #1147, I think I found an issue with the way djdt interacts with binary data with pymysql is used, so this PR adds testing with pymysql.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,10 @@ matrix:
       env: TOXENV=mariadb
       addons:
         mariadb: "10.3"
+    - python: 3.7
+      env: TOXENV=pymysql
+      addons:
+        mariadb: "10.3"
     - env: TOXENV=flake8
     - python: 3.7
       env: TOXENV=style
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -88,6 +88,13 @@ elif os.environ.get("DJANGO_DATABASE_ENGINE") == "mysql":
     DATABASES = {
         "default": {"ENGINE": "django.db.backends.mysql", "NAME": "debug_toolbar"}
     }
+elif os.environ.get("DJANGO_DATABASE_ENGINE") == "pymysql":
+    import pymysql
+
+    pymysql.install_as_MySQLdb()
+    DATABASES = {
+        "default": {"ENGINE": "django.db.backends.mysql", "NAME": "debug_toolbar"}
+    }
 else:
     DATABASES = {"default": {"ENGINE": "django.db.backends.sqlite3"}}
 
--- a/tox.ini
+++ b/tox.ini
@@ -7,6 +7,7 @@ envlist =
     py{36,37}-djmaster
     postgresql,
     mariadb,
+    pymysql,
     flake8,
     style,
     readme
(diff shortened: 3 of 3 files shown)


Naddiseo:
The error caused by this is the one I saw in #1147; I'm not sure why it's being caused. Anyone at @jazzband care to help me through this?

matthiask:
I'm not sure I follow. What should have happened in your opinion?

You wanted the PR #1147 to fail with the `surrogates not allowed` message but it didn't. This one does (with a different position but the same byte) so that should be fine?

Naddiseo:
@matthiask, I might be misunderstanding how unicode works here, but looking at the error message for this PR it says it can't encode `'\udcff'`, however, I don't see `\udc` inside of `test_binary_param_force_text` anywhere, which is what is confusing me. 

TBH, I wasn't expecting this PR to fail at all, but to use it as the basis for reproducing the error I observed in #1147, so I guess I'm looking for direction.

I think I'm XY-problem-ing myself, layout how I got here:
* I have a django model with a Binary field that I'm saving a pickled object into.
* I wanted to use DJDT to profile a page that updates the pickled object
* The request completes, but takes 20seconds, and there's a very large traceback in the console about unicode encoding errors
* I see that the traceback originates inside of DJDT
* Looking at the traceback, it appears that it's caused in the sql form where it tries to deduplicate the queries by hashing them, and consequently forcing the sql params into bytes
* I tried to create a failing test case in DJDT to show the error (#1147), however it didn't fail
* I see the only different is the database driver (pymysql vs mysql-python)
* I come to the conclusion that there's a bug in how DJDT and pymysql interact with binary data, but I don't know where, or in which package.

Do you think this is a bug in DJDT, or in pymysql? Are my two PRs worth pursuing?

tim-schilling:
This looks to be in pymysql. Running the tests in #1147 with postgres run fine locally, but running them with pymysql break.

GytisZ:
Happened upon this issue as well. We're using `pymysql` and `django-debug-toolbar` and djdt fails when the code triggers SQL that touches binary fields.

```
[django] Traceback (most recent call last):
[django]   File "/home/vagrant/.virtualenvs/audienceapi/lib/python3.6/site-packages/django/core/handlers/exception.py", line 47, in inner
[django]     response = get_response(request)
[django]   File "/home/vagrant/.virtualenvs/audienceapi/lib/python3.6/site-packages/debug_toolbar/middleware.py", line 67, in __call__
[django]     panel.generate_stats(request, response)
[django]   File "/home/vagrant/.virtualenvs/audienceapi/lib/python3.6/site-packages/debug_toolbar/panels/sql/panel.py", line 214, in generate_stats
[django]     query["form"] = SQLSelectForm(auto_id=None, initial=copy(query))
[django]   File "/home/vagrant/.virtualenvs/audienceapi/lib/python3.6/site-packages/debug_toolbar/panels/sql/forms.py", line 37, in __init__
[django]     initial["hash"] = self.make_hash(initial)
[django]   File "/home/vagrant/.virtualenvs/audienceapi/lib/python3.6/site-packages/debug_toolbar/panels/sql/forms.py", line 82, in make_hash
[django]     m.update(force_bytes(item))
[django]   File "/home/vagrant/.virtualenvs/audienceapi/lib/python3.6/site-packages/django/utils/encoding.py", line 99, in force_bytes
[django]     return str(s).encode(encoding, errors)
[django] UnicodeEncodeError: 'utf-8' codec can't encode character '\udc80' in position 96: surrogates not allowed
```

Would be very interested in a fix or a guidance towards some workaround. As it stands - it means that we need to disable djdt.

PR #1160, opened 2019-05-05
Django 2.2's admin app checks for SessionMiddleware's presence

matthiask:


matthiask:
One unrelated MariaDB build still there... :-/

PR #1165, opened 2019-05-08
Updated configuration documentation

huberu:
Debug toolbar settings look for DEBUG_TOOLBAR_CONFIG in django settings.
Documentation implies this dict is called CONFIG_DEFAULTS.


matthiask:
Thanks!

PR #1167, opened 2019-05-11
also instrument `chunked_cursor`

### LINKED ISSUE #1041: Missing SQL queries
While recently trying to diagnose a slow admin page, I noticed that certain queries were missing from the SQL panel. By looking at the `django.db.backends` log I can see that the missing queries are exactly those that are using named cursors (`DECLARE ... NO SCROLL CURSOR WITH HOLD FOR SELECT `) instead of regular `SELECT` statements. I believe that these are server-side cursors used with PostgreSQL for code that uses `queryset.iterator()` (such as [`ModelChoiceField`](https://github.com/django/django/blob/a38ae914d89809aed6d79337b74a8b31b6d3849a/django/forms/models.py#L1136)).

So my guess is that DJDT is not recording queries that use these cursors, but I'm not sure.

I'm using DJDT version 1.9.1 and Django 2.0.1.

davidszotten:
In addition to wrapping `cursor`, also wrap `chunked_cursor` so that
queries using server side cursors get recorded.

Fixes #1041

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -36,6 +36,7 @@ recording = state.recording  # export function
 def wrap_cursor(connection, panel):
     if not hasattr(connection, "_djdt_cursor"):
         connection._djdt_cursor = connection.cursor
+        connection._djdt_chunked_cursor = connection.chunked_cursor
 
         def cursor(*args, **kwargs):
             # Per the DB API cursor() does not accept any arguments. There's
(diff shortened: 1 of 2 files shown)


matthiask:
Thanks!

davidszotten:
Thanks for the quick response!

What's the process for releasing new versions of the library?

matthiask:
I just released an alpha version of django-debug-toolbar 2.0; testing is very welcome!

davidszotten:
amazing, thanks! alpha seems to work for me, though i pretty much only use the sql panel

PR #1169, opened 2019-05-25
Remove unsupported Python 3.4 references from additional places

jdufresne:
Support for Python 3.4 was dropped in 733986f541de72e8718ce7f502fafc95e6139465 but a few references remained.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 style:
 	isort -rc debug_toolbar example tests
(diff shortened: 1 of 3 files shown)


jdufresne:
Looks like the trailing comma from Black is a known upstream bug: https://github.com/python/black/issues/759

matthiask:
Good find!

PR #1170, opened 2019-06-20
Simplify TemplatesPanel with Django's Model._meta.label property

jdufresne:
Update test to assert the full panel context content to demonstrate that
there is no change in the rendered content.


matthiask:
Excellent, thanks!

PR #1175, opened 2019-06-20
Simplify TemplatesPanel to use a single associated list

jdufresne:
Rather than have two lists -- and risk their indexes go out of sync -- use a single list of tuples. That way, the key/values can never go out of sync.


jdufresne:
I have rebased to resolve merge conflicts.

matthiask:
Makes sense. I'm a bit hesitant here since performance and code stability has been problematic in this area, but let's merge this and see. It might even improve performance since it only seems to scan the list of seen layers once.

PR #1176, opened 2019-06-20
Run the fast style check targets first in the test matrix

jdufresne:
Relative to the tests, the style checks run very first. When running the
full test suite, it is useful to get fast early feedback rather than
wait for multiple tests to run only to realize double quotes should have
been used or there is an unused import.


matthiask:
Makes sense! Thanks.

PR #1181, opened 2019-07-04
Update default panels in docs

francoisfreitag:
Profiling panel is enabled by default since c74b6319116e39b9923ae1798aadbf2af90b786c.


matthiask:
Thank you!

PR #1183, opened 2019-07-07
Simplifying versions of Django in Tox

ad-m:


pauloxnet:
Can be simpler using tilde notation ?
e.g `Django~=2.2.0` instead of `Django==2.2.*`

See the `Compatible release` version clause in PEP440
https://www.python.org/dev/peps/pep-0440/#compatible-release

PR #1190, opened 2019-08-05
Possible fix for #902.

allan-simon:
The SQL Panel does not show the correct template source context for {% include %}'d template files.

I've tested it with latest version of django debug tool bar and latest stable version of django 1.11 and I confirms it works


allan-simon:
closed, black was failing, and i haven't write right on this repo, so I've recreated it

PR #1191, opened 2019-08-06
fix #902, correctly display the stacktrace for SQL requests in template

### LINKED ISSUE #902: SQL panel points a query in an included template to the same line of its parent template

allan-simon:


matthiask:
Thank you!

A test would be great for this.

allan-simon:
sorry as I've been mostly recycling someone's else MR I don't think I will be able to provide a test :(

gilmrjc:
I'm working on this on #1436. It's safe to close this PR.

Thank you @allan-simon for the initial work 😄 .

PR #1194, opened 2019-08-10
Update URL: python/black → psf/black

jdufresne:


matthiask:
Thanks! Good to know the repo moved.

PR #1198, opened 2019-09-04
Correct ProfilingPanel documentation

tim-schilling:
The ProfilingPanel is not included by default, nor is it disabled by default.


matthiask:
It should be included but disabled by default though? See https://github.com/jazzband/django-debug-toolbar/commit/61dcdfd03ffe597d9e7727e9675c0bbcf75df2ad

tim-schilling:
You're correct, sorry for the confusion.

PR #1210, opened 2019-11-24
Update the HIDE_IN_STACKTRACES documentation to reflect settings.

### LINKED ISSUE #1192: Docs have incorrect default for HIDE_IN_STACKTRACES
At https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html , the `HIDE_IN_STACKTRACES` default is stated to be `('socketserver', 'threading', 'wsgiref', 'debug_toolbar', 'django')`. The actual default only hides a few Django modules, not all of it.

In 2.0, the actual default is:
```
    "HIDE_IN_STACKTRACES": (
        "socketserver",
        "threading",
        "wsgiref",
        "debug_toolbar",
        "django.db",
        "django.core.handlers",
        "django.core.servers",
        "django.utils.decorators",
        "django.utils.deprecation",
        "django.utils.functional",
    ),
```

Edit: I just realized that these are prefixes rather than exact matches, so submodules are also omitted, which is great. Would this be worth documenting?

Related question: Is anyone else having their browser lag and max out a CPU core when looking at the SQL panel with stacktraces enabled? The reason I noticed this doc issue was I wanted to add django.test.utils, django.template.base, and django.template.loader_tags to the hidden modules to see if that fixed my issue. (I get a lot of those in certain SQL stacktraces.)

tim-schilling:
Closes #1192

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -141,8 +141,21 @@ Panel options
 
 * ``HIDE_IN_STACKTRACES``
 
-  Default: ``('socketserver', 'threading', 'wsgiref', 'debug_toolbar',
-  'django')``.
+  Default::
+
+    (
+        "socketserver",
+        "threading",
+        "wsgiref",
+        "debug_toolbar",
+        "django.db",
+        "django.core.handlers",
+        "django.core.servers",
+        "django.utils.decorators",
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks!

PR #1211, opened 2019-11-24
Reformat Server-Timing headers to updated W3C spec.

### LINKED ISSUE #1209: Server-Timing header formatted incorrectly?
The `Server-Timing` response header which was added in #928 exposes a lot of valuable information, but it looks like it's not formatted according to [the latest version of the W3C spec](https://www.w3.org/TR/server-timing/).

A demonstration of the problem can be seen in Chrome's dev tools. `Server-Timings` should be parsed and displayed here, but it's not quite working. Here's the header which is being sent (with added whitespace for readability):

```
Server-Timing: TimerPanel_utime=64.00000000000006; "User CPU time",
    TimerPanel_stime=16.000000000000014; "System CPU time",
    TimerPanel_total=80.00000000000007; "Total CPU time",
    TimerPanel_total_time=80.66487312316895; "Elapsed time",
    SQLPanel_sql_time=1.6126632690429688; "SQL 1 queries",
    CachePanel_total_time=0; "Cache 0 Calls"
```

…and here's how Chrome shows this:

Screenshot 2019-11-15 at 11 29 13

I think the header should be something like this:

```
Server-Timing: TimerPanel_utime;dur=64.00000000000006;desc="User CPU time",
    TimerPanel_stime;dur=16.000000000000014;desc="System CPU time",
    TimerPanel_total;dur=80.00000000000007;desc="Total CPU time",
    TimerPanel_total_time;dur=80.66487312316895;desc="Elapsed time",
    SQLPanel_sql_time;dur=1.6126632690429688;desc="SQL 1 queries",
    CachePanel_total_time;dur=0;desc="Cache 0 Calls"
```

tim-schilling:
Closes #1209

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1,4 +1,5 @@
 import os
+import re
 import unittest
 
 import html5lib
@@ -92,6 +93,7 @@ class DebugToolbarIntegrationTestCase(TestCase):
     def test_middleware(self):
         response = self.client.get("/execute_sql/")
         self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "djDebug")
 
     @override_settings(DEFAULT_CHARSET="iso-8859-1")
     def test_non_utf8_charset(self):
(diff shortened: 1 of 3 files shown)


jdufresne:
Thanks for the fix. Are you up for adding a test?

tim-schilling:
Ah, I didn't realize that entire function has no test. I'll add one for that.

tests/test_integration.py
```diff
@@ -92,6 +93,7 @@ class DebugToolbarIntegrationTestCase(TestCase):
     def test_middleware(self):
         response = self.client.get("/execute_sql/")
         self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "djDebug")
```
tim-schilling: This caused the test to fail until the views returned a response containing `</body>`.

tests/test_integration.py
```diff
@@ -245,6 +247,20 @@ def test_incercept_redirects(self):
...
+        expected_partials = [
+            r'TimerPanel_utime;dur=(\d)*(\.(\d)*)?;desc="User CPU time", ',
+            r'TimerPanel_stime;dur=(\d)*(\.(\d)*)?;desc="System CPU time", ',
+            r'TimerPanel_total;dur=(\d)*(\.(\d)*)?;desc="Total CPU time", ',
+            r'TimerPanel_total_time;dur=(\d)*(\.(\d)*)?;desc="Elapsed time", ',
+            r'SQLPanel_sql_time;dur=(\d)*(\.(\d)*)?;desc="SQL 1 queries", ',
+            r'CachePanel_total_time;dur=0;desc="Cache 0 Calls"',
+        ]
+        for expected in expected_partials:
+            re.compile(expected).search(server_timing)
```
jdufresne: This is missing the assert.

tim-schilling: Cripes. Thanks for catching that.

tim-schilling:
Rebased on master.

Issue #1213, opened 2019-12-03
Current 2.1 version doesn't support Django 1.11 


mohammadrabetian:
In the docs it's been said "The current stable version of the Debug Toolbar is 2.1. It works on Django ≥ 1.11." which means it should support 1.11 but it doesn't. simply because the MIDDLEWARE in settings file is written like MIDDLEWARE_CLASSES by the old style. in the debug_toolbar/apps.py line 25 the codes it's using MIDDLEWARE 

` for i, middleware in enumerate(settings.MIDDLEWARE):`

mohammadrabetian:
I looked at the 1.11 version which works. 

`   setting = getattr(settings, "MIDDLEWARE", None)
    setting_name = "MIDDLEWARE"
    if setting is None:
        setting = settings.MIDDLEWARE_CLASSES
        setting_name = "MIDDLEWARE_CLASSES"`

mohammadrabetian:
I would like to help and fix this issue if you allow me.

tim-schilling:
@mohammadrabetian This was disabled as a part of #1140. I don't think we will want to add support for that deprecated setting, especially when Django's end of support for 1.11 is in April. I think having a warning about the incompatible setting would be a nice addition. A PR would be appreciated.

PR #1217, opened 2019-12-07
added warning for incompatibility with middleware_classes setting.

mohammadrabetian:
This pr is to resolve issue #1213

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -21,6 +21,17 @@ def check_middleware(app_configs, **kwargs):
     gzip_index = None
     debug_toolbar_indexes = []
 
+    # If old style MIDDLEWARE_CLASSES is being used, report an error.
+    if not settings.MIDDLEWARE:
+        errors.append(
+            Warning(
+                "debug_toolbar is incompatible with MIDDLEWARE_CLASSES setting.",
+                hint="Use MIDDLEWARE instead of MIDDLEWARE_CLASSES",
+                id="debug_toolbar.W000",
+            )
+        )
+        return errors
+
     # Determine the indexes which gzip and/or the toolbar are installed at
     for i, middleware in enumerate(settings.MIDDLEWARE):
         if is_middleware_class(GZipMiddleware, middleware):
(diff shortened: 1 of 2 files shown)


debug_toolbar/apps.py
```diff
@@ -21,6 +21,17 @@ def check_middleware(app_configs, **kwargs):
     gzip_index = None
     debug_toolbar_indexes = []
 
+    # If old style MIDDLEWARE_CLASSES is being used, report an error.
+    if not settings.MIDDLEWARE:
+        errors.append(
+            Warning(
+                "debug_toolbar is incompatible with MIDDLEWARE_CLASSES setting.",
+                hint="Use MIDDLEWARE instead of MIDDLEWARE_CLASSES",
+                id="debug_toolbar.W000",
```
matthiask: We count warnings from 1; 1-3 are already taken. This should be "W004".

debug_toolbar/apps.py
```diff
@@ -21,6 +21,17 @@ def check_middleware(app_configs, **kwargs):
     gzip_index = None
     debug_toolbar_indexes = []
 
+    # If old style MIDDLEWARE_CLASSES is being used, report an error.
+    if not settings.MIDDLEWARE:
```
matthiask: `django.conf.settings.is_overridden("MIDDLEWARE_CLASSES")` should work better for this.

mohammadrabetian:
> Please add a test in `tests/test_integration.py` for this. Thanks!

it seems there's an integration test I've missed. I'll change my test.

matthiask:
Nice! Now you're getting a test failure because Django 1.11 also raises warnings because of `MIDDLEWARE` / `MIDDLEWARE_CLASSES`

`[<Warning: level=30, msg="The MIDDLEWARE_CLASSES setting is deprecated in Django 1.10 and the MIDDLEWARE setting takes precedence. Since you've set MIDDLEWARE, the value of MIDDLEWARE_CLASSES is ignored.", hint=None, obj=None, id='1_10.W001'>,`

Maybe you want to unconditionally suppress this warning using `SILENCED_SYSTEM_CHECKS = ["1_10.W001"]  # Suppress the MIDDLEWARE_CLASSES warning of Django 1.11`

PR #1220, opened 2019-12-07
Remove unnecessary JavaScript function applyStyle()

jdufresne:
This function iterated over elements and converted `data-*` attributes to
style attributes of the same name. Instead of using JavaScript, simply
assign these values as style attributes to begin with. This removes the
JavaScript execution time and reduces the number of JavaScript files
included with the project as well as downloaded at runtime.


matthiask:
Good find!

PR #1221, opened 2019-12-08
Use JavaScript dataset API to access data-* attributes

jdufresne:
Modern browser offer the dataset API on elements to access `data-*` attributes as a property.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset


matthiask:
Right, even IE11 supports this. https://caniuse.com/#search=dataset

PR #1232, opened 2020-01-23
Add LICENSE metadata to Python package distributions

johnthagen:
Currently [pip-licenses](https://github.com/raimon49/pip-licenses) can't include the LICENSE file for this project because the correct metadata is not setup.

This PR sets up the metadata correctly so the LICENSE file is properly included/linked to in source and wheel distributions.


matthiask:
Thanks!

PR #1241, opened 2020-03-07
Remove long commented out CSS code

jdufresne:
The code was commented in commit
ecd7f2abe62996b19720e9ade61b45eafac590e8 (Aug 24, 2009). The old CSS
will likely never be restored and provides no value being left in.


matthiask:
Thanks!

PR #1246, opened 2020-03-14
Remove vendor prefix and outdated CSS for obsolete browser versions

jdufresne:
The vendor prefixes are no longer required or recommended as the CSS is
supported across modern browsers.

Removing these prefixes will help simplify CSS and reduce its overall
asset size.

In addition, a syntax error was discovered in the use of the
linear-gradient() CSS function. This has been fixed.


matthiask:
Very nice!

PR #1250, opened 2020-03-20
Support HistoryPanel to capture ajax requests.

tim-schilling:
This creates a new panel, HistoryPanel which makes use of the
Toolbar.store to support a history of the toolbar as requests are
made.

The interface changes as follows:
Panel.is_historical - Indicates that the panel's button and content
should be updated when switching between snapshots of the history.

Toolbar.store - Will no longer generate a new store_id when the
instance already has a value.

DEBUG_TOOLBAR_CONFIG.HISTORY_POST_TRUNC_LENGTH - Allows the request's
POST content to be truncated in the history panel's content.

LoggingPanel and StaticFilesPanel now utilize the ``get_stats`` method
to fetch panel data for nav_subtitle.

Credit to @djsutho for creating the original third party panel:
https://github.com/djsutho/django-debug-toolbar-request-history
The core concepts were derived from that package.


debug_toolbar/panels/history/panel.py
```diff
@@ -0,0 +1,75 @@
...
+        ]
+
+    @property
+    def nav_subtitle(self):
+        return self.get_stats().get("request_url", "")
+
+    def generate_stats(self, request, response):
+        cleansed = request.POST.copy()
+        for k in cleansed:
+            cleansed[k] = CLEANSED_SUBSTITUTE
```
matthiask, line 45: Django contains a cleaning machinery somewhere which only substitutes secrets and keys. Might it be worth it reusing that here?

debug_toolbar/settings.py
```diff
@@ -37,6 +37,7 @@
         "django.utils.deprecation",
         "django.utils.functional",
     ),
+    "HISTORY_POST_TRUNC_LENGTH": 0,
```
matthiask, line 40: Nitpick: I think `TRUNCATE` should be spelled out and not abbreviated. Except for `MAX` no other key contains an abbreviated word currently.

tim-schilling: Makes sense.

debug_toolbar/panels/history/panel.py
tim-schilling: There is. Think we should use the [private method](https://github.com/django/django/blob/09341856ed9008875c1cc883dc0c287670131458/django/contrib/auth/__init__.py#L40) or copy it?

tim-schilling:
If we think this is going to be merged, we'll need to do a major version bump. But before we do that, I'll want to go through the third party panels identified in the docs to at least identify if any will break.

matthiask:
@tim-schilling Compatibility problems only really surface after a release, unfortunately. I think it would be alright to merge this and go ahead with it, maybe with an alpha/beta release (similar to 2.0a1) so that it is installable using `pip --pre`. 

But I certainly wouldn't want to push this without your approval or if there is anything you'd still like to check before.

tim-schilling:
I have not forgotten about this, I've been busy with paid work and haven't had time for FOSS development. I will get back around to this though.

debug_toolbar/settings.py
tim-schilling: I removed this setting entirely. Instead we're going to show a table. There's a risk in the results cache growing too large, but I'd rather not include settings if possible.

debug_toolbar/panels/history/panel.py
tim-schilling: I copied the method out of Django. You were absolutely right, sanitizing all values was unbearable to look at.

tim-schilling:
@matthiask This requires another review due to me making a few changes:

- Rebasing on the new main branch.
- Displaying the request variables as a table similar to the cache panel rather than a JSON strong.
- Supporting a refresh button within the history panel. This allows the user to load up the debug toolbar info for requests that have been performed since the original request
- Increased RESULTS_CACHE_SIZE to 25 (from 10). The reason for this is because the original request's data was being dropped before I opened the toolbar during my testing. With ajax-y applications, I feel like this new setting better reflects the number of ajax requests between full requests.

docs/panels.rst
```diff
@@ -9,6 +9,14 @@ Default built-in panels
 
 The following panels are enabled by default.
 
+History
+~~~~~~~
+
+Path: ``debug_toolbar.panels.history.VersionsPanel``
```
matthiask, line 15: Typo? (`VersionsPanel`)

debug_toolbar/panels/history/panel.py
```diff
@@ -0,0 +1,106 @@
...
+CLEANSED_SUBSTITUTE = "********************"
+
+
+def _clean_data(data):
+    """
+    Clean a dictionary of potentially sensitive info before
+    sending to less secure functions.
+    Not comprehensive
+
+    Taken from django.contrib.auth.__init__.credentials
```
matthiask, line 29: I was hoping there would be a way to use https://docs.djangoproject.com/en/3.1/ref/settings/#default-exception-reporter-filter for this. I didn't even know there was a setting for this, I assumed it was internal-only. I'd have referenced the setting earlier if I was aware that it existed at all.

(Also, maybe Django should use the `SafeExceptionReporterFilter` instead of reimplementing parts of it in django.contrib.auth.)

docs/panels.rst
tim-schilling: Yeah, copy paste error. Thanks for catching it.

debug_toolbar/panels/history/panel.py
tim-schilling: Ah, good call. I'll rework this to make use of that.

tim-schilling: I'm thinking I need to remove this. The django implementation only uses it for debug cases which defeats the purpose of this.

tim-schilling:
@matthiask I'm curious about your v3.0 release plan. Would this be included or would it be smarter to do a 3.0 and then a `3.1.0a` immediately after? I can make a post on forum.djangoproject.com asking for alpha testers.

matthiask:
@tim-schilling I think it might be worth it to do another alpha. Generally I don't worry too much about stability since mostly people aren't supposed to use django-debug-toolbar in production :-)

It's enough work to create a release already, and I'd rather speed the release process up a bit.

tim-schilling:
Sounds good. Let me know where I can help out.

matthiask:
@tim-schilling I have just pushed 3.0a3 now, if you want to add a post in the forum that would be awesome. Thanks!

tim-schilling:
@matthiask Thanks for the update! I'll do that this morning.

matthiask:
@tim-schilling It may be worth it uploading another release; I fixed a crash which happened when using `request` as a file-like object (see https://github.com/jazzband/django-debug-toolbar/compare/3.0a3...master). What do you think?

tim-schilling:
No, I doubt it's an edge-case if it was found that quickly. I agree it's worth to bump the version since it was caught and fixed.

matthiask:
Alright, https://pypi.org/project/django-debug-toolbar/3.0a4/ is available on PyPI.

tim-schilling:
Thanks, here's the forum post https://forum.djangoproject.com/t/django-debug-toolbar-alpha-release-v3/4250

Issue #1032, opened 2018-01-21
Allow AJAX support via configuration option


jgbishop:
The `process_request` routine within the `DebugToolbarMiddleware` class (in **middleware.py**) disables rendering the toolbar for AJAX requests:

    if request.is_ajax():
        return

This logic prevents third party panels from accessing data on AJAX requests and responses (e.g. the Django Debug Panel doesn't work properly with this logic in place). I propose a configuration option to allow users to turn this behavior back on. It would be fine for the option to be disabled by default, but I'm willing to put up with any performance / memory hits for the perk of being able to view AJAX data.

In investigating further, I note that this change was implemented as a part of #926 (which fixed #834). As a developer, I'm willing to risk the potential security weakness in my setup, as I only run DDT in local setups or on dedicated sandbox systems.

jdufresne:
Can you provide more details about your use case? Why do you want to render the panels in a AJAX request? Is your AJAX request returning full HTML documents such that injecting the HTML makes sense? Do you have details about what this AJAX request is doing and what you want Django debug toolbar to do with it? Do you have a minimal example project to demonstrate this use case?

Instead of adding a new option, I'd like to better understand the use case to determine if we can make the middleware work securely for you.

jgbishop:
I'm actually not interested in rendering the panels at all in an AJAX use case; I'm more interested in accessing the data the toolbar would normally gather. Primarily, I'm interested in what SQL queries are being generated by my AJAX routines (e.g. am I missing the necessary `select_related` or `prefetch_related` calls?).

As far as I can tell, no data is ever gathered for AJAX requests because the data gathering step seems to be tightly coupled to actually rendering the toolbar. The [django-debug-toolbar-request-history](https://github.com/djsutho/django-debug-toolbar-request-history) panel (a third-party panel for DDT) works around this limitation by monkey-patching your monkey-patches. In my own attempts at creating a custom panel for DDT, I don't seem to get any data for AJAX calls due to the code I mentioned in my original report above. As soon I as I disable those lines, I get data.

Perhaps the data gathering and toolbar rendering need to be decoupled? Or perhaps I'm missing something obvious?

Place1:
> Perhaps the data gathering and toolbar rendering need to be decoupled? Or perhaps I'm missing something obvious?

I'd really like to see this. I'm interested in building a real time monitoring webapp (for devs locally) that shows a history of requests and for each request you can view the panel content for each registered panel.

[...]

PR #1253, opened 2020-03-28
Remove use of deprecated request.is_ajax()

jdufresne:
The method was deprecated in upstream commit
https://github.com/django/django/commit/e348ab0d4382b0d7cb0cab9d1261c916c3d0ce6c.

Refs #1032

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -5,6 +5,7 @@ Debug Toolbar middleware
 import re
 from functools import lru_cache
 
+import django
 from django.conf import settings
 from django.utils.module_loading import import_string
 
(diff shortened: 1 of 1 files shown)


tim-schilling:
Shouldn't this replace the is_ajax logic with another type of check that
inspects the request to be an ajax/json request to avoid running too much
of the toolbar unnecessarily?

jdufresne:
Good point.

I've updated the change to use the new `request.accepts()`. Let me know what you think.

jdufresne:
If #1253 lands, this change would be unnecessary as that also removes `.is_ajax()`.

debug_toolbar/middleware.py
```diff
@@ -48,7 +49,12 @@ def __call__(self, request):
         # Decide whether the toolbar is active for this request. Don't render
         # the toolbar during AJAX requests.
         show_toolbar = get_show_toolbar()
-        if not show_toolbar(request) or request.is_ajax():
+        accepts_html = (
+            not request.is_ajax()
+            if django.VERSION < (3, 1)
+            else request.accepts("text/html")
+        )
```
adamchainz, line 56: Can you rejig this to only check `accepts_html` in the case of `show_toolbar` returning `True`? Perhaps split it to another function call. This would keep the `DEBUG=False` pathway minimal.

jdufresne: Good call, thanks.

Now short circuits on `not show_toolbar(request)`.

PR #1254, opened 2020-03-28
Remove use of deprecated Signal.providing_args

jdufresne:
The attribute was deprecated in upstream commit
https://github.com/django/django/commit/769cee525222bb155735aba31d6174d73c271f3c.


matthiask:
Thanks!

PR #1255, opened 2020-03-28
Enable Python warnings during tests

jdufresne:
Helps show deprecation warnings early as well as general programming
edge cases.


matthiask:
Yes, that's useful!

PR #1256, opened 2020-03-28
Remove unnecessary '{% load static %}' template tags

jdufresne:
These templates do not use the {% static ... %} template tag.


tim-schilling:
Thanks!

PR #1257, opened 2020-03-28
Normalize indentation of HTML and CSS files

jdufresne:
Previously, these files used a mix of spaces and tabs and had varying
indentation amounts. Now, add an .editorconfig to the root of the
project with the same values as Django. This normalizes the files to
spaces. CSS files use 4 spaces and HTML files use 2.


matthiask:
Thanks!

PR #1260, opened 2020-04-14
Slightly optimize show_toolbar()

adamchainz:
Avoid inspecting the request if `DEBUG` is not `True`, and remove unnecessary cast to `bool`.


matthiask:
Thanks!

PR #1267, opened 2020-05-17
Use localStorage API to track toolbar state

jdufresne:
Replaces storing the state in cookies. For simple key/value storage, the
API is simpler. Including a cookie with each HTTP request/response is
not necessary in modern browsers.

https://developer.mozilla.org/en-US/docs/Web/API/Storage

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -82,10 +82,6 @@ class DebugToolbarMiddleware:
         ):
             return response
 
-        # Collapse the toolbar by default if SHOW_COLLAPSED is set.
-        if toolbar.config["SHOW_COLLAPSED"] and "djdt" not in request.COOKIES:
-            response.set_cookie("djdt", "hide", 864000)
-
         # Insert the toolbar in the response.
         content = response.content.decode(response.charset)
         insert_before = dt_settings.get_config()["INSERT_BEFORE"]
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -199,19 +199,17 @@
                 document.removeEventListener('mousemove', onHandleMove);
                 if (djdt.handleDragged) {
                     event.preventDefault();
-                    djdt.cookie.set('djdttop', handle.offsetTop, {
-                        path: '/',
-                        expires: 10
-                    });
+                    localStorage.setItem('djdt.top', handle.offsetTop);
                     setTimeout(function () {
                         djdt.handleDragged = false;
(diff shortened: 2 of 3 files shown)


matthiask:
It's definitely nicer and it's the same thing Django's admin does. I still remember the recommendation to wrap localStorage because it might not work or because it silently fails (iOS' incognito mode?). 

What is your view on this @jdufresne ? Do you know a more current recommendation? Or maybe we can just ignore those browsers now?

jdufresne:
> I still remember the recommendation to wrap localStorage because it might not work or because it silently fails (iOS' incognito mode?).

I'm not familiar with this recommendation and I do not have an iOS device to test it, unfortunately.

Judging by the [MDN compatibility chart](https://developer.mozilla.org/en-US/docs/Web/API/Storage#Browser_compatibility), iOS is fully supported. If you have a source for the recommendation, I'm interested to take a look.

In the absence of a source or real life test case, I'm in favor of continuing with the change. I'm happy to have it reverted if something comes up through further testing.

matthiask:
@jdufresne The following post was shared internally where I work:

https://michalzalecki.com/why-using-localStorage-directly-is-a-bad-idea/

It is from 2017 so the information may still be current.

(I also think we should go ahead with this change!)

jdufresne:
Rebased to resolve conflicts.

Thanks for that article. I tested the change in a Firefox private window and Chrome incognito mode. Both worked as expected with no regression. Of course, this isn't a proper test for iOS, but better than nothing.

matthiask:
Additional things coming to mind (no merge blockers to me, just discussion points):

- The `djdt.cookie.get` and `djdt.cookie.set` are documented but now unused. Should we deprecate them, or just remove them and tell third party panel authors to write to localStorage directly instead?
- The `djdt.top` key is using a namespace, maybe use `djdt.show = (true|false)` or something instead of using `djdt = (show|hide)` directly?

jdufresne:
> The djdt.cookie.get and djdt.cookie.set are documented but now unused. Should we deprecate them, or just remove them and tell third party panel authors to write to localStorage directly instead?

Unfortunately, it is still used. It is used in the event handler defined in `toolbar.js` as:

```js
$$.on(djDebug, 'click', '.djDebugPanelButton input[type=checkbox]', function() {
```

I this value is used server side, so I didn't see a good way to get rid of this use. If you're able to come up with one, please let me know.

> The djdt.top key is using a namespace, maybe use djdt.show = (true|false) or something instead of using djdt = (show|hide) directly?

This makes sense to me! I've gone ahead with this change in the latest revision.

matthiask:
Thanks!

PR #1272, opened 2020-05-18
Remove unnecessary guard for empty data-toggle-id value

jdufresne:
data-toggle-id is always defined on djToggleSwitch, so the guard is
unnecessary. The JavaScript can expect correctly constructed HTML.


matthiask:
OK. If a guard would become necessary again it should probably better protect against `container` being null.

PR #1275, opened 2020-05-18
Declare local variable djDebug before use

jdufresne:


matthiask:
Thanks, that's a good cleanup.

PR #1276, opened 2020-05-18
Run eslint during CI

jdufresne:
ESLint is a tool for linting JavaScript code, with the goal of making
code more consistent and avoiding bugs. In many ways, it is similar to
flake8, but for JavaScript.

This replaces the previous use of JSHint, which was not checked during
CI. ESLint is the more popular choice in the JavaScript community and
offers more rules and features.

https://eslint.org/

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,17 @@
+{
+    "env": {
+        "browser": true,
+        "es6": true
+    },
+    "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaVersion": 6
+    },
+    "rules": {
+        "curly": ["error", "all"],
+        "dot-notation": "error",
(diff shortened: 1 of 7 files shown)


matthiask:
I can't reproduce the failure locally. Do you have any idea why the eslint job fails on Travis CI @jdufresne ?

jdufresne:
Fixed.

The version of Node on Travis is too old. It is currently running v8.12.0, but the latest eslint requires ^10.12.0, or >=12.0.0 (ref: https://eslint.org/docs/user-guide/getting-started).

To solve, I have downgraded eslint to the most recent compatible version. Whenever Travis updates their images, we can resume using the latest version of eslint.

matthiask:
Great, thanks!

PR #1282, opened 2020-05-18
Make AJAX views return JSON instead of HTML

jdufresne:
Returning JSON allows for arbitrary data to be returned, not only HTML.
This in turn allows for returning scripts as a list instead of embedded
in the HTML. The script tag embedded in the HTML didn't immediately
execute anyway, so handling it as separate data makes more sense.

Separating the JavaScript makes it simpler to migrate to JavaScript
modules as inserting the new scripts is handled entirely by toolbar.js.

Removes the unused $$.executeScripts() call in the .remoteCall AJAX
handler. No panels using .remoteCall inject scripts in this way.

### DIFF
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -94,6 +94,13 @@ class Panel:
         if self.has_content:
             return render_to_string(self.template, self.get_stats())
 
+    @property
+    def scripts(self):
+        """
+        Scripts used by the HTML content of the panel when it's displayed.
+        """
+        return []
+
     # URLs for panel-specific views
 
     @classmethod
--- a/debug_toolbar/panels/sql/views.py
+++ b/debug_toolbar/panels/sql/views.py
@@ -1,5 +1,5 @@
-from django.http import HttpResponseBadRequest
-from django.template.response import SimpleTemplateResponse
+from django.http import HttpResponseBadRequest, JsonResponse
+from django.template.loader import render_to_string
 from django.views.decorators.csrf import csrf_exempt
 
 from debug_toolbar.decorators import require_show_toolbar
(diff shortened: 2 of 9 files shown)


matthiask:
I like this very much. At the same time, there is no backwards compatibility for third party panels or for pending pull requests (e.g. https://github.com/jazzband/django-debug-toolbar/pull/1250) -- do you think it would be possible to add a fallback? Or is it even worth it? (The backend changes for individual panels are small after all.)

jdufresne:
> there is no backwards compatibility for ... for pending pull requests (e.g. #1250)

I took a very quick look at this PR. It looks like it adds a new JavaScript function `ajaxJson`, so if we move all AJAX requests to JSON, that function will no longer be needed. In short, I think these two PRs are very compatible with one another.

If PR #1250 lands first, I'll happily amend this PR to take it into account. If, on the other hand, this PR lands first, I think PR #1250 could be easily simplified a bit by using `ajax`. I'm happy to help out there as well.

> At the same time, there is no backwards compatibility for third party panels ... -- do you think it would be possible to add a fallback? Or is it even worth it? (The backend changes for individual panels are small after all.)

Is there an existing policy for this? If there is, I'll happily follow it. But, it will complicate the JavaScript code a bit to handle two different server side APIs.

Is it worth it? This is hard to predict and perhaps a personal opinion. On one hand, the compatibility shim adds complication to django-debug-toolbar, but on the other not adding one requires action from third party panel maintainers if they wish to add support for the next djdt version.

IMO, the changes are small enough and the third party panel community is also small enough such that adding the compatibility shim is not worth it. But I'm happy to add it anyway if it helps move the PR forward. So please let me know your preference and I'll follow it.

Something to keep in mind, if a compatibility shim is added, it will eventually be dropped in a future version leading us back to the same question. So we'll expect third party panels to make changes down the road anyway.

Either way, some words should be added to the release notes, I'll go ahead and do that before this should be merged. If we feel the API change is large enough, we can also bump the major release number too.

jdufresne:
Added docs and made a note in `chagnes.rst` in the latest revision.

matthiask:
Thanks for the docs and for the explanation. I wholeheartedly agree.

PR #1287, opened 2020-05-20
Remove undocumented djdt.init() from the exported namespace

jdufresne:
The function is not documented and there is no reason a third party
panel would need to call init() a 2nd time.


matthiask:
djdt.init was made available for supporting TurboLinks and similar software, so there's a good reason for it being part of the public API. Or maybe there exists a better way to accomplish this but exporting init certainly works.

matthiask:
Here's the pull request which exported `djdt.init`: https://github.com/jazzband/django-debug-toolbar/pull/1180

jdufresne:
Thanks for that link. I'll take a deeper look into that issue and try to grok it.

PR #1293, opened 2020-05-21
Move setup information to setup.cfg

jdufresne:
Use a declarative syntax to avoid mixing code and configuration.
Simplifies handling of long description file by reducing some
boilerplate.

For details on this setuptools feature, see:
https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

Django also uses supet.cfg, see commit:
https://github.com/django/django/commit/85efc14a2edac532df1a9ad4dd9b6d4a4dcf583e

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,47 @@
+[metadata]
+name = django-debug-toolbar
+version = 2.2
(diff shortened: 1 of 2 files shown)


matthiask:
Thanks! That's awesome.

PR #1294, opened 2020-05-22
Addiontal readme warning when using docker

daniel-butler:
https://gist.github.com/douglasmiranda/9de51aaba14543851ca3#file-option2-py

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -102,3 +102,15 @@ settings module::
 You can change the logic of determining whether or not the Debug Toolbar
 should be shown with the :ref:`SHOW_TOOLBAR_CALLBACK <SHOW_TOOLBAR_CALLBACK>`
 option.  This option allows you to specify a custom function for this purpose.
+
+.. warning::
+
+    If using Docker the following will set your internal ips correctly only if you are in Debug mode.::
+    
+        if DEBUG:
+            import os  # only if you haven't already imported this
+            import socket  # only if you haven't already imported this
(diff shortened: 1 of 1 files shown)


matthiask:
I also like this option (obviously) https://stackoverflow.com/a/49818040/317346

```
if DEBUG:
    # `debug` is only True in templates if the vistor IP is in INTERNAL_IPS.
    INTERNAL_IPS = type(str('c'), (), {'__contains__': lambda *a: True})()
```

daniel-butler:
That works too! I didn't see that stack overflow Q/A

tim-schilling:
Is there a functional difference between your suggestion @matthiask and setting `SHOW_TOOLBAR_CALLBACK` to the following function:

```
def show_toolbar(request):
    return settings.DEBUG
```

If the INTERNAL_IPS is configured to be a pass-through, then the only thing the default callback does is check `DEBUG`.

matthiask:
> Is there a functional difference between your suggestion @matthiask and setting `SHOW_TOOLBAR_CALLBACK` to the following function:
> 
> ```
> def show_toolbar(request):
>     return settings.DEBUG
> ```
> 
> If the INTERNAL_IPS is configured to be a pass-through, then the only thing the default callback does is check `DEBUG`.

I don't think there's a difference for django-debug-toolbar. Maybe Django itself does some things differently when `INTERNAL_IPS` matches. TBH I don't remember why I went with this `type(...)()` thing mentioned above – it probably worked well enough to be carried from project to project.

PR #1296, opened 2020-05-23
Replace deprecated django.conf.urls.url() with path() and re_path()

jdufresne:
django.urls.path() and re_path() have been available since Django 2.0.

django.conf.urls.url() is deprecated in Django 3.1. When testing against
Django 3.1, fixes warning of the form:

    RemovedInDjango40Warning: django.conf.urls.url() is deprecated in
    favor of django.urls.re_path().


matthiask:
Thanks!

PR #1298, opened 2020-06-23
Add cache's missing touch method

mkay849:
Add cache's missing touch method to prevent "NotImplemented" error of base class


matthiask:
Good find, thanks!

billtrik:
I am having the same issue. 
Is there an estimate on when a new public release that would include this might happen?

matthiask:
@billtrik It's available for testing as `django-debug-toolbar==3.0a2`

Please report any problems or successes you're getting :)

PR #1299, opened 2020-07-09
Allow using ProfilingPanel as non last panel

### LINKED ISSUE #1208: ProfilingPanel info not collected
In my usage the ProfilingPanel info is not being collected/processed, despite being enabled.

I have run in debug mode and the piece that looks most suspicious (in `generate_stats`) is that:
```
root_func = self.stats.get_root_func()
```
is returning a falsey `root_func`.

noamkush:
The profiling panel profiles `process_request` so that's always the root function, but if it's not the last panel, it might have callers. As a result, the panel won't work unless it's last. Since we know what's the root we can use that instead of searching for it.

Fixes #1208


auvipy (changes requested):
shouldn't we update the tests or verify the changes with proper test cases?

noamkush:
Well, this panel is already covered by tests and I'm not sure there's anything new to test for.

PR #1301, opened 2020-08-10
Explicitly set the 'box-sizing' property of the .djDebugClose to 'content-box'.

dimiro1:
It is common for some CSS and reset frameworks[1] to use set the 'box-sizing' to 'border-box' to all elements on the page. This causes the close button inside each panel to show up in the wrong way [2].

This set the 'box-sizing' attribute of the .djDebugClose class to always be  'content-box' [3].

[1] Screen Shot 2020-08-10 at 16 12 11

[2] Screen Shot 2020-08-10 at 16 11 52

[3]  Screen Shot 2020-08-10 at 16 22 00

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -316,6 +316,7 @@
     font-weight: 900;
     font-size: 20px;
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks! I'd prefer moving towards using `border-box` everywhere too, but that's a good fix for now.

PR #1303, opened 2020-08-19
 Simplify isort config using 'profile = black'

jdufresne:
isort now has builtin support for black through its "profiles" feature.

https://timothycrosley.github.io/isort/docs/configuration/profiles/#black

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
 .PHONY: flake8 example test coverage translatable_strings update_translations
 
 style:
-	isort debug_toolbar example tests
+	isort .
 	black --target-version=py35 debug_toolbar example tests setup.py
 	flake8 debug_toolbar example tests
 
 style_check:
-	isort -c debug_toolbar example tests
+	isort -c .
 	black --target-version=py34 --check debug_toolbar example tests setup.py
 
 flake8:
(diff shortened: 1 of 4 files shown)


debug_toolbar/panels/cache.py
```diff
@@ -5,11 +5,13 @@
...
 from django.core import cache
-from django.core.cache import CacheHandler, caches as original_caches
+from django.core.cache import CacheHandler
+from django.core.cache import caches as original_caches
 from django.core.cache.backends.base import BaseCache
 from django.dispatch import Signal
 from django.middleware import cache as middleware_cache
-from django.utils.translation import gettext_lazy as _, ngettext as __
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import ngettext as __
```
matthiask, line 14: I have never seen this convention until now (`__ = ngettext`). I don't necessarily disagree with it but I'm not sure it's well-known enough.

matthiask (commented):
Nice! I didn't know isort had support for profiles.

debug_toolbar/panels/cache.py
jdufresne: I restored the `combine_as_import` option so this line remains the same as before.

matthiask: Yes, I agree. In this case however I think keeping `combine_as_imports` is preferable. 

The `django` profile also uses `combine_as_imports` and the resulting diff is smaller.

PR #1304, opened 2020-08-19
Replaced a link to Line Profiler Third-party panel.

mikekeda:
https://github.com/dmclain/django-debug-toolbar-line-profiler is abandoned (last commit was 5years ago)
It doesn't work with django 3 and django-debug-toolbar 2
Changes - https://github.com/dmclain/django-debug-toolbar-line-profiler/pull/21/files

This fork works with Django 3.1 and django-debug-toolbar 2.2

### DIFF
--- a/docs/panels.rst
+++ b/docs/panels.rst
@@ -172,7 +172,7 @@ to.
 Line Profiler
 ~~~~~~~~~~~~~
 
-URL: https://github.com/dmclain/django-debug-toolbar-line-profiler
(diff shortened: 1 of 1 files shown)


matthiask:
Thank you! It's a bit confusing that the repository and the import path doesn't match the package name on PyPI. 

Maybe you should rename the repository and imports while you're at it? 

(That's not a request for change, just a question/suggestion.)

mikekeda:
I discussed this with @Mogost and decided to change only the name on pypi and not change the internal structure, this will give a smooth transition from the old version to the new one.

PR #1312, opened 2020-09-04
Change close button styling

### LINKED ISSUE #1311: Close button styling in 3.0a2
I am using version 3.0a2, and I noticed that the close button styling is broken when viewing the SQL panel.

In addition to the normal "×", there is a circle, and that circle changes color when hovered over.

Close circle

This issue can be fixed by removing the following CSS rules from `toolbar.css`:

```css
/* line 302: */#djDebug .djdt-panelContent .djDebugClose {
...
    border: 6px solid #ddd;/* Remove this */
    border-radius: 50%;/* and this */
    background: #fff;/* and this */
...
}
```

In addition, the `.djDebugClose:hover` CSS rule on line 321 should change the `color` rather than the `background`.

OutOfFocus4:
Closes #1311

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -307,9 +307,6 @@
     width: 16px;
     line-height: 16px;
     padding: 5px;
-    border: 6px solid #ddd;
-    border-radius: 50%;
-    background: #fff;
     color: #ddd;
     text-align: center;
     font-weight: 900;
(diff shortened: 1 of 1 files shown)


matthiask:
Hmm. 3.0a2 looks like this for me:
image

With the new code it looks like this:
image

This doesn't seem to be an improvement in my opinion. Do you see something else locally?

OutOfFocus4:
The button doesn’t look like that to me. See my screenshot in #1311.

What browser are you using?

matthiask:
Oh, I think I see what's going on here. The problem should have been fixed in https://github.com/jazzband/django-debug-toolbar/commit/f85c14060f9056788f2a6b0dd54101930c7f6c56 already.

matthiask:
I just released 3.0a3, does the new version work for you?

OutOfFocus4:
That fixed it. Thanks!

PR #1319, opened 2020-09-18
Log a warning when django static files settings are misconfigured.

### LINKED ISSUE #1318: FileNotFoundError: No such file or directory (static)
Same as #1123.

I installed django-debug-toolbar. When I start my app, I get this error:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/paul/.dev/police-records/static'

Here's the backtrace:
```
Internal Server Error: /
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/debug_toolbar/middleware.py", line 92, in __call__
    panel.generate_stats(request, response)
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/debug_toolbar/panels/staticfiles.py", line 128, in generate_stats
    "staticfiles_finders": self.get_staticfiles_finders(),
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/debug_toolbar/panels/staticfiles.py", line 140, in get_staticfiles_finders
    for path, finder_storage in finder.list([]):
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/django/contrib/staticfiles/finders.py", line 130, in list
    for path in utils.get_files(storage, ignore_patterns):
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/django/contrib/staticfiles/utils.py", line 23, in get_files
    directories, files = storage.listdir(location)
  File "/Users/paul/.dev/police-records/.direnv/python-3.8.5/lib/python3.8/site-packages/django/core/files/storage.py", line 316, in listdir
    for entry in os.scandir(path):
FileNotFoundError: [Errno 2] No such file or directory: '/Users/paul/.dev/police-records/static'
```

tim-schilling:
If the django static files are misconfigured in some way or a
directory doesn't exist, the toolbar should still run, but
attempt to alert the user to the issue.

Fixes #1318

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -74,7 +74,18 @@ def check_middleware(app_configs, **kwargs):
                 id="debug_toolbar.W003",
             )
         )
+    return errors
 
+
+@register
+def check_panel_configs(app_configs, **kwargs):
+    """Allow each panel to check the toolbar's integration for their its own purposes."""
+    from debug_toolbar.toolbar import DebugToolbar
+
+    errors = []
+    for panel_class in DebugToolbar.get_panel_classes():
+        for check_message in panel_class.run_checks():
+            errors.append(check_message)
     return errors
 
 
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -211,3 +211,15 @@ class Panel:
 
         Does not return a value.
         """
+
+    @classmethod
+    def run_checks(cls):
+        """
+        Check that the integration is configured correctly for the panel.
+
+        This will be called as a part of the Django checks system when the
+        application is being setup.
+
+        Return a list of :class: `django.core.checks.CheckMessage` instances.
+        """
+        return []
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -3,6 +3,7 @@ from os.path import join, normpath
 
 from django.conf import settings
 from django.contrib.staticfiles import finders, storage
+from django.core.checks import Warning
 from django.core.files.storage import get_storage_class
 from django.utils.functional import LazyObject
 from django.utils.translation import gettext_lazy as _, ngettext as __
@@ -138,17 +139,21 @@ class StaticFilesPanel(panels.Panel):
         """
         finders_mapping = OrderedDict()
         for finder in finders.get_finders():
-            for path, finder_storage in finder.list([]):
-                if getattr(finder_storage, "prefix", None):
-                    prefixed_path = join(finder_storage.prefix, path)
-                else:
-                    prefixed_path = path
-                finder_cls = finder.__class__
-                finder_path = ".".join([finder_cls.__module__, finder_cls.__name__])
-                real_path = finder_storage.path(path)
-                payload = (prefixed_path, real_path)
-                finders_mapping.setdefault(finder_path, []).append(payload)
-                self.num_found += 1
+            try:
+                for path, finder_storage in finder.list([]):
+                    if getattr(finder_storage, "prefix", None):
+                        prefixed_path = join(finder_storage.prefix, path)
+                    else:
+                        prefixed_path = path
+                    finder_cls = finder.__class__
+                    finder_path = ".".join([finder_cls.__module__, finder_cls.__name__])
+                    real_path = finder_storage.path(path)
+                    payload = (prefixed_path, real_path)
+                    finders_mapping.setdefault(finder_path, []).append(payload)
+                    self.num_found += 1
+            except OSError:
+                # This error should be captured and presented as a part of run_checks.
+                pass
         return finders_mapping
 
     def get_staticfiles_dirs(self):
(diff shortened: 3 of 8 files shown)


tim-schilling:
Hmm, this would require users to make a change to their `LOGGERS` setting. I'm not sure if we want to rely on that.

tim-schilling:
We could do something like the following in the staticfiles panel.

```
Path                                                | Location
--                                                  | --
test.css                                            | /home/.../Projects/debug_toolbar/example/static/test.css
django.contrib.staticfiles.finders.FileSystemFinder | [Errno 2] No such file or directory: '/home/.../Projects/debug_toolbar/example/invalid'
```

Another option would be to write a formal django Check class into the toolbar that would validate that the `STATICFILEDIRS` exist.

I'm leaning towards the that last option, the django Check.

matthiask:
Yeah, requiring users to configure `LOGGING` is probably a non-starter. I also went back to print statements in [django-http-fallback-storage](https://github.com/matthiask/django-http-fallback-storage/blob/master/http_fallback_storage.py#L17-L28) because I couldn't trust myself to keep `LOGGING` correct and up-to-date...

I'm asking myself now whether silently skipping invalid entries in `STATICFILES_DIRS` wouldn't be "good enough" for us? After all, the panel lists files it could find and it makes some sense to say that no files could be found in non-existing directories.

tim-schilling:
I don't think silently skipping invalid entries is sufficient. The problem is that if the non-existentent directory is the first in `STATICFILES_DIRS`, then the loop breaks for the remaining directories. This would cut the entires for the `FileSystemFinder` short, or in this case, it'd be empty even though files are being rendered properly.

We could implement our own version of `FileSystemFinder.list` that would allow us to capture the specific errors on the paths, but that seem unappealing as well.

jdufresne:
> Another option would be to write a formal django Check class into the toolbar that would validate that the STATICFILEDIRS exist.

Do you mean as an upstream contribution to Django? IMO, that makes more sense than a change to Debug Toolbar.

tim-schilling:
> Do you mean as an upstream contribution to Django? IMO, that makes more sense than a change to Debug Toolbar.

No, I didn't do a good job of explaining myself. I ended up writing what I was thinking and pushed a new version of this PR.

Basically rather than logging something to the console while the application is running, we'll log it to the console at start-up via the checks system.

tim-schilling:
This will require an update after #1345 is merged.

jdufresne:
Oo. I do like the idea that panels have a `run_checks` entry point.

debug_toolbar/apps.py
```diff
@@ -72,7 +72,18 @@ def check_middleware(app_configs, **kwargs):
...
 
+@register
+def check_panel_configs(app_configs, **kwargs):
+    """Allow each panel to check the toolbar's integration for their its own purposes."""
+    from debug_toolbar.toolbar import DebugToolbar
+
+    errors = []
+    for panel_class in DebugToolbar.get_panel_classes():
+        for check_message in panel_class.run_checks():
+            errors.append(check_message)
```
jdufresne, line 86:
```suggestion
        errors += panel_class.run_checks()
```

debug_toolbar/panels/__init__.py
```diff
@@ -211,3 +211,15 @@ def generate_server_timing(self, request, response):
...
+
+    @classmethod
+    def run_checks(cls):
+        """
+        Check that the integration is configured correctly for the panel.
+
+        This will be called as a part of the Django checks system when the
+        application is being setup.
+
+        Return a list of :class: `django.core.checks.CheckMessage` instances.
```
jdufresne, line 223:
```suggestion
        Return a list of :class:`django.core.checks.CheckMessage` instances.
```

matthiask:
I like this too! Just one question: Is there a reason why we cannot use `@django.core.checks.register` like here https://github.com/jazzband/django-debug-toolbar/blob/916172c6a8124c65cce48131e98fd86e06329d08/debug_toolbar/apps.py#L16-L18 – maybe because individual panels only get a chance to register their checks *after* Django completed checking?

tim-schilling:
@matthiask I didn't consider trying to have the panels register their own checks, but that might be possible. We would need to have the panel verify that it's actually enabled before running. And I'm not sure how to cleanly have that happen within the Panel class rather than requiring the subclasses do the validation themselves.

Let me know what you think.

debug_toolbar/panels/staticfiles.py
```diff
@@ -138,17 +139,21 @@ def get_staticfiles_finders(self):
...
+                    else:
+                        prefixed_path = path
+                    finder_cls = finder.__class__
+                    finder_path = ".".join([finder_cls.__module__, finder_cls.__name__])
+                    real_path = finder_storage.path(path)
+                    payload = (prefixed_path, real_path)
+                    finders_mapping.setdefault(finder_path, []).append(payload)
+                    self.num_found += 1
+            except OSError:
+                # This error should be captured and presented as a part of run_checks.
```
matthiask, line 155: I'm not sure whether this is an old comment which is not relevant anymore or whether there's anything left to do?

matthiask (commented):
That's an excellent point. I think `run_checks` is better than inspecting settings to find out whether the panel is active at all.

Only one question left :) Thanks!

tim-schilling:
Since we're eating the error I wanted to explain myself to other devs
rather than having them grok the panels code entirely.

matthiask:
Makes sense, thanks! My fault.

tim-schilling:
That was a good question. If I came off as abrasive I didn't mean to.
Thanks for keeping up with all the PRs we've been creating.

>

PR #1321, opened 2020-09-20
Remove 'font-variant: small-caps' CSS property

jdufresne:
The capitalization of text is misleading. For example, the URL `/LOGIN/` and
`/login/` are different and should be read differently. Another example, the
view `LOGINVIEW` and LoginView` are different.

For the most accurate display of the information, drop the small-caps property.


jdufresne:
### Originally

debug-with-small-caps

### Now

debug-without-small-caps

matthiask:
Thank you! That's much better.

PR #1324, opened 2020-09-20
Remove support for HTML in panel titles

jdufresne:
This features was used by a single panel only and has the potential for
mistakenly introducing XSS vulnerabilities.

Support for HTML in titles first appeared in
5015057cfe3448028ff79ca36a96f1baa4224226.

### DIFF
--- a/debug_toolbar/panels/settings.py
+++ b/debug_toolbar/panels/settings.py
@@ -24,7 +24,7 @@ class SettingsPanel(Panel):
     nav_title = _("Settings")
 
     def title(self):
-        return _("Settings from <code>%s</code>") % settings.SETTINGS_MODULE
+        return _("Settings from %s") % settings.SETTINGS_MODULE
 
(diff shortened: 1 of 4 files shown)


jdufresne:
### previously

title

matthiask:
Thanks! I didn't know that the title of panels could contain HTML, and I think that's really surprising.

I think it would be better if the title wasn't marked as safe in `debug_toolbar/templates/debug_toolbar/includes/panel_content.html` and we instead would change the settings panel to stop using `<code>` at all. All other panels do not use HTML tags. 

What do you think?

jdufresne:
That works for me and is probably safer in the long run. I'll work towards that change.

Just as a point of reference, this HTML in titles first appeared in 5015057cfe3448028ff79ca36a96f1baa4224226.

matthiask:
Thank you!

PR #1325, opened 2020-09-20
Make it simpler to run the example project

jdufresne:
Now, can run the example project with a single command:

    make example

The make target now migrates the database and creates a superuser. This removes
some manual steps when doing simple manual testing of Django Debug Toolbar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,9 @@ flake8:
 	flake8
 
 example:
+	python example/manage.py migrate --noinput
+	-DJANGO_SUPERUSER_PASSWORD=p python example/manage.py createsuperuser \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -13,6 +13,9 @@ flake8:
 	flake8
 
 example:
+	python example/manage.py migrate --noinput
+	-DJANGO_SUPERUSER_PASSWORD=p python example/manage.py createsuperuser \
+		--noinput --username=$(USER) --email=$(USER)@mailinator.com
```
matthiask, line 18: That looks very useful! Sholdn't this be `"$(USER)"` though, for safety? OTOH people will only hack themselves 😅

jdufresne: Good point :+1:. I added quotes in the latest revision.

PR #1328, opened 2020-09-21
Improve and simplify the "stats" CSS on the SQL panel

jdufresne:
The previous version was misaligned and could sometimes display an awkwardly
placed list bullet point.

The end result is simpler/less CSS.


jdufresne:
### before

old-stats-sql

### after

new-stats-sql

PR #1331, opened 2020-09-21
Allow empty data with content type application/json

### LINKED ISSUE #1330: Empty request with json content type fails with a JSONDecodeError when using the history panel
If an API request is made with an empty body and the content type `application/json`, the history panel raises an error and the whole API request fails.

For instance:

```
GET /api/whatever
Content-Type: application/json

Body: 
```

Result in `500 Internal Server Error`, `JSONDecodeError`, `Expecting value: line 1 column 1 (char 0)`

We could argue that an empty body is not a valid `application/json` body (https://greenbytes.de/tech/webdav/rfc7159.html#values), but in this case, I believe it's not the role of django debug toolbar to raise an error if django accepts this kind of requests.

MDziwny:
fix #1330

### DIFF
--- a/tests/panels/test_history.py
+++ b/tests/panels/test_history.py
@@ -28,13 +28,21 @@ class HistoryPanelTestCase(BaseTestCase):
         self.assertEqual(data["foo"], "bar")
 
     def test_post_json(self):
-        self.request = rf.post(
-            "/", data={"foo": "bar"}, content_type="application/json"
-        )
-        response = self.panel.process_request(self.request)
-        self.panel.generate_stats(self.request, response)
-        data = self.panel.get_stats()["data"]
-        self.assertEqual(data["foo"], "bar")
+        for data, expected_stats_data in (
+            ({"foo": "bar"}, {"foo": "bar"}),
+            ("", {}),
+        ):
+            with self.subTest(data=data):
+                self.request = rf.post(
+                    "/",
+                    data=data,
+                    content_type="application/json",
+                    CONTENT_TYPE="application/json",  # Force django test client to add the content-type even if no data
+                )
+                response = self.panel.process_request(self.request)
+                self.panel.generate_stats(self.request, response)
+                data = self.panel.get_stats()["data"]
+                self.assertDictEqual(data, expected_stats_data)
 
     def test_urls(self):
         self.assertEqual(
--- a/debug_toolbar/panels/history/panel.py
+++ b/debug_toolbar/panels/history/panel.py
@@ -48,7 +48,11 @@ class HistoryPanel(Panel):
                 data = request.POST.copy()
             # GraphQL tends to not be populated in POST. If the request seems
             # empty, check if it's a JSON request.
-            if not data and request.META.get("CONTENT_TYPE") == "application/json":
+            if (
+                not data
+                and request.body
+                and request.META.get("CONTENT_TYPE") == "application/json"
+            ):
                 # Python <= 3.5's json.loads expects a string.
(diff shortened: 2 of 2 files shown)


tests/panels/test_history.py
```diff
@@ -28,13 +28,21 @@ def test_post(self):
...
+        for data, expected_stats_data in (
+            ({"foo": "bar"}, {"foo": "bar"}),
+            ("", {}),
+        ):
+            with self.subTest(data=data):
+                self.request = rf.post(
+                    "/",
+                    data=data,
+                    content_type="application/json",
+                    CONTENT_TYPE="application/json",  # Force django test client to add the content-type even if no data
```
matthiask, line 40: Are both lines necessary? Or is it possible to remove one of the content_type parameters? (I didn't check what they do, do maybe it isn't a very smart question.)

matthiask:
Thanks, this looks very good!

tests/panels/test_history.py
tim-schilling: I think the second is to force the header option.

matthiask: Ah yes I see. The first makes Django automatically encode the `data` as JSON, the second adds the header even if `data` is falsy. 

It's certainly convenient that Django automatically encodes the data if the content_type has some particular value. Sometimes the interdependence of different arguments just isn't obvious (enough) to me.

MDziwny: Yes, unfortunately, this looks more like a trick, the django client don't add the `content_type` if there is no data, so I add manually (extra kwargs which are not known by the test client are considered as extra headers).

israel-tk:
If you send content type = json but don't send JSON the system fails as well. Pseudo-code in django-debug-toolbar 3.1.1:
```
>>> Client(enforce_csrf_checks=True).post(
            "some-valid-url", data="invalid-json", content_type="application/json",
        )
/usr/local/lib/python3.8/site-packages/django/test/client.py:742: in post
    response = super().post(path, data=data, content_type=content_type, secure=secure, **extra)
/usr/local/lib/python3.8/site-packages/django/test/client.py:404: in post
    return self.generic('POST', path, post_data, content_type,
/usr/local/lib/python3.8/site-packages/django/test/client.py:470: in generic
    return self.request(**r)
/usr/local/lib/python3.8/site-packages/django/test/client.py:710: in request
    self.check_exception(response)
/usr/local/lib/python3.8/site-packages/django/test/client.py:571: in check_exception
    raise exc_value
/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py:47: in inner
/usr/local/lib/python3.8/site-packages/debug_toolbar/middleware.py:67: in __call__
    panel.generate_stats(request, response)
/usr/local/lib/python3.8/site-packages/debug_toolbar/panels/history/panel.py:57: in generate_stats
    data = json.loads(
/usr/local/lib/python3.8/json/__init__.py:357: in loads
    return _default_decoder.decode(s)
/usr/local/lib/python3.8/json/decoder.py:337: in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
```

matthiask:
Oh right, that's a new case. Do you want submit a pull request @israel-tk ? Or maybe at least open an issue so that this problem isn't forgotten.

israel-tk:
> Oh right, that's a new case. Do you want submit a pull request @israel-tk ? Or maybe at least open an issue so that this problem isn't forgotten.

Done creating a new issue: https://github.com/jazzband/django-debug-toolbar/issues/1403

PR #1334, opened 2020-09-22
Replace indicator.png with CSS

jdufresne:
The image is simply a colored triangle. The same effect can be accomplished
using CSS which removes a web request.


matthiask:
Nice!

PR #1338, opened 2020-09-25
Introduce Prettier to format JavaScript and CSS files

jdufresne:
Prettier is an opinionated formatting tool for JavaScript and CSS.
Similar to how Black is for Python, Prettier automates all formatting
decisions for JS and CSS files, thus eliminating differences in
preferences. Contributors can use "make style" just like before to
restyle files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: bionic
 language: python
 cache: pip
 matrix:
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,17 @@
 .PHONY: flake8 example test coverage translatable_strings update_translations
 
-style:
+PRETTIER_TARGETS = '**/*.(css|js)'
+
+style: package-lock.json
 	isort .
 	black --target-version=py35 .
 	flake8
+	npx prettier --ignore-path .gitignore --write $(PRETTIER_TARGETS)
 
-style_check:
+style_check: package-lock.json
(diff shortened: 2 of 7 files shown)


tim-schilling (approved):
Thanks for adding this!

tim-schilling (changes requested):
I didn't notice that Travis broke on the `style` test. Sorry for the extra email.

jdufresne:
I'm a bit confused why this fails on Travis. I do not get the same error when running locally. Any ideas?

tim-schilling:
Hmm, the additional logging was eaten somewhere. I can't reproduce this locally with node 12.16.2 that Travis claims to use.

jdufresne:
Ah ok. This is failing because Travis runs an old version of node.js, but prettier requires 10+. I'll see what our options are here.

jdufresne:
I have updated the Travis `dist` to `bionic`, green again :slightly_smiling_face: .

tim-schilling:
Good find!

matthiask:
Thanks, this is awesome! I didn't merge this yet because of all the pull requests in flight. Let's merge this when things have calmed down a bit :-)

matthiask:
I've cherry-picked this change and re-run `prettier` on the current code in `master`. The code change is still attributed to you @jdufresne (as it should be) Thanks!

PR #1340, opened 2020-09-25
Unify show/hide JS logic through the djdt-hidden class

jdufresne:
Using a class -- rather than manipulating inline CSS -- is more robust
as it doesn't assume the hidden object is styled as 'display: block;'.
The djdt-hidden class already existed, so use it for all hide/show code.


tim-schilling:
I'll admit, I'm a bit cautious after the last HTML/JS change we made that broke something else. What tests did you all run as a part of verifying this code change?

jdufresne:
I tested this by running the selenium tests (all pass) and by doing manual testing in the example app. I tried to hit all areas I know show/hide HTML. If you happen to find a bug, let me know and I'll happily add Selenium test to cover it.

tim-schilling:
Nope, that's good enough for me. Thank you for providing that information!

tim-schilling (approved):
Looks good to me!

PR #1341, opened 2020-09-25
Remove never rendered <li id="djDebugButton">DEBUG</li>

jdufresne:


tim-schilling:
I'm imagining that this could be rendered if someone had their toolbar cookies such that it was open, but changed their project/settings such that none of the panels were enabled. Given that, the claim that it's never rendered isn't entirely true.

The question becomes do we need to inform the user that they have the toolbar enabled, but none of the panels? This might be the second case (first is #1319) of desiring a Django check regarding configuration.

matthiask:
Hmm we already have system checks specific to django-debug-toolbar. I think it would be appropriate to raise warnings when debug toolbar is running with disabled panels only.

jdufresne:
> I'm imagining that this could be rendered if someone had their toolbar cookies such that it was open

I thought the same and so I tested it, but it still did not render. If you find different, let me know. I'll add a test.

>  I think it would be appropriate to raise warnings when debug toolbar is running with disabled panels only.

Excellent idea! I'll add that before this is ready for review.

jdufresne:
I have add the system check with docs and tests. Thanks for the suggestion!

PR #1344, opened 2020-09-26
Remove "Requirements Status" badge from README.rst

jdufresne:
The badge always displays as "outdated". The casual observer may
conclude that Django Debug Toolbar is not maintained. As the library
dependencies aren't pinned to an upper bound, there really isn't
anything that can go out of date.

Better to avoid the mixed message.


tim-schilling (approved):
Makes sense to me.

PR #1345, opened 2020-09-26
Document Django Debug Toolbar system checks

jdufresne:
Follows a layout that is similar to Django's documentation:
https://docs.djangoproject.com/en/dev/ref/checks/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ htmlcov
 .tox
 node_modules
 package-lock.json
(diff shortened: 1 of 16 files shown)


jdufresne:
This is also included #1341 so we can close this if that gets merged.

tim-schilling (approved):

matthiask:
Thanks! This is already merged as part of #1341

PR #1346, opened 2020-09-26
Prefer literal over HTML entity

jdufresne:
Literals are easier to read when viewing the source (don't need to
memorize code points). The files are encoded in UTF-8 and all modern
browser support UTF-8, so it is compatible.


tim-schilling (approved):
This makes sense to me.

matthiask:
Thanks! Indeed, nobody knows what `&#187;` should be :) I'd understand `&raquo;` but the literal is so much better.

PR #1347, opened 2020-09-26
Replace .map() with .forEach() in JavaScript

jdufresne:
The code is not mapping the data to a new array. It was being used for
iteration only. Therefore, forEach() is the more appropriate function
which does not build a new Array.


matthiask:
I like this. But, the new Chromium-based Edge browser doesn't support array destructuring by default yet (https://caniuse.com/mdn-javascript_operators_destructuring_rest_in_arrays)

jdufresne:
Thanks, that makes sense. I've reverted the destructuring in that latest revision, but kept the `forEach`.

matthiask:
Thanks!

PR #1348, opened 2020-09-26
Simplify fetch() calls in JavaScript

jdufresne:
The fetch() init object does not have a property "url", so remove it.

The element being either a button or a element is mutually exclusive, so
use an "else if".

When a button is pressed, an enclosing form element already exist, so
remove the guard.

Assign the object properties in the object literal instead of after.


matthiask:
I think this looks good, thanks!

tim-schilling (approved):
Looks good! I verified things are working locally.

matthiask:
🎉

PR #1350, opened 2020-09-27
Move history panel specific JavaScript to its own file

jdufresne:
Like the timer panel, the JavaScript that only applies to the history
panel should be contained within its own JavaScript module. This avoids
loading the history event handler unnecessarily.

The utility functions $$ and ajax have been moved to a utils.js es6
module so they can be imported by the main toolbar.js file as well as in
panel specific scripts.

The new utils.js module also helps move Django Debug Toolbar's
JavaScript files to a more modularized approached.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,8 @@
     },
     "extends": "eslint:recommended",
     "parserOptions": {
-        "ecmaVersion": 6
+        "ecmaVersion": 6,
+        "sourceType": "module"
     },
     "rules": {
         "curly": ["error", "all"],
--- a/debug_toolbar/panels/history/panel.py
+++ b/debug_toolbar/panels/history/panel.py
@@ -5,6 +5,7 @@ from collections import OrderedDict
 from django.conf import settings
 from django.http.request import RawPostDataException
 from django.template.loader import render_to_string
+from django.templatetags.static import static
 from django.urls import path
 from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
@@ -94,3 +95,9 @@ class HistoryPanel(Panel):
                 ),
             },
         )
+
+    @property
+    def scripts(self):
+        scripts = super().scripts
+        scripts.append(static("debug_toolbar/js/history.js"))
+        return scripts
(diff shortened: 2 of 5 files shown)


matthiask:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import the support for `import`ing statically should be good enough for this.

It has at times been useful to add a few polyfills and run the odd test with the debug toolbar in Internet Explorer. Are we sure we want to leave IE behind for good, in debugging tools? I'm somewhere between an enthusiastic "of course" and a wary "will debugging be even worse in the future?" right now.

jdufresne:
IE is nearing end of life, so I think so. But I'm just one user, others may have a different perspective or needs. In the projects I work on, I have dropped IE support, but I know not everyone has that luxury. It will be nice to continue to evolve the JavaScript as the community supports new features, especially real module support. Users needing to support IE can continue to use an older version of Debug Toolbar as well.

But if it not time yet, no big deal.

tim-schilling:
The older version of the toolbar may not support all versions of Django so that might not always be an option. And I don't know if we have the maintaining effort pool to start maintaining multiple versions of the toolbar, if there was an argument to go that route. 

If [Microsoft's apps end of life support for IE11 is Aug 2021](https://techcommunity.microsoft.com/t5/microsoft-365-blog/microsoft-365-apps-say-farewell-to-internet-explorer-11-and/ba-p/1591666), that is a fair amount of time to make some developers' lives rougher. If there's no other way to do this, then I'm fine with it, but if we can work around it to support IE a little longer, I think some poor soul working with IE may be better off.

jdufresne:
FWIW, the Django admin doesn't support IE and has started adopting JavaScript that is not compatible with it. Just once example.

https://docs.djangoproject.com/en/dev/faq/admin/

> The admin provides a fully-functional experience to the recent versions of modern, web standards compliant browsers. On desktop this means Chrome, Edge, Firefox, Opera, Safari, and others.

tim-schilling:
That certainly makes the argument to drop IE support stronger. I take back
my earlier active reservation.

>

matthiask:
Yeah, true. The probability that I'll personally suffer from dropping IE is very small anyway since we dropped IE support several years ago 👍

tim-schilling (approved):

PR #1351, opened 2020-09-27
Simplify DebugToolbar.store()

jdufresne:
django.utils.datastructures.SortedDict has been unused since
a0569d7b473a2292afbda35b4066bc0aa4429823.

Can access class variables using self so drop unnecessary type() call.

range() supports passing a first argument larger than the second, in
which case it doesn't yield any values.

    >>> list(range(2, 1))
    []


matthiask:
Thanks! I added c543358797bf12a3a45ae6f6ab42e163a4ccc44e; this should help future readers understand that `self._store` is a class variable.

PR #1353, opened 2020-09-29
Fix #1184 -- Replace loading gif with css animation

### LINKED ISSUE #1184: Missing 2x highres graphics
I noticed that the icons are 72dpi PNGs. I'd like to switch them to SVGs, to support all pixel densities.

@idan after some digging, it seems that you originally provided the graphics. Do you happen have any vector graphics, or should I just recreate them (as good as I can)?

codingjoe:
You can test it by loading a page, then disabling the network in the browsers debug console followed by opening any toolbar tab.

The whole thing looks like this:
Screenshot 2020-09-29 at 14 46 06


debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -236,6 +236,24 @@
...
+    border-radius: 50%;
+    border-top: 6px solid #ffe761;
+    width: 38px;
+    height: 38px;
+    -webkit-animation: spin 2s linear infinite; /* Safari */
+    animation: spin 2s linear infinite;
+}
+
+/* Safari */
+@-webkit-keyframes spin {
```
jdufresne, line 249: According to MDN, Safari supports `@keyframes`, so I think we should drop this vendor prefix:

https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes

We made some efforts in the past to drop vendor prefixes: 22dbea38dd4d52ea2eb8783a2c8c8a5c8a0442a1

Worst case for older browsers, this should degrade gracefully so I don't think it will be a major issue for them.

codingjoe: Sure thing, I removed them.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -236,6 +236,17 @@
 #djDebug .djDebugPanelContent .djdt-loader {
     display: block;
```
jdufresne, line 237: Now that this is `<div>`, can remove display block as it is the default.

jdufresne (approved):
Nice! This is really neat. Thanks for contributing it.

debug_toolbar/static/debug_toolbar/css/toolbar.css
codingjoe: Good point, I removed it.

codingjoe:
@jdufresne I removed the display property as suggested. I guess you are most familiar with the merge and release proceedings. I'd leave that to you, if that's ok.

matthiask:
Thank you, that's awesome!

PR #1354, opened 2020-10-02
Make history views accept GET requests

jdufresne:
The history views are safe(they do not have server-side side-effects).
Therefore, these should be GET requests. This removes the need for the
csrf_exempt decorator.

The new JavaScript function ajaxForm() is a helper that uses an HTML
form element to collect data to make a fetch() call.


matthiask:
Thanks!

PR #1357, opened 2020-10-02
Drop '--loglevel debug' from prettier command

jdufresne:
The output is too verbose. The extra logging was added during
development to diagnose a problem, but as an oversight it wasn't removed
after the issue was fixed.


matthiask:
Makes sense :)

PR #1359, opened 2020-10-04
Remove unused a.button CSS

jdufresne:
No anchor tag includes the class "button". Slightly simplifies and
reduces the CSS.


matthiask:
Nice!

PR #1360, opened 2020-10-04
Simplify djToggleSwitch HTML

jdufresne:
Every instance of djToggleSwitch uses the same text to represent open &
close. Therefore, parameterizing it adds unnecessary HTML to each use. Now,
simply assume "-" and "+" will be used.

Should new uses of djToggleSwitch be added, this reduces the amount of
copy & paste necessary.


matthiask:
I haven't found a reason why these attributes were added in the first place; removing them (and simplifying the code) makes sense to me. Thanks!

PR #1361, opened 2020-10-04
Fix "go back" button in djDebugWindow

jdufresne:
Regression introduced in 344baae7bce6570b93807c078c15652aa24699bc where
the return statement in visible() mistakenly went missing.

A new Selenium test was added to help ensure the regression doesn't
return.


tim-schilling (approved):
Good find.

PR #1362, opened 2020-10-04
Correct inaccurate comments in DebugToolbarLiveTestCase

jdufresne:
Likely the result of a copy & paste that wasn't edited.


matthiask:
Looks like it!

PR #1363, opened 2020-10-04
Combine DebugToolbarSystemChecksTestCase with ChecksTestCase

jdufresne:
Both classes aim to test system checks.

Using a separate module is nice to avoid test_integrations.py from
becoming too big.


tim-schilling (approved):
Good call.

PR #1369, opened 2020-10-04
Simplify SQL HTML and CSS

jdufresne:
Remove ineffective z-index. Using the Firefox developer tools reports:

    z-index has no effect on this element as it's not a positional element.

Per MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/word-break)
using "word-break: break-word" is deprecated. Replace usage with
"overflow-wrap: anywhere".

This allows removing some CSS and one div element.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -413,10 +413,6 @@
     border-radius: 3px;
 }
 
-#djDebug .djDebugSqlWrap {
-    position: relative;
-}
-
 #djDebug .djDebugCollapsed {
     color: #333;
 }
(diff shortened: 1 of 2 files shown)


matthiask:
I haven't checked the specific location yet. But, `overflow-wrap: break-word` works and a change has been accepted here https://github.com/django/django/commit/01c6caa3e68047dbbb3292bed4b11c101c97a918 – if there is a really long word in the SQL somewhere we might want to continue inserting line breaks?

jdufresne:
Good call. I've moved the CSS to use `overflow-wrap: anywhere` which MDN recommends over the deprecated one.

Here is the result which in my testing looks identical to the previous result:

long-word

PR #1371, opened 2020-10-04
Improve developer ergonomics of DebugToolbarLiveTestCase

jdufresne:
Add method "get()". This removes the need for new tests to repetitively
call:

    self.selenium.get(self.live_server_url + "...")

Now:

    self.get("...")

Add new property "wait". This removes the need for new tests to
repetitively call:

    WebDriverWait(self.selenium, timeout=10).until(...

Now:

    self.wait.until(...

The timeout value has also been reduced as 10s is too long. In a local
test environment, these tests should pass very quickly. In the event
that something goes wrong, provide feedback to the developer quicker
rather than waiting a full 10s.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -338,8 +338,15 @@ class DebugToolbarLiveTestCase(StaticLiveServerTestCase):
         cls.selenium.quit()
         super().tearDownClass()
 
+    def get(self, url):
+        self.selenium.get(self.live_server_url + url)
+
+    @property
+    def wait(self):
+        return WebDriverWait(self.selenium, timeout=3)
+
     def test_basic(self):
-        self.selenium.get(self.live_server_url + "/regular/basic/")
+        self.get("/regular/basic/")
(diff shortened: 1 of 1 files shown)


matthiask:
What is the chance of the shorter wait generating spurious test failures?

jdufresne:
> What is the chance of the shorter wait generating spurious test failures?

Was this the cause of past test failures? Is that why the timeout is so high? If so, perhaps we can use a value of 2 or 3. 10s still feels very long to me.

matthiask:
I remember that other projects had problems with short timeouts; I'm not sure whether these problems will affect django-debug-toolbar too.

It's mainly a problem when tests fail and contributors do not have the right to restart the CI build. I think 2-3 seconds might be more safe than 1sec, but that's just a gut feeling and maybe it's not necessary at all.

jdufresne:
Makse sense. Lets play it safe with 3s. Done in the latest revision.

PR #1372, opened 2020-10-04
Move eslint/flake8 to style/style_check make targets

jdufresne:
This makes it more likely to for contributors to catch errors alongside
style errors before hitting CI servers. Developers need only run one
make target rather than several.


matthiask:
Excellent, I long wondered why we had so many targets.

PR #1373, opened 2020-10-04
Move repetitive HTML in request.html to include-able template

jdufresne:
The two column table used to display key/value data structures was
repeated several times. To ensure these tables always remain in sync,
move to a common reusable template.

The POST table was missed when these were altered in
d45808fedcfd9d760050f87347e78dc9cb5108cb, which helps show how factoring
out the common HTML helps future maintenance.


matthiask:
Thanks!

PR #1377, opened 2020-10-10
Prefer function declarations over assigning function expressions

jdufresne:
The non-inline functions can be declared at parse-time rather than at
run-time, so allow the JavaScript engine to do so. This also always
assigns a name to the function in the console and tracebacks.


matthiask:
Thanks!

PR #1378, opened 2020-10-11
Simplify CSS classes and rules on th/td elements

jdufresne:
These CSS classes don't style th element, but td elements. So remove
from th elements for slightly smaller HTML.

The rule:

    #djDebug .djdt-panelContent thead th {
        white-space: nowrap;
    }

duplicates the rule on line 350 with the same selector. A it is
redundant, remove it.

djdt-query had no CSS and no JavaScript associated with it. As it has no
effect, remove it.

djdt-timeline only adds a width. There already exist a series of generic
classes to set widths "djdt-width-*', so use this pattern instead of a
new one-off CSS class.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -514,10 +514,6 @@
     color: #333;
 } /* Comment.Preproc */
 
-#djDebug .djdt-timeline {
-    width: 30%;
-}
-
 #djDebug svg.djDebugLineChart {
     width: 100%;
     height: 1.5em;
@@ -534,9 +530,6 @@
     stroke: #94b24d;
 }
 
-#djDebug .djdt-panelContent thead th {
-    white-space: nowrap;
-}
 #djDebug .djDebugRowWarning .djdt-time {
     color: red;
 }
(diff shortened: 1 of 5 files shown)


debug_toolbar/static/debug_toolbar/js/toolbar.timer.js
```diff
@@ -20,7 +20,7 @@ function addRow(stat, endStat) {
             "<td>" +
             stat.replace("Start", "") +
             "</td>" +
-            '<td class="djdt-timeline"><svg class="djDebugLineChart" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 100 5" preserveAspectRatio="none"><rect y="0" height="5" fill="#ccc" /></svg></td>' +
+            '<td><svg class="djDebugLineChart" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 100 5" preserveAspectRatio="none"><rect y="0" height="5" fill="#ccc" /></svg></td>' +
```
matthiask, line 23: I'm confused. Do these `td`'s not require a `djdt-width-30` CSS class now?

jdufresne: Correct. The styling was already handled by the `<colgroup>` element which applies to the entire column:

timer

PR #1379, opened 2020-10-11
Move toolbar.timer.js → timer.js to match naming convention

jdufresne:
All other panel specific JavaScript files use the same name as their
panel. So too should the timer panel.


matthiask:
Thanks!

PR #1380, opened 2020-10-11
Remove unnecessary guard in refresh history callback

jdufresne:
The function ajaxForm() rejects the fetch promise when the response is
not a 200, so we can assume that if the code is being executed a 200 was
returned.

In the event a 200 is returned, the requests attribute is always a
JavaScript Array.


matthiask:
Makes sense, thanks!

PR #1381, opened 2020-10-11
Prefer document.getElementById() for simply ID lookups

jdufresne:
document.getElementById() is highly optimized by all browsers. There is
no need to introduce selector parsing when the ID is already known.

Further, DOM IDs must be unique. There is no advantage to querying an ID
as a descendant of another element. Using document.getElementId() will
always return a unique element if it exists. By removing the
intermediate ancestor, there are fewer lookups to do.


matthiask:
Makes sense I guess. I don't think `.getElementById("id").querySelectorAll("selector")` is a win vs. `.querySelectorAll("#id selector")` but it's not worse either.

jdufresne:
Sure thing, I have reverted that selector.

PR #1382, opened 2020-10-11
Remove calls to deprecated request.is_ajax()

### LINKED ISSUE #1368: Remove calls to request.is_ajax?
Following up on the discussion in #1316. Should calls to `HttpRequest.is_ajax` be replaced or just removed?

jdufresne:
Since Django 3.1, the method request.is_ajax() is deprecated and will be
removed in a future version. Using Django 3.1 with django-debug-toolbar
results in the warning:

    django-debug-toolbar/debug_toolbar/middleware.py:80: RemovedInDjango40Warning: request.is_ajax() is deprecated. See Django 3.1 release notes for more details about this deprecation.
      request.is_ajax(),

The calls were removed with no replacements as these check are
unnecessary anyway. If the response is not HTML (e.g. JSON) then the
check `or content_type not in _HTML_TYPES` will evaluate as false and
the toolbar will not be included in the response. If the response is an
HTML fragment, it will not contain an `</body>` element (the
INSERT_BEFORE pattern) and so will also not be included in the response.

Fixes #1368

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -69,20 +69,18 @@ class DebugToolbarMiddleware:
 
         response = self.generate_server_timing_header(response, toolbar.enabled_panels)
 
+        # Always render the toolbar for the history panel, even if it is not
+        # included in the response.
+        rendered = toolbar.render_toolbar()
+
         # Check for responses where the toolbar can't be inserted.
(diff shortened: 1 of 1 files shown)


matthiask:
Thank you! The explanation makes a lot of sense to me. Nice that we can simply drop the `is_ajax` calls.

PR #1383, opened 2020-10-11
Add a "Troubleshooting" section to installation docs

### LINKED ISSUE #1352: toolbar.js was blocked because of a dissallowed MIME type ("text/plain")
Hello,

I have followed each step from [this guide](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html).

However the toolbar will not appear.

Inspecting the HTML of a webpage when the django server is running in debug mode confirms that the panels are present; although hidden.

image

Modifying the CSS confirms that the base panel can be displayed as well:

image

Django is returning a HTTP 200 status on the files required:

image

I found that the problem lies with the javascript file was not loading due to it's MIME type:

image

I am able to resolve this by importing the javascript file directly within my own templates; Note that this is without the 'type="module"' attribute:

image

After reloading the webpage, the toolbar works as expected. This is only a temporary fix to the real issue which was intruduced at these commits [#1](https://github.com/jazzband/django-debug-toolbar/commit/54fc973c4fe5dd3289f619d28c285200de96b856#diff-acd782edde03ccd403a2a00b80e0ed64R6) and [#2](https://github.com/jazzband/django-debug-toolbar/commit/54fc973c4fe5dd3289f619d28c285200de96b856#diff-4f4a91c0e3639b7eb74d3dc23799f72dR4).

jdufresne:
This section describes MIME type issues on some platforms when using the
Django runserver command.

Fixes #1352

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -97,3 +97,16 @@ settings module::
 You can change the logic of determining whether or not the Debug Toolbar
 should be shown with the :ref:`SHOW_TOOLBAR_CALLBACK <SHOW_TOOLBAR_CALLBACK>`
 option.  This option allows you to specify a custom function for this purpose.
+
+Troubleshooting
+---------------
+
+On some platforms, the Django ``runserver`` command may use incorrect content
+types for static assets. To guess content types, Django relies on the
+:mod:`mimetypes` module from the Python standard library, which itself relies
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks. This is helpful.

PR #1384, opened 2020-10-11
Regenerate Sphinx documentation configuration for modern usage

jdufresne:
Use a newly generated Sphinx configuration that matches modern Sphinx
practices. This simplifies a lot of the Makefile and conf.py to lean on
defaults instead.


matthiask:
Wow, that's great! I have to check my own projects too :)

PR #1392, opened 2020-10-19
Changed the INTERNAL_IPS CHECK

daveusa31:
You can set "*" in INTERNAL_IPS for access from all APIs

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -18,7 +18,12 @@ def show_toolbar(request):
     """
     Default function to determine whether to show the toolbar on a given page.
     """
-    return settings.DEBUG and request.META.get("REMOTE_ADDR") in settings.INTERNAL_IPS
+    INTERNAL_IPS = settings.INTERNAL_IPS
+    if (settings.DEBUG) and ("*" == INTERNAL_IPS[0] or request.META.get("REMOTE_ADDR") in INTERNAL_IPS):
+        response = True
+    else:
+        response = False
+    return response
 
 
 @lru_cache()


matthiask:
Thank you for your contribution! 

We are reusing Django's `INTERNAL_IPS` setting which doesn't allow `"*"`. 

The behavior you want can already be achieved by overriding `SHOW_TOOLBAR_CALLBACK`. Maybe you could add a paragraph or two to the documentation with examples for overriding the callback?

daveusa31:
I did this:

```python
from debug_toolbar import middleware as dt_middleware

def show_toolbar(_):
    return True
dt_middleware.show_toolbar = show_toolbar
```

matthiask:
Adding this to your settings should do the same thing, without monkey-patching:

```python
DEBUG_TOOLBAR_CONFIG = {
    # DO NOT DEPLOY IN A LIVE ENVIRONMENT
    "SHOW_TOOLBAR_CONFIG": lambda request: True,
}
```

daveusa31:
> Adding this to your settings should do the same thing, without monkey-patching:
> 
> ```python
> DEBUG_TOOLBAR_CONFIG = {
>     # !!!DO NOT DEPLOY IN A LIVE ENVIRONMENT!!!
>     "SHOW_TOOLBAR_CONFIG": lambda request: True,
> }
> ```
Yes, my code looks ugly.
But I tried running yours and it doesn't work

daveusa31:
> Adding this to your settings should do the same thing, without monkey-patching:
> 
> ```python
> DEBUG_TOOLBAR_CONFIG = {
>     # !!!DO NOT DEPLOY IN A LIVE ENVIRONMENT!!!
>     "SHOW_TOOLBAR_CONFIG": lambda request: True,
> }
> ```
It should have been like this

```python
DEBUG_TOOLBAR_CONFIG = {
    "SHOW_TOOLBAR_CALLBACK": lambda request: True,
}
```

matthiask:
Ah yes of course. A stupid typo by me.

PR #1397, opened 2020-11-02
Added missing signals to panel by default.

pope1ni:
I looked through the [signals documentation](https://docs.djangoproject.com/en/stable/ref/signals/) and added some of the missing ones: [`m2m_changed`](https://docs.djangoproject.com/en/stable/ref/signals/#m2m-changed), [`pre_migrate`](https://docs.djangoproject.com/en/stable/ref/signals/#pre-migrate), and [`setting_changed`](https://docs.djangoproject.com/en/stable/ref/signals/#setting-changed).

Note that [`setting_changed`](https://docs.djangoproject.com/en/stable/ref/signals/#setting-changed) comes under the test signals, but says that it can be imported from `django.core.signals`. I left out [`template_rendered`](https://docs.djangoproject.com/en/stable/ref/signals/#template-rendered) as it is "only available during testing".

There are also [`django.contrib.auth.signals`](https://docs.djangoproject.com/en/stable/ref/contrib/auth/#module-django.contrib.auth.signals). I don't think adding them here would make a hard dependency on `django.contrib.auth` so they could be added?

### DIFF
--- a/debug_toolbar/panels/signals.py
+++ b/debug_toolbar/panels/signals.py
@@ -1,15 +1,22 @@
 import weakref
 
-from django.core.signals import got_request_exception, request_finished, request_started
+from django.core.signals import (
+    got_request_exception,
+    request_finished,
+    request_started,
+    setting_changed,
+)
 from django.db.backends.signals import connection_created
 from django.db.models.signals import (
     class_prepared,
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks!

PR #1398, opened 2020-11-02
Add real defaults for configuration

hubdotcom:
as per https://github.com/jazzband/django-debug-toolbar/blob/ab08d14e3bf03a18a2d828d70f18baf98f64efea/debug_toolbar/settings.py


matthiask:
Thanks!

PR #1399, opened 2020-11-04
Document CORS errors and how to fix them

### LINKED ISSUE #1396: CORS required as of v3.0
#1283 introduced an undocumented change: it is not required to set up CORS on the domain that serves static files (given that it's documented, I'd say it's more of a regression):

> Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://static.myapp.localhost:8000/static/debug_toolbar/js/toolbar.js. (Reason: CORS request did not succeed).

I understand that ES6 modules are "the new thing", however, it took me quite a while to figure out why debug toolbar was no longer showing up. This new requirement is not documented anywhere, nor is it in the setup instructions (as such, setting up a new project also fails, unless you serve static files using Django).

Is this change intentional? Is there any way to avoid having to set up CORS on the staticfiles domain? If it's intentional, can it please be documented properly?

jdufresne:
Fixes #1396

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -85,7 +85,14 @@ Change log
   :attr:`Panel.scripts <debug_toolbar.panels.Panel.scripts>` property.
 * Removed support for end of life Django 1.11. The minimum supported Django is
   now 2.2.
+* The Debug Toolbar now loads a `JavaScript module`_. Typical local development
+  using Django ``runserver`` is not impacted. However, if your application
+  server and static files server are at different origins, you may see CORS
+  errors in your browser's development console. See the "Cross-Origin Request
+  Blocked" section of the :doc:`installation docs <installation>` for details
+  on how to resolve this issue.
 
+.. _JavaScript module: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules
 
 2.2 (2020-01-31)
 ----------------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -110,3 +110,40 @@ need to be updated. This can be achieved, for example, by installing or
 updating the ``mailcap`` package on a Red Hat distribution, ``mime-support`` on
 a Debian distribution, or by editing the keys under ``HKEY_CLASSES_ROOT`` in
 the Windows registry.
+
+Cross-Origin Request Blocked
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Debug Toolbar loads a `JavaScript module`_. Typical local development using
+Django ``runserver`` is not impacted. However, if your application server and
+static files server are at different origins, you may see `CORS errors`_ in
+your browser's development console:
+
+.. code-block:: text
+
+    Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://localhost/static/debug_toolbar/js/toolbar.js. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).
+
+Or
+
+.. code-block:: text
+
+    Access to script at 'http://localhost/static/debug_toolbar/js/toolbar.js' from origin 'http://localhost:8000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
+
+To resolve, configure your static files server to add the
+`Access-Control-Allow-Origin header`_ with the origin of the application
+server. For example, if your application server is at ``http://example.com``,
+and your static files are served by NGINX, add:
+
+.. code-block:: nginx
+
+    add_header Access-Control-Allow-Origin http://example.com;
+
+And for Apache:
+
+.. code-block:: apache
+
+    Header add Access-Control-Allow-Origin http://example.com
+
+.. _JavaScript module: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules
+.. _CORS errors: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin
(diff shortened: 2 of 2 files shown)


docs/changes.rst
```diff
@@ -85,7 +85,14 @@ Change log
   :attr:`Panel.scripts <debug_toolbar.panels.Panel.scripts>` property.
 * Removed support for end of life Django 1.11. The minimum supported Django is
   now 2.2.
-
+* JavaScript is now loaded as a `JavaScript module`_. When developing using the
+  Django development server, this will normally load fine and not present an
```
WhyNotHugo, line 89: This is not correct, CORS is required in any scenario where static files are served under a different domain, development or not.

jdufresne: The Django development server is under one origin: `localhost:8000`. What you're describing is the 2nd part of the sentence "if your application server and static files server are at different locations".

Suggestions on wording?

WhyNotHugo: How about:

```rst
JavaScript is now loaded as a `JavaScript module`_. If your application server
and static files server are at different origins (e.g. different host/port),
you may see CORS errors in your browser's development console. See the
"Cross-Origin Request Blocked" section of the :doc:`installation docs
<installation>` for details on how to resolve.
```

There's no need to mention development mode since it's not really relevant; only the origin is relevant.

docs/installation.rst
```diff
@@ -110,3 +110,40 @@ need to be updated. This can be achieved, for example, by installing or
 updating the ``mailcap`` package on a Red Hat distribution, ``mime-support`` on
 a Debian distribution, or by editing the keys under ``HKEY_CLASSES_ROOT`` in
 the Windows registry.
+
+Cross-Origin Request Blocked
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Debug Toolbar loads a `JavaScript module`_. When developing using the
```
WhyNotHugo, line 117: `When serving staticfiles under the same origin as the development server[...]`

docs/changes.rst
jdufresne: There is no mention of "development mode", perhaps this could be clearer. There is only a mention of the development server (aka `runserver`). I want to make clear to users that have a very standard environment -- e.g. the one outline in  the Django tutorial and docs -- that no action is necessary.

WhyNotHugo: > There is no mention of "development mode", perhaps this could be clearer.

It's not relevant at all. Why mention something that's not relevant? It doesn't matter if someone is using `runserver`, `daphne`, or anything else. It doesn't matter if they've followed the django tutorial, or have a production-ready setup.

The only factor that plays any part in this new requirements is if the origins are different.

jdufresne: I think it is worthwhile to reassure users (especially new users) that are using a very typical setup that no additional action is required. Advanced users using an advanced setup are more likely to understand the nuances of different origins and CORS.

tim-schilling: > I think it is worthwhile to reassure users (especially new users) that are using a very typical setup that no additional action is required. Advanced users using an advanced setup are more likely to understand the nuances of different origins and CORS.

Agreed, but I'm not sure the change log is read by new users. I think something closer to what @WhyNotHugo suggested would be better. Maybe we can highlight the happy flow case, then follow-up with the exception?

    The Debug Toolbar loads a `JavaScript module`_.  Typical local 
    machine development is not be impacted. However, if the
    application and static files are served from different hosts...

jdufresne: Thanks for the feedback, all. I've tried to clean this up a little from the provided feedback.

docs/changes.rst
```diff
@@ -85,7 +85,14 @@ Change log
   :attr:`Panel.scripts <debug_toolbar.panels.Panel.scripts>` property.
 * Removed support for end of life Django 1.11. The minimum supported Django is
   now 2.2.
-
+* JavaScript is now loaded as a `JavaScript module`_. Typical local development
+  using Django ``runserver`` is not be impacted. However, if your application
```
matthiask, line 89: will not be impacted or is not impacted?

docs/changes.rst
```diff
@@ -85,7 +85,14 @@ Change log
...
+  server and static files server are at different origins, you may see CORS
+  errors in your browser's development console. See the "Cross-Origin Request
+  Blocked" section of the :doc:`installation docs <installation>` for details
+  on how to resolve.
```
matthiask, line 93: "on how to resolve this issue"? (I'm not a native speaker so I don't know whether this would be an improvement.)

docs/installation.rst
```diff
@@ -110,3 +110,40 @@ need to be updated. This can be achieved, for example, by installing or
 updating the ``mailcap`` package on a Red Hat distribution, ``mime-support`` on
 a Debian distribution, or by editing the keys under ``HKEY_CLASSES_ROOT`` in
 the Windows registry.
+
+Cross-Origin Request Blocked
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Debug Toolbar loads a `JavaScript module`_. Typical local development using
```
matthiask, line 117: The wording is almost the same as above but not quite. Either the JavaScript loads a module or JavaScript is loaded as modules. Both may be correct (nitpicking, sorry)

jdufresne:
Thanks for the feedback @matthiask. I've applied your suggestion so take another look and let me know what you think. Please feel free to nitpick away, I think this is important to get right.

tim-schilling (approved):

auvipy (approved):

WhyNotHugo:
Thanks!

Issue #1402, opened 2020-11-19
Some SQL queries make debug toolbar rendering very slow


alkihis:
Hi, I recently migrated to the last version of django debug toolbar (3.1.1) and I experienced one major performance issue.

When a view makes a "huge" (in textual form) SQL request, rendering of toolbar is very, very slow.
The view takes around 1s to render, and the toolbar around 58-59s on my machine.

The problem is kind of solved by commenting `stack.enable_grouping()`, in function `parse_sql` of `panels/sql/utils.py` (line 34).

[EXAMPLE] This is the kind of request that's causing the problem:
```
SELECT ••• FROM "sample" LEFT OUTER JOIN "sample_parents" ON ("sample"."id" = "sample_parents"."from_sample_id") INNER JOIN "sample_parents" T4 ON ("sample"."id" = T4."to_sample_id") WHERE T4."from_sample_id" IN (856, 858, 860, 862, 900, 1154, 1155, 1156, [...more than 2K numbers...], 29583) GROUP BY ("sample_parents"."from_sample_id"), "sample"."id"
```

I know the problem is linked to `sqlparse` package, but disable grouping (or, make it as an option) doesn't make a big difference in results presentation.

Thank you!

Configuration:
- Python 3.7.9
- Django-debug-toolbar 3.1.1
- PostgreSQL 12
- sqlparse 0.4.1

tim-schilling:
That seems reasonable. Would you mind creating a PR?

As an aside, you can improve your queries performance (and the toolbars) if you can adjust your `Sample` queryset such that the 

```
to_sample_ids = list(Sample.objects.filter(some_filtering=values))
Sample.objects.filter(from_sample__to_sample__id__in=to_sample_ids)
```

To:
```
Sample.objects.filter(from_sample__to_sample__some_filtering=values)
```

PostgreSQL queries that use massive `IN` collections perform poorly. Converting the query to use a subquery that fetches those ids will significantly improve the performance of your query.

alkihis:
Thanks for the quick reply!

I agree with your suggestion ; unfortunately, code that causing this kind of query is slightly more complicated than that, but I'm working on fixing it.

I've made a PR #1404 whose add a new setting in order to control behavior of SQL token grouping. It should be set to `False` when performance issues are encountered.

federicobond:
[...]

PR #1404, opened 2020-11-20
Add SQL_TOKEN_GROUPING setting to control token grouping in SQL panel

alkihis:
As discussed in #1402, this PR expose a new setting, "SQL_TOKEN_GROUPING" which controls token grouping in SQL panel.
It is set as `True` by default in order to preserve current behaviour. 

It should be disabled if a django application makes a very long textual SQL query, which will cause a huge slowdown during toolbar render.


debug_toolbar/settings.py
```diff
@@ -41,6 +41,7 @@
     "SHOW_TEMPLATE_CONTEXT": True,
     "SKIP_TEMPLATE_PREFIXES": ("django/forms/widgets/", "admin/widgets/"),
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
+    "SQL_TOKEN_GROUPING": True,
```
tim-schilling, line 44: To keep in line with the rest of the settings, this should likely be either `ENABLE_SQL_TOKEN_GROUPING` or `SQL_ENABLE_TOKEN_GROUPING`. @matthiask opinions?

docs/configuration.rst
```diff
@@ -221,6 +221,16 @@ Panel options
...
 
+* ``SQL_TOKEN_GROUPING``
+
+  Default: ``True``
+
+  Panel: SQL
+
+  Controls SQL token grouping.
+  When set to ``True``, it might cause render slowdowns
+  when a view make long SQL textual queries.
```
tim-schilling, line 232: This part of the documentation should briefly explain what grouping of SQL tokens means so that the developer understands what this setting does when it's set to `True` besides the possible performance impact.

tim-schilling (changes requested):
This PR should include a test for the functional (not performance) differences when this setting is disabled.

tim-schilling:
@alkihis This is a great start, thank you for your work so far!

debug_toolbar/settings.py
matthiask: The pull request which introduced the code in question was this one: https://github.com/jazzband/django-debug-toolbar/pull/1116

I don't think anyone understands what "token grouping" should be in this context. Maybe something like `PRETTIFY_SQL` would describe the intent better and allow us to disable or enable additional prettification features with one setting. The `NOUN_VERB` format seems to be used with many settings, e.g. "show template context", "enable stacktraces" etc.

matthiask:
Hmm, something went wrong while updating this pull request it seems. Since https://github.com/jazzband/django-debug-toolbar/pull/1404/commits/300d61e4528889657c9f6716f1f08e1c842edc11 the pull request diff contains unrelated changes applied to the master branch in the meantime. 

I'd be grateful if you would clean up this pull request and only include your changes, not everything else. (Don't hesitate to ask for help if you need some support with git for this.)

alkihis:
Replaced by #1438

Issue #1406, opened 2020-11-27
AttributeError: 'dict' object has no attribute 'getlist'


folt:
when using django-debug-toolbar = 3.1 version I get error

```
Traceback (most recent call last):
  File "/home/folt/env_work/temes-Yu7Jpmcx-py3.7/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/folt/env_work/temes-Yu7Jpmcx-py3.7/lib/python3.7/site-packages/sentry_sdk/integrations/django/middleware.py", line 134, in call
    return f(*args, **kwargs)
  File "/home/folt/env_work/temes-Yu7Jpmcx-py3.7/lib/python3.7/site-packages/sentry_sdk/integrations/django/middleware.py", line 90, in sentry_wrapped_method
    return old_method(*args, **kwargs)
  File "/home/folt/env_work/temes-Yu7Jpmcx-py3.7/lib/python3.7/site-packages/debug_toolbar/middleware.py", line 67, in call
    panel.generate_stats(request, response)
  File "/home/folt/env_work/temes-Yu7Jpmcx-py3.7/lib/python3.7/site-packages/debug_toolbar/panels/request.py", line 30, in generate_stats
    "post": [(k, request.POST.getlist(k)) for k in sorted(request.POST)],
  File "/home/folt/env_work/temes-Yu7Jpmcx-py3.7/lib/python3.7/site-packages/debug_toolbar/panels/request.py", line 30, in <listcomp>
    "post": [(k, request.POST.getlist(k)) for k in sorted(request.POST)],
AttributeError: 'dict' object has no attribute 'getlist'
```

when using django-debug-toolbar = 2.0 version everything works correctly

```
[tool.poetry.dependencies]
python = "3.7.8"

django = "^2.2.11"
django-split-settings = "^1.0"
django-feature-policy = "^3.5"

psycopg2 = "^2.8"
gunicorn = "^20.0"
python-decouple = "^3.3"
bcrypt = "^3.2"
structlog = "^20.1.0"
importlib-metadata = "^2.0.0"
django-redis = "^4.12.1"
django-oauth-toolkit = "^1.3.2"
raven = "^6.10.0"
mysqlclient = "^2.0.1"
djongo = "^1.3.3"
djangorestframework = "^3.12.1"
django-filter = "^2.4.0"
django-constance = {extras = ["redis"], version = "^2.7.0"}
[...]

PR #1407, opened 2020-11-27
fix AttributeError: 'dict' object has no attribute 'getlist' #1406

folt:
fix AttributeError: 'dict' object has no attribute 'getlist' #1406


matthiask:
Uh... no, `getlist` isn't the same as `get`. Also, tests are missing for this change. I'm going to revert this.

PR #1409, opened 2020-11-28
Remove support for Python 3.5

matthiask:
The Travis CI build started failing because of a Selenium upgrade. Instead of pinning Selenium to a lower version let's just drop support for Python 3.5 and remove a workaround or two.


jezdez:
Good idea, let's do this.

PR #1410, opened 2020-11-28
Migrate to GitHub Actions.

jezdez:
Travis CI has a new pricing model which places limits on open source.

* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
* https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Many projects are moving to GitHub Actions instead, including Jazzband projects:

* https://github.com/orgs/jazzband/projects/1
* https://github.com/jazzband-roadies/help/issues/206

This is based on https://github.com/jazzband/contextlib2/pull/26.

TODO:

* [x] @jezdez to add `JAZZBAND_RELEASE_KEY` to the repo secrets.


jezdez:
So this is a bit more complicated GHA config since it tests three databases and has extra test environments to do linting, and a single environment to do Selenium testing (which is installde by default).

I'm a bit worried that this takes way too long for each Python version, but I couldn't figure out how to do this better (e.g. using more dimensions like database engine).

If any of the reviewers have an idea how to improve this, I'd appreciate it! Thank you :)

hugovk:
One idea is to factor things back out to the matrix. The earlier Travis config had more or less one job per Python x Django version. Maybe do that here, or for the database engines.

Have a look at this example, you can see an OS x Python matrix:

https://github.com/hugovk/pypistats/blob/7033d0adff90726f439b1b804485fac3fe92c6ec/.github/workflows/test.yml#L10-L12

You can also use `include:` to things not generated by the matrix. For example, something like this should give us 4x3 jobs (4 Python x 3 OS) + 1 extra (3.5 on Ubuntu):

```yml
      matrix:
        python-version: [3.6, 3.7, 3.8, 3.9]
        os: [ubuntu-latest, macos-latest, windows-latest]
        include:
          - { python-version: 3.5, os: ubuntu-latest }
```

(The includes in https://github.com/hugovk/pypistats/blob/7033d0adff90726f439b1b804485fac3fe92c6ec/.github/workflows/test.yml#L13-L18 are a bit different: because the OS matches something in the matrix, it _adds_ a new variable for those.)

Similarly, it's possible to exclude things. This should give us 4x3 - 1 (skip 3.6 on macOS):

```yml
      matrix:
        python-version: [3.6, 3.7, 3.8, 3.9]
        os: [ubuntu-latest, macos-latest, windows-latest]
        exclude:
          - { python-version: 3.6, os: ubuntu-macos-latest }
```

Would something like this help?

jezdez:
@hugovk I've stolen some of techniques for this from @laymonage's [django-jsonfield-backport](https://github.com/laymonage/django-jsonfield-backport) and I think just specifying multiple jobs is the only way to cater to the mix of GHA-level factors and tox-level factors. Or at least I wasn't able to make this happen using tox-gh-actions. I can live with the slight code duplication. WDYT?

hugovk:
I think this is good, especially to get the migration done, and it can always be iterated on later.

hugovk (approved):

jezdez:
Thanks, agreed! 😌

matthiask:
Thanks, this is awesome!

PR #1411, opened 2020-11-29
Add Python3.9 support.

hramezani:


tox.ini
```diff
@@ -3,10 +3,7 @@ envlist =
     docs
     style
     readme
-    py{36,37,38}-dj22-sqlite
-    py{36,37,38}-dj{30,31}-sqlite
-    py{36,37,38}-djmaster-sqlite
-    py{37,38}-dj{22,30,31}-{postgresql,mysql}
+    py{36,37,38,39}-dj{22,30,31,master}-{sqlite,postgresql,mysql}
```
matthiask, line 6: I'd like to ask you to revert this change. Upcoming Django versions may drop support for Python versions and while the previous format is a bit more verbose it's a bit more obvious to me how to manage/remove individual combinations.

matthiask (changes requested):
Looks good to me (apart from the comment above)

tox.ini
hramezani: I reverted this change and add `39` to the end of each list

hramezani:
> Looks good to me (apart from the comment ~above~below)

I address your comment and also add Python and Django supported version badges to `README.rst` in a separate commit.

matthiask:
Thanks!

PR #1418, opened 2020-12-02
Allow overriding CSS and JavaScript

bittner:
If I want to override the CSS and JavaScript shipped with debug toolbar there is no easy way to do this cleanly.

This change adds two template blocks to the base template, which allows extending the template cleanly, patching just CSS and/or JavaScript, as follows:

```django
{# FILE: templates/debug_toolbar/base.html #}

{% extends "debug_toolbar/base.html" %}
{% load static %}

{% block css %}
{{ block.super }}
<link rel="stylesheet" href="{% static 'css/debug-toolbar-custom.css' %}">
{% endblock %}

{% block js %}
{{ block.super }}
<script src="{% static 'js/debug-toolbar-custom.js' %}" async></script>
{% endblock %}
```

I also fixed a minor syntax highlighting issue, but I included that with a separate commit. I hope that's fine.

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/base.html
+++ b/debug_toolbar/templates/debug_toolbar/base.html
@@ -1,7 +1,11 @@
-{% load i18n %}{% load static %}
+{% load i18n static %}
+{% block css %}
 <link rel="stylesheet" href="{% static 'debug_toolbar/css/print.css' %}" media="print">
 <link rel="stylesheet" href="{% static 'debug_toolbar/css/toolbar.css' %}">
+{% endblock %}
+{% block js %}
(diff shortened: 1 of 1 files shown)


matthiask (approved):
I'm fine with this. I asked myself whether you wouldn't want to *add* your own CSS and JS on top instead of replacing the bundled assets (e.g. using `{% block extrahead %}` similar to Django's administration panel) but then I remembered that django-debug-toolbar isn't added to the `<head>` at all.

bittner:
Thanks for reviewing, @matthiask! :100: 

Merging. :rocket:

bittner:
Can we push a new release to PyPI with this soon?

PR #1421, opened 2020-12-08
Adding code to run on Powersystem

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 dist: bionic
+arch:
+    - amd64
+    - ppc64le
 language: python
 cache: pip
 matrix:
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks!

I'm doubtful whether this is helpful:

- CI runs will take longer to finish.
- The debug toolbar doesn't use any native code or other tricks which may impact compatibility with other architectures. 

It seems to me that this is a net negative with doubtful upsides, if any. Maybe I'm missing something?

matthiask:
Also, we aren't using Travis CI anymore. You would have to update the PR for GitHub actions, but let's discuss the merits of this change first.

matthiask:
I'm closing this for now. You can of course still work on it and resubmit against current master (with GitHub actions instead of Travis CI) but please also provide a reasoning for using more CI resources. Thanks!

Issue #423, opened 2013-10-23
Error debugging GeoDjango queries


rodrigoolmo:
When debugging queries that use GeoDjango they are parsed like normal strings.
Example:

```
params = [u"ST_GeomFromEWKB('\\x0101000020e6100000555458045a9d16c010470d5cd47c4440'::bytea)"]
sql = 'SELECT (ST_distance_sphere("location"."point",%s)) AS "distance" FROM "location"'
```

What it should execute:

```
SELECT (ST_distance_sphere("location"."point", ST_GeomFromEWKB('\x0101000020e6100000555458045a9d16c010470d5cd47c4440'::bytea))) AS "distance" FROM "location"
```

What it really executes:

```
SELECT (ST_distance_sphere("location"."point", 'ST_GeomFromEWKB(''\x0101000020e6100000555458045a9d16c010470d5cd47c4440''::bytea)')) AS "distance" FROM "location"
```

I am using django-debug-toolbar version 0.9.4, Django 1.4 and PostgreSQL with PostGIS.

jperelli:
I have the same problem, using `django==2.1.4` and `django-debug-toolbar==1.11`

tim-schilling:
Reopend as per #1547

jperelli:
I used this today and I was very surprised that just worked fine!

I couldn't find it in the changelog but it seems you landed the fix in https://github.com/jazzband/django-debug-toolbar/pull/1426

I think you can close this one @tim-schilling / @aaugustin 

Thank you very much guys!!

tim-schilling:
@jperelli that change was reverted as it mutates data going into your database. You should upgrade versions of the toolbar.

jperelli:
@tim-schilling You are right, sorry abot that. I can see it now in a query that uses a geom as a parameter
![image](https://user-images.githubusercontent.com/1418446/185187520-a2fd3904-ccdb-4431-9447-7a67caeb3e2d.png)
![image](https://user-images.githubusercontent.com/1418446/185187614-2439fd27-93d1-48b9-a6e9-46d6a066eba4.png)
![image](https://user-images.githubusercontent.com/1418446/185187822-c1d1164c-d019-4c31-bffd-ac0494b91e56.png)
:+1:

PR #1426, opened 2020-12-22
Fix SQL selected / SQL explain for gis queries

jieter:
Without this fix, pushing the 'sel' or 'explain' button for a query containing some EWKB-encoded geometry 
as parameter results in this crash:
```
Internal Server Error: /__debug__/sql_explain/
  File "/Users/jieter/.pyenv/versions/obs/lib/python3.8/site-packages/django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.InternalError_: parse error - invalid geometry
LINE 1: ...ure" IN (0, 1, 5)) AND "waarneming"."geo_point" @ 'ST_GeomFr...
                                                             ^
HINT:  "ST" <-- parse error at position 2 within geometry
```

I'm not sure if this is the appropriate location in the code, but with this fix, both `sql_select` and `sql_explain` 
work without flaws.

Previous PR adding a similar fix: #1130
Fixes: #423

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -130,6 +130,14 @@ class NormalCursorWrapper:
     def _record(self, method, sql, params):
         start_time = time()
         try:
+            if isinstance(params, list):
+
+                def strip_GeomFromEWKB(param):
+                    if isinstance(param, str):
+                        return param.lstrip("ST_GeomFromEWKB('\\x").rstrip("'::bytea)")
+                    return param
+
+                params = [strip_GeomFromEWKB(param) for param in params]
             return method(sql, params)
         finally:
             stop_time = time()


matthiask (changes requested):
Thanks! This looks very hacky. I think I understand why this may be necessary but I fear the complexity a bit. Also, this change would definitely need a test to prevent regressions in the future.

jieter:
> Thanks! This looks very hacky. I think I understand why this may be necessary but I fear the complexity a bit. Also, this change would definitely need a test to prevent regressions in the future.

Yes, fully agreed! I only tried to make the existing solution a bit more readable by avoiding the numerical offsets.

I just looked into adding tests, which would require another test matrix entry with `DB_BACKEND=postgis`. Do you mind if I add that @matthiask?

matthiask:
@jieter Regarding `DB_BACKEND=postgis`: Not at all, I think that would be great.

jieter:
I've added postgis to the tests (not yet added to github actions).

I also added a test, but this still crashes with the same error, and (for as far as I can see) the same input as captured while running debug-toolbar with the fix in my development environment. I'll try to revisit this later.

PR #1427, opened 2020-12-22
Really ugly work in progress hack to fix tests on Django@master

matthiask:
Django 3.2 has changed a few internals in the `django.core.cache` module. Our instrumentation doesn't pick up cache calls anymore. Here's a hack which has worked locally. Since django-debug-toolbar relies on monkey patching anyway, something like this may work.

(Test failures on released Django versions are to be expected.)

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -130,13 +130,20 @@ class CacheStatTracker(BaseCache):
         return self.cache.decr_version(*args, **kwargs)
 
 
-class CacheHandlerPatch(CacheHandler):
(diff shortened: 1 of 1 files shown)


tim-schilling:
I think in order to commit this, we'd need to figure out a way to have the cache tracking disabled.

tim-schilling:
This will enable the cache panel for the middleware by default. I don't think it's possible to avoid that and still capture the middleware cache calls before the toolbar's middleware.

tim-schilling:
With #1437 merged, this can be closed.

PR #1429, opened 2020-12-25
Update installation.rst

AnjaneyuluBatta505:
added comment in urls section to make it easy to copy and paste to `urls.py`


matthiask:
If you do want to make it copy pasteable you should probably write `urlpatterns += [...]`

PR #1431, opened 2021-01-02
Default the make test command to use sqlite database.

tim-schilling:
With the change to use github actions, we need to supply env variables to
tox and our tests. It takes a bit to find that information as a new dev,
so let's default to sqlite to make running the tests easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,11 +27,13 @@ package-lock.json: package.json
 	touch $@
 
 test:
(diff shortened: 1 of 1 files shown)


matthiask:
@tim-schilling Is there a reason why you closed this (except for the inactivity?) I like it because I never know how to run tests locally :-)

tim-schilling:
I think it'd be better to have the ability to allow a user to run the tests with any of the databases. I'll open another that will replace this.

Issue #1430, opened 2020-12-28
gather information from async functions


smmoosavi:
If some queries are executed in async mode, these queries are missing in the panel. How can we collect these?

```python
data = asyncio.run(parallel_calc_aggregations(queryset, aggregations))
```

**note1:** I am using `django-debug-toolbar` via [`django-graphiql-debug-toolbar`][django-graphiql-debug-toolbar]
**note2:** all queries are inside function wrapped by [`sync_to_async`][sync_to_async]

The same issue exists when graphql `ThreadExecutor` is used. read more about graphql executors [graphql-314]

[django-graphiql-debug-toolbar]: https://github.com/flavors/django-graphiql-debug-toolbar
[sync_to_async]: https://docs.djangoproject.com/en/3.1/topics/async/#asgiref.sync.sync_to_async
[graphql-314]: https://github.com/graphql-python/graphene/issues/314#issuecomment-251585796

tim-schilling:
The toolbar currently does not support async and the collection of statistics is tightly coupled to a request. If you do end up diving into how to modify the toolbar to collect this information, we'd appreciate any conclusions or findings of yours.

smmoosavi:
Is there any decorator function we can add around these functions and collect data required by debug-toolbar?

```python
@collect_toolbar_inf
def get_data():
  # pure django

async def async_get_data():
      return await sync_to_async(get_data, thread_sensitive=False)()

async def parallel_get_data():
  return await asyncio.gather(async_get_data(), async_get_data(), async_get_data())

data = asyncio.run(parallel_get_data())
```

tim-schilling:
Currently the only API for collecting the data is done via the middleware and the methods on the panels themselves.

smmoosavi:
maybe it's possible to wrap [`sync_to_async`][sync_to_async]. all async function using graphql orm should use this function and without `sync_to_async` function causing error.

[sync_to_async]: https://github.com/django/asgiref/blob/5b86d9684e1f3c28956fb0dfe8d4f4a427be8f87/asgiref/sync.py#L372

tim-schilling:
@michaelurban that's a good idea. Or at least log some type of message. Would you be able to write up a PR to introduce that feature?

Issue #1413, opened 2020-11-30
Request History do not work


KStenK:
Hello,

I remember in the past at one point I got the request history panel working, but now if I need to use it, it won't work.

On the front end I have enabled:
History, Versions, Time, and SQL panels.

Now if I click on the History tab, then there no rows.
And if i click on the Refresh button,  i get 400 error:
![image](https://user-images.githubusercontent.com/53740661/100651754-9e3a0400-334e-11eb-90b3-33553c19d458.png)

I have no idea what I am doing wrong or how to troubleshoot that problem.
Any suggestions, what should I try next?

Python 3.9.0

Dependencies:
Django==3.1.3
django_debug_toolbar==3.1.1

folt:
Can you share your app settings?

KStenK:
> Can you share your app settings?

Updated the issue with some configuration.
Is there anything specific you would like to know?

folt:
[There is a support request for python 3.9](https://github.com/jazzband/django-debug-toolbar/pull/1411). 
Is there an error remaining if I downgrade it?

matthiask:
I'm using Python 3.9 as well, I don't think Python is the problem here. I have no idea what the problem might be :/

KStenK:
In the 400 response error, there is 'Form errors'.
![image](https://user-images.githubusercontent.com/53740661/100740807-f79f4300-33e1-11eb-9e4c-ce42baabd159.png)

folt:
Do you have applications installed in your browser that can interfere with the execution of javascript on the page?
There is also a possibility that the page may contain your scripts from the template that can break the panel.
I do not see similar problems with the given configuration.

KStenK:
> Do you have applications installed in your browser that can interfere with the execution of javascript on the page?
> There is also a possibility that the page may contain your scripts from the template that can break the panel.
> I do not see similar problems with the given configuration.

I have tried with different browsers and even on the empty HTML page that occurs.
That is in Fetch init:
![image](https://user-images.githubusercontent.com/53740661/100755393-cd0ab580-33f4-11eb-88ba-0a23b11a5390.png)

folt:
![Screenshot_20201202_123949](https://user-images.githubusercontent.com/3757488/100855931-ce42ed80-349b-11eb-888a-b67313e9b634.png)

I came across just such a message.
But the problems with the refresh could not be repeated.

KStenK:
Okei, I found the source of the problem.
[...]

PR #1432, opened 2021-01-03
Support serializable toolbar

tim-schilling:
In order to support Channels' non-django request possibilities, we need to start making the toolbar more flexible.In order to support Django Channels out of the box we need to support multiple processes running. This implies serializing the toolbar's data into some type of backend (database, memory, redis, etc). Currently the toolbar stores everything in memory as non-primitives.

My assumption is that we need to serialize the panel's template context into primitives so that it can be stored in any medium between requests and then rendered at some future time (or even stored elsewhere). I could be wrong here, so feel free to push back. I suspect this is going to change the toolbar's internals dramatically.

This should not be merged as it's not complete, but I wanted to get it out there for feedback and to solicit for ideas/alternative options. Currently this PR does the following:

- Creates a global store for the toolbar and the panels' stats (`debug_toolbar.store.MemoryStore`)
- Reworks all the existing panels to function with their context being serialized to JSON and back (this should be pulled out for the `MemoryStore`, but I'm doing it to prove that it does work).

Future work:

- Create ~~`LocMemCacheStore`~~ `DBCacheStore`
- Verify compatibility with Django Channels sync Django views
- Determine what's needed to support async Django Channels views.

This relates to #1430 and I hope we can structure it in a way such that it also works for async views in Channels and #1413.


tim-schilling:
@tapaswenipathak tagging you here. Feel free to create a new PR or continue working off of this one.

debug_toolbar/panels/__init__.py
```diff
@@ -150,19 +157,42 @@ def disable_instrumentation(self):
...
+        return data
+
+    @classmethod
+    def serialize_stats(cls, stats):
+        """
+        Serialize stats for the store.
+
+        Provided to support future store mechanisms overriding a panel's content.
+        """
+        return stats
```
tapaswenipathak, line 176: are these two methods complete?

1. `deserialize_stats`
2. `serialize_stats`

debug_toolbar/store.py
```diff
@@ -0,0 +1,90 @@
...
+            return str(o)
+
+
+def serialize(data):
+    return json.dumps(data, cls=DebugToolbarJSONEncoder)
+
+
+def deserialize(data):
+    return json.loads(data)
+
```
tapaswenipathak, line 24: are these two methods complete?

1. `serialize()`
2. `deserialize()`

debug_toolbar/store.py
```diff
@@ -0,0 +1,90 @@
...
+    def exists(cls, store_id):
+        raise NotImplementedError
+
+    @classmethod
+    def set(cls, store_id):
+        raise NotImplementedError
+
+    @classmethod
+    def delete(cls, store_id):
+        raise NotImplementedError
```
tapaswenipathak, line 47: should these four be implemented?

1. `ids()`
2. `exists()`
3. `set()`
4. `delete()`

debug_toolbar/store.py
```diff
@@ -0,0 +1,90 @@
...
+        cls._stats[store_id][panel_id] = serialize(stats)
+
+    @classmethod
+    def panel(cls, store_id, panel_id):
+        try:
+            data = cls._stats[store_id][panel_id]
+        except KeyError:
+            data = None
+        return {} if data is None else deserialize(data)
+
```
tapaswenipathak, line 87: are these two method definitions complete?

1. `save_panel()`
2. `panel()`

debug_toolbar/templates/debug_toolbar/panels/history_tr.html
```diff
@@ -43,7 +43,7 @@
     </td>
     <td class="djdt-actions">
         <form method="get" action="{% url 'djdt:history_sidebar' %}">
-            {{ store_context.form }}
+            {{ history.form }}
```
tapaswenipathak, line 46: `store_context` is functionally `history` after this PR?

debug_toolbar/toolbar.py
```diff
@@ -88,22 +88,8 @@ def should_render_panels(self):
...
-            return
-        self.store_id = uuid.uuid4().hex
-        self._store[self.store_id] = self
-        for _ in range(self.config["RESULTS_CACHE_SIZE"], len(self._store)):
-            self._store.popitem(last=False)
-
-    @classmethod
-    def fetch(cls, store_id):
-        return cls._store.get(store_id)
+        get_store().set(self.store_id)
```
tapaswenipathak, line 92: now i would:

`return import_string(dt_settings.get_config()["TOOLBAR_STORE_CLASS"])`

are these edits/changes functional and complete?

debug_toolbar/store.py
tim-schilling: Not here. BaseStore is the interface for the API. The subclasses of BaseStore will have different implementations.

debug_toolbar/store.py
tim-schilling: Yes.

debug_toolbar/toolbar.py
tim-schilling: The function `store` here is more the verb case of "store". What's happening is that we want to store this version of the toolbar into the storage mechanism. The naming could be improved.

tim-schilling:
This is going to be replaced by the work associated with https://github.com/orgs/jazzband/projects/9/

@matthiask can you review those high level tasks and let me know what context I'm missing? When these parts are picked up, they should be converted to issues. Though perhaps I should do that now?

matthiask:
@tim-schilling Thanks for starting this! 

I think we should check if we can use the cache framework instead of writing our own store classes; we should document that you're supposed to use a cache which supports cross-process caching and maybe warn if users use e.g. the local memory or dummy cache backend.

I'm sure that the point could be made that this data doesn't belong in a cache but it would mean much less work for us and many many options immediately available for our users.

I also think Django Channels is nice to have, not essential -- Django itself comes with async views, middleware and ORM methods. I have used Channels in the past with success and I like it, but supporting Django's own async machinery is surely more important.

Re. the serialization framework: Do you think it would be feasible to enforce usage of these methods so that we do not have to support two code paths in the future? I think it would be great if we also added a pending deprecation warning to using current-style panels which do not use serialization at all.

(I haven't thought long and hard about these points so I'm not all that sure that we should do what I'm proposing.)

tim-schilling:
> I think we should check if we can use the cache framework instead of writing our own store classes; we should document that you're supposed to use a cache which supports cross-process caching and maybe warn if users use e.g. the local memory or dummy cache backend

@matthiask Good point. I'm not sure why I went into specific backends. In reality, it should have been MemoryStore, CacheStore and DatabaseStore. However, I'm happy to start with using the cache, then extending if folks want to.

> I also think Django Channels is nice to have, not essential

Another good point. But let's reach for it still. Especially if we get help from GSoC.

> Re. the serialization framework: Do you think it would be feasible to enforce usage of these methods so that we do not have to support two code paths in the future? I think it would be great if we also added a pending deprecation warning to using current-style panels which do not use serialization at all. 

Yes, I think we'll drop the double code paths in the future. I don't have a timeline for you, but if you feel we should define one, I'll say 1 year from release we'll drop the deprecated path.

matthiask:
> Another good point. But let's reach for it still. Especially if we get help from GSoC.

I totally agree.

> Yes, I think we'll drop the double code paths in the future. I don't have a timeline for you, but if you feel we should define one, I'll say 1 year from release we'll drop the deprecated path.

I don't think we need a timeline for this, but we should be intentional about it, and communicate clearly that we don't expect to keep both paths indefinitely.

Thanks!

tim-schilling:
I'm closing this as this approach is largely being dropped. It can still be referred to for inspiration, but it won't be merged in.

PR #1436, opened 2021-01-20
Show template context on included templates

### LINKED ISSUE #902: SQL panel points a query in an included template to the same line of its parent template

gilmrjc:
This PR makes possible to see the template context of a nested template on the SQL panel.

Fix #902


auvipy (approved):

auvipy (changes requested):
can you please check the sqlite failures?

matthiask:
The test failures are probably related to https://github.com/jazzband/django-debug-toolbar/pull/1427 

They are not related to sqlite3 – the problems appear in the sqlite3 build only because sqlite3 is the only build which runs tests using Django@master.

tests/templates/sql/nested.html
```diff
@@ -0,0 +1,4 @@
+{% extends "base.html" %}
+{% block content %}
+  {% include "sql/included.html" %}
```
tim-schilling, line 3: How come you went with an `include` here rather than using `{{ users }}`?

gilmrjc: From the issue #902 

> The SQL query of Question.objects.all() is executed at the line 5 of inner.html, but in the SQL panel, it is incorrectly pointed to the line 5 of its parent template index.html

The problem is that the SQL panel does not show the template context when using included templates. This file to test the correct behavior.

tim-schilling: I'm sorry, I didn't realize we were in the tests directory.

tim-schilling (approved):
Looks good. I ran this locally with some changes to the example app.

gilmrjc:
I rebased the branch to include the new changes.

matthiask:
Thanks!

PR #1437, opened 2021-01-20
Patch default cache when using the cache panel

gilmrjc:
The test for the cache panel are broken on master (ea65261). The problem is caused for a change in how Django handles the default cache connection on version 3.2. This PR handles the new connection system to avoid this problem.

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -3,9 +3,20 @@ import sys
 import time
 from collections import OrderedDict
 
+try:
+    from django.utils.connection import ConnectionProxy
+except ImportError:
+    # Django 3.1 and below doesn't have a ConnectionProxy
+    pass
+
 from django.conf import settings
 from django.core import cache
-from django.core.cache import CacheHandler, caches as original_caches
+from django.core.cache import (
+    DEFAULT_CACHE_ALIAS,
+    CacheHandler,
+    cache as original_cache,
+    caches as original_caches,
+)
 from django.core.cache.backends.base import BaseCache
 from django.dispatch import Signal
 from django.middleware import cache as middleware_cache
@@ -246,8 +257,14 @@ class CachePanel(Panel):
         else:
             cache.caches = CacheHandlerPatch()
 
+        try:
+            cache.cache = ConnectionProxy(cache.caches, DEFAULT_CACHE_ALIAS)
+        except NameError:
+            pass
+
     def disable_instrumentation(self):
         cache.caches = original_caches
+        cache.cache = original_cache
         # While it can be restored to the original, any views that were
         # wrapped with the cache_page decorator will continue to use a
         # monkey patched cache.


gilmrjc:
There is a problem with Codecov uploads from the sqlite branch, which is the one testing the integration with the Django master branch. Taken that into account there is no coverage loss with the proposed changes.

auvipy (approved):

tim-schilling (changes requested):
It doesn't appear to be necessary to import `ConnectionProxy`. I was able to get the tests to pass with:

    cache.cache = cache.caches[DEFAULT_CACHE_ALIAS]

Thanks for pointing out the simpler solution than I was working on.

gilmrjc:
> It doesn't appear to be necessary to import `ConnectionProxy`. I was able to get the tests to pass with:
> 
> ```
> cache.cache = cache.caches[DEFAULT_CACHE_ALIAS]
> ```
> 
> Thanks for pointing out the simpler solution than I was working on.

I used the `ConnectionProxy` because that's the new class used in the Django codebase. It's possible to only change the default cache without creating a new instance of `ConnectionProxy` but maybe it can cause problems somewhere else. I haven't tested a large application with this so I'm not sure.

tim-schilling:
Ah, that's why I was patching the existing connections in #1427

gilmrjc:
The difference with #1427 is the patched class there is `CacheHandler` but now the default cache is initialized with the new `ConnectionProxy` class.

In Django 3.1 and below the default proxy created the connection on access time with `CacheHandler`, including the default cache connection. If the patch was applied before any call the connection would be instrumented.

With the new logic (https://github.com/django/django/blob/7b3ec6bcc8309d5b2003d355fe6f78af89cfeb52/django/core/cache/__init__.py#L49) the default cache is instantiated on load time and when the patch to `CacheHandler` is applied the connection is already encapsulated in a `ConnectionProxy` object.

This is why the old patch can't instrument the default cache but it works when accessing `cache.caches["default"]` or other caches, because `cache.caches["default"]` and `cache.cache` are two different objects, where in Django 3.1 and below they were the same.

tim-schilling:
Sounds good, but can we work the inline import out of the code? One option would be to create our own version of `ConnectionProxy` in a `compat.py` file, then import the compat version when the Django version is less than 3.1.

gilmrjc:
Can I ask why you want to avoid the import? It's actually used to detect if the new patch is needed:

```python
try:
    from django.utils.connection import ConnectionProxy

    cache.cache = ConnectionProxy(cache.caches, DEFAULT_CACHE_ALIAS)
except ImportError:
    pass
```

If `ConnectionProxy` is not found it means we are on Django 3.1 or below and nothing happens. If the class is imported then the patch to `cache.cache` is needed. This is a idiomatic way to detect  API changes in Python when dealing with multiple versions of a dependency.

I'm open to change it to `compat.py`, just want to understand the logic behind the suggestion.

tim-schilling:
The reason for the suggestion is to get the import statement to the top of the file rather than having it inline. It's not about the actual try/except flow.

gilmrjc:
Ok, the import is now at the top. It is a minor change and it's still just patching when needed 😄

debug_toolbar/panels/cache.py
```diff
@@ -3,9 +3,20 @@
 import time
 from collections import OrderedDict
 
+try:
+    from django.utils.connection import ConnectionProxy
+except ImportError:
+    # Django 3.1 and below doesn't have a ConnectionProxy
+    pass
```
matthiask, line 10:
```suggestion
    ConnectionProxy = None
```

debug_toolbar/panels/cache.py
```diff
@@ -246,8 +257,14 @@ def enable_instrumentation(self):
         else:
             cache.caches = CacheHandlerPatch()
 
+        try:
+            cache.cache = ConnectionProxy(cache.caches, DEFAULT_CACHE_ALIAS)
+        except NameError:
+            pass
```
matthiask, line 263:
```suggestion
        # Wrap the patched cache inside Django's ConnectionProxy
        if ConnectionProxy:
            cache.cache = ConnectionProxy(cache.caches, DEFAULT_CACHE_ALIAS)
```

matthiask (approved):
I think we should move ahead with this. I have suggested two small-ish changes so that we do not have to use the broad `try ... except NameError` clause.

tim-schilling (approved):

PR #1438, opened 2021-01-21
Add PRETTIFY_SQL setting to control token grouping in SQL panel

alkihis:
As discussed in #1402, this PR expose a new setting, "PRETTIFY_SQL" which controls token grouping in SQL panel.
It is set as True by default in order to preserve current behaviour.

It should be disabled if a django application makes a very long textual SQL query, which will cause a huge slowdown during toolbar render.

This is a new PR (original #1404) to fix some commit issues as reported in [this](https://github.com/jazzband/django-debug-toolbar/pull/1404#issuecomment-738878417) comment.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -180,6 +180,39 @@ Panel options
   Useful for eliminating server-related entries which can result
   in enormous DOM structures and toolbar rendering delays.
 
+* ``PRETTIFY_SQL``
+
+  Default: ``True``
+
+  Panel: SQL
+
+  Controls SQL token grouping.
+
+  Token grouping allows pretty print of similar tokens,
+  like aligned indentation for every selected field.
+
+  When set to ``True``, it might cause render slowdowns
+  when a view make long SQL textual queries.
+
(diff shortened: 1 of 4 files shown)


docs/configuration.rst
```diff
@@ -221,6 +221,40 @@ Panel options
   The SQL panel highlights queries that took more that this amount of time,
   in milliseconds, to execute.
 
+* ``PRETTIFY_SQL``
```
tim-schilling, line 224: Can you move this above `PROFILER_MAX_DEPTH` to make the settings alphabetical please?

tim-schilling (changes requested):
Well done! I appreciate the example in the documentation.

docs/configuration.rst
alkihis: Done in [cbf8f84](https://github.com/jazzband/django-debug-toolbar/pull/1438/commits/cbf8f84c06264060033780a70e1155e2503c88c5).

tim-schilling (approved):
Thank you for the PR!

PR #1440, opened 2021-01-22
Correct RENDER_PANELS functionality and when enabled disable HistoryPanel

tim-schilling:
The render panels setting used to control whether the toolbar would keep
the contents of the panels in the background or render the toolbar on
every page. The history panel relies on loading panels in the background.
If panels should be rendered on the request, then the history panel is
can't/shouldn't work.

References #1413


docs/changes.rst
```diff
@@ -9,6 +9,10 @@ Next version
 * Added ``PRETTIFY_SQL`` configuration option to support controlling
   SQL token grouping. By default it's set to True. When set to False,
   a performance improvement can be seen by the SQL panel.
+* ``HistoryPanel`` will be disabled when ``RENDER_PANELS`` is ``True``
+  or if it's not set, but the server is running with multiple processes.
+* Fixes ``RENDER_PANELS`` functionality so that when ``True`` panels are
+  rendered during the request and not loaded asynchronously.
```
matthiask, line 15: I have a hard time understanding the first bullet point. The second bullet point should probably use past tense too for consistency.

tim-schilling: Yeah, that's written poorly. I'll fix it.

matthiask:
Thanks!

PR #1441, opened 2021-01-22
Support JS events when loading a panel.

### LINKED ISSUE #1439: Time panel doesn't re-run timer.js for new requests
The Time panel will load the `timer.js` file and run it on the first time accessing the Time panel. After that, when the user switches to a new request via the History panel, any subsequent usages of the Time panel will result in the "Browser timing" section not being shown due to the `timer.js` module not being re-run after loading the panel.

tim-schilling:
This fixes the problem of the Timer Panel not inserting the browser
timings section after being loaded via the HistoryPanel.

These events could be wired into to better render panels or support
a more dynamic toolbar and/or panel.

Fixes #1439

### DIFF
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -107,6 +107,10 @@ class Panel:
     def scripts(self):
         """
         Scripts used by the HTML content of the panel when it's displayed.
+
+        When a panel is loaded on the frontend, a JavaScript event will be
+        dispatched of the format ``djdt.panel.[panel_id]``. The scripts used can
+        listen for this event to execute functionality on panel load.
         """
         return []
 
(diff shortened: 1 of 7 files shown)


matthiask:
I think I like it. 

@jdufresne I requested a review because you added the new panel loading code; please feel free to skip this if you don't find the time or energy for a review!

tim-schilling:
I was thinking more about this and I'm going to make a single event with a
property that contains the panel id. It makes more sense than the event
being the panel id. I can create a public function that does the boiler
plater listener logic that does the basic filtering. It seems cleaner in my
mind.

tim-schilling:
I changed the approach to put the responsibility of handling the async nature of a panel's scripts on the panel's script rather than `toolbar.js`. I also changed the functionality to operate on rendering rather than ajax loading. I figured this would be more flexible in the future.

These commits should be squashed before merging (if approved).

matthiask:
This looks good I think. @tim-schilling Do you think it's ready to merge or is there anything left to do here?

auvipy (approved):

tim-schilling:
Thanks for the reviews folks. Squashed and merged in.

PR #1442, opened 2021-01-23
show namespaces alongside url_name in request panel

AlexCLeduc:
As far as I know, no one really asked for this, but I work on projects that make heavy use of namespaces with repetitive urls (e.g. create, edit, index) and having namespaces available in the request panel helps disambiguate and makes for a quicker grep. 

Before this PR:
Screen Shot 2021-01-23 at 11 08 42 AM

After this PR:
Screen Shot 2021-01-23 at 11 09 20 AM

I used `:` as a separator because that's what you'd pass to e.g. `reverse("admin:login")`. We can alternatively add a separate 'namespace' title/content section, but this seemed cleaner to me

### DIFF
--- a/debug_toolbar/panels/request.py
+++ b/debug_toolbar/panels/request.py
@@ -44,7 +44,16 @@ class RequestPanel(Panel):
             view_info["view_func"] = get_name_from_obj(func)
             view_info["view_args"] = args
             view_info["view_kwargs"] = kwargs
-            view_info["view_urlname"] = getattr(match, "url_name", _("<unavailable>"))
+
+            if getattr(match, "url_name", False):
+                url_name = match.url_name
+                if match.namespaces:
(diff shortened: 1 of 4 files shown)


matthiask:
Thanks, this is great. I think including namespaces directly in the URL name is fine. 

However, the new test seems to fail quite often. @AlexCLeduc  Do you have an idea why?

AlexCLeduc:
@matthiask  Sorry about that, I had trouble running tests locally and figured CI was misconfigured. Turns out the test app never included the admin URLs, looks like it works now.

tim-schilling:
@AlexCLeduc If possible, could you move this test to the `RequestPanelTestCase` test case in `tests/panels/test_request.py`? If we don't need it in the integration test case, I'd like to avoid that.

tim-schilling (approved):

matthiask:
Thank you! 💪🎉

PR #1445, opened 2021-02-02
Remove unsave inline scripts and styles

### LINKED ISSUE #1248: Regression: Broken inline styles with CSP
The `applyStyle()` function that was removed as "unnecessary" in #1220 had actueally been added in #704 to fix issues with CSP. I suggest to revert #1220.

xi:
Fixes #1248 

~~As a first step I pushed a test. It works locally but I was not sure whether it would work with CI, so I pushed it. I hope that is OK.~~ For some reason CI does not run so I have no clue whether the test I added actually works.

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/panels/profiling.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/profiling.html
@@ -14,7 +14,7 @@
     {% for call in func_list %}
       <tr class="{% for parent_id in call.parent_ids %} djToggleDetails_{{ parent_id }}{% endfor %}" id="profilingMain_{{ call.id }}">
         <td>
-          <div data-padding-left="{{ call.indent }}px">
+          <div data-djdt-styles="paddingLeft:{{ call.indent }}px">
             {% if call.has_subfuncs %}
               <button type="button" class="djProfileToggleDetails djToggleSwitch" data-toggle-name="profilingMain" data-toggle-id="{{ call.id }}">-</button>
             {% else %}
--- a/debug_toolbar/templates/debug_toolbar/panels/profiling.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/profiling.html
@@ -14,7 +14,7 @@
     {% for call in func_list %}
       <tr class="{% for parent_id in call.parent_ids %} djToggleDetails_{{ parent_id }}{% endfor %}" id="profilingMain_{{ call.id }}">
         <td>
-          <div style="padding-left:{{ call.indent }}px">
+          <div data-padding-left="{{ call.indent }}px">
             {% if call.has_subfuncs %}
               <button type="button" class="djProfileToggleDetails djToggleSwitch" data-toggle-name="profilingMain" data-toggle-id="{{ call.id }}">-</button>
             {% else %}
--- a/debug_toolbar/templates/debug_toolbar/panels/sql.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/sql.html
@@ -2,7 +2,7 @@
 <ul>
   {% for alias, info in databases %}
     <li>
-      <strong><span class="djdt-color" style="background-color:rgb({{ info.rgb_color|join:', ' }})"></span> {{ alias }}</strong>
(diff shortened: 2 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -220,3 +220,6 @@ jobs:
 
     - name: Test with tox
       run: tox -e docs,style,readme
```
tim-schilling, line 222: I think it's more appropriate for this check of inline styles or scripts to exist in the `Makefile`, then including it here with the other static checks run by tox.

Doing so will also make it easier for developers to run this check before pushing to github and having CI catch it.

debug_toolbar/templates/debug_toolbar/panels/sql.html
```diff
@@ -122,3 +122,5 @@
 {% else %}
   <p>{% trans "No SQL queries were recorded during this request." %}</p>
 {% endif %}
+
+<script src="{% static 'debug_toolbar/js/toolbar.sql.js' %}" async></script>
```
tim-schilling, line 126: This should be included via the `SQLPanel.scripts` property.

debug_toolbar/templates/debug_toolbar/panels/profiling.html
```diff
@@ -32,3 +32,5 @@
     {% endfor %}
   </tbody>
 </table>
+
+<script src="{% static 'debug_toolbar/js/toolbar.profiling.js' %}" async></script>
```
tim-schilling, line 36: This should be included via the `ProfilingPanel.scripts` property.

debug_toolbar/static/debug_toolbar/js/toolbar.profiling.js
```diff
@@ -0,0 +1,3 @@
+(function () {
+    djdt.applyStyle('padding-left');
+})();
```
tim-schilling, line 3: I think I'd be more in favor of an approach that would find all elements that have a class such as `djdt-dynamic-style`, then have it apply the styles listed in the data attribute. That way the javascript doesn't need to know anything about the specific styles that the HTML elements are having applied.

tim-schilling (commented):
Thank you for the PR. My comments vary between discussions and change requests.

- I'd like to see the scripts moved to the panel classes.
- I also feel pretty strongly about moving the check to the `Makefile`.
- The comment regarding the `applyStyle` approach is more of a discussion.

xi:
Thanks for the quick feedback. I had tried to simply revert 87364f7049d16073a7ba8f2ad95e23867b9e622c which I now realize was a bit naive.

I would like to avoid making `applyStyles()` more generic. It is just meant to be compatibility code until a better solution is found. New code should ideally not use it IMHO.

tim-schilling:
@xi I squashed your commits and converted the `applyStyles` function to something I'd prefer to maintain. I understand your concern regarding "It is just meant to be compatibility code until a better solution is found.", but I don't think there's a great solution out there. I find this approach more flexible than having to call `applyStyles` per dynamic style desired.

matthiask:
I'm afraid I don't understand how this works. Does `element.style["background-color"]` really set the background color of an element? Shouldn't this be `element.style["backgroundColor"]`? I did a quick local test and it seems to work but I'm surprised that it does. 

(`applyStyles` is ugly, yes, but I don't have a better idea.)

tim-schilling:
@matthiask There's a brief mention of using the dashed names here in the docs of [CSSStyleDeclaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration#css_properties). However, you're right, it's probably best to avoid using the dashed names.

tim-schilling:
I updated the applyStyles function to mention the expected format and switched `background-color` to `backgroundColor`.

xi:
For completeness' sake:

> Setting the dashed attribute attribute must invoke setProperty() with the first argument being dashed attribute, as second argument the given value, and no third argument. Any exceptions thrown must be re-thrown.
https://www.w3.org/TR/2016/WD-cssom-1-20160317/#the-cssstyledeclaration-interface

So apperantly `element.style["background-color"]` is fine. That spec is a draft though, so no clue how widely this is supported.

tim-schilling:
I missed `padding-left`. This should be good to go now.

matthiask:
Thank you! 🎉

PR #1447, opened 2021-02-02
Remove pkg_resources usage

### LINKED ISSUE #1446: Slower than necessary start up (import) times
Hi,
I am digging into slow start up times of a Django app at the moment. I used `` to see which imports are slow and the `debug_toolbar` package is among them. But it would not have to be slow if we could get rid of the `pkg_resources` import.

```
$ PYTHONPROFILEIMPORTTIME=1 python -c "import debug_toolbar" 2>&1 | tail -n2
import time:    117915 |     179477 |   pkg_resources
import time:       624 |     180101 | debug_toolbar
```
So, on my machine the import took ~180ms of which ~179ms can be attributed to `pkg_resources`.

Here is an article that details the problem with `pkg_resources` and put me onto this path :-)
https://dev.to/methane/how-to-speed-up-python-application-startup-time-nkf

I am preparing a PR, more details over there...

moser:
Fixes #1446 

I replaced the usage of `pkg_resources` by a hard-coded version number to save ~30% on the import time (and thus start-up time of any django app, script, test run etc that uses debug_toolbar).

Before:
```
$ time python -c "import debug_toolbar; import debug_toolbar.apps" 2>&1
python -c "import debug_toolbar; import debug_toolbar.apps" 2>&1  0,28s user 0,03s system 99% cpu 0,316 total
```

After:
```
$ time python -c "import debug_toolbar; import debug_toolbar.apps" 2>&1
python -c "import debug_toolbar; import debug_toolbar.apps" 2>&1  0,18s user 0,03s system 100% cpu 0,207 total
```

Do you think this is worth the additional hassle for the maintainer/release person to update the version number in one more place? :-)

### DIFF
--- a/debug_toolbar/__init__.py
+++ b/debug_toolbar/__init__.py
@@ -1,13 +1,9 @@
 __all__ = ["VERSION"]
 
 
-try:
-    import pkg_resources
-
-    VERSION = pkg_resources.get_distribution("django-debug-toolbar").version
-except Exception:
-    VERSION = "unknown"
-
+# Do not use pkg_resources to find the version but set it here directly!
+# see issue #1446
+VERSION = "3.2"
 
 # Code that discovers files or modules in INSTALLED_APPS imports this module.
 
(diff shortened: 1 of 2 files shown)


matthiask:
Interesting! I didn't know `pkg_resources` is so slow. I agree, let's remove it.

matthiask (approved):

auvipy (approved):

PR #1448, opened 2021-02-07
Improve local testing for contributors.

tim-schilling:
Changes make test to default to SQLite via tests/settings.py's defaults.

Changes the GitHub actions configuration to use database connection
credentials specific to the toolbar. This makes the local setup and
the CI tests more similar.

Pulls the database environment configuration out of GitHub actions
and into tox. This allows a developer to run tox and test against
all locally installed and setup databases. The contributing docs
have been updated to contain database shell commands to setup the
local environment.

This is a follow-up to jazzband/django-debug-toolbar#1431

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,8 +15,7 @@ jobs:
       mariadb:
         image: mariadb:10.3
         env:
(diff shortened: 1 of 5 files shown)


matthiask:
Thanks, this is great!

PR #1449, opened 2021-02-10
Fix #1434: Mention the need to run collectstatic in the installation notes

### LINKED ISSUE #1434: SQL Panel: + button is no longer functional in debug-tool-bar 3.2
Steps to reproduce:

1. Refresh a page
2. Open a the debug toolbar
3. Select the SQL panel
4. Press the + button for an SQL request

Result:
- Nothing happens, the + button is not responsive (whatever the SQL request)
- (There is no such issue in earlier versions, including 2.2, 3.0, 3.1.)

Expected result:
- The + button should show the details of the SQL request

Environement:
- django 3.1.5
- debug-toolbar 3.2
- python 3.6.8

matthiask:


matthiask:
sqlite tests are failing because Django@master (the upcoming 4.0) has dropped support for Python 3.6 and 3.7. See https://github.com/jazzband/django-debug-toolbar/pull/1450

tim-schilling:
I was under the impression that `collectstatic` is only needed in `DEBUG = False` environments. I feel like there's an underlying issue that should be investigated.

matthiask:
Good point. 

Also, we shouldn't recommend running collectstatic in development.

PR #1453, opened 2021-02-12
Fixed issue with toolbar expecting urls to start with __debug__.

tim-schilling:
When the history panel was added and we expanded the number of
urls we were instrumenting, there was a check added to exclude
the requests to the toolbar's views. This was done by checking
the path of the request to see if it started with `__debug__`.
However, the documentation states that is not a requirement.
This change checks the urls to see if the namespace of the
resolved url matches the toolbar's namespace.

This still isn't perfect as another url path could contain the
last namespace of djdt, but this is much less likely than a user
defining the toolbar's urls at a different path than `/__debug__/`

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -44,7 +44,7 @@ class DebugToolbarMiddleware:
     def __call__(self, request):
         # Decide whether the toolbar is active for this request.
         show_toolbar = get_show_toolbar()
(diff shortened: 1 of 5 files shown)


matthiask:
Thanks, this makes a lot of sense. I thought at first it may be a better idea to just reverse the debug toolbar's own URL and to a prefix match, but checking the namespace is probably better. The `.resolver_match` is even documented so that's good :)

PR #1455, opened 2021-03-06
Stop checking STATICFILES_DIRS with Django 4 or better

matthiask:
Django 4.0 added its own check for this (staticfiles.W004).


tim-schilling (approved):
Ha! I was in the process of opening my PR to fix this and saw yours. Thank you.

PR #1461, opened 2021-04-07
Drop support for Django 3.0.

hramezani:


matthiask:
Thanks! Do you have any idea where the coverage decrease comes from? We cannot drop backwards compatibility shims yet because of 2.2, I'm a bit surprised that it changed anything...

hramezani:
> Thanks! Do you have any idea where the coverage decrease comes from?

I don't know exactly. based on the report it comes from `debug_toolbar/panels/profiling.py lines 64,66,68` which seems weird to me.
After I pushed another commit it passed successfully :man_shrugging:

hramezani:
> We cannot drop backwards compatibility shims yet because of 2.2, I'm a bit surprised that it changed anything...

The `3.0.x` series no longer receive security updates or bug fixes.

matthiask:
> The `3.0.x` series no longer receive security updates or bug fixes.

Yes, but since we still support the older 2.2 series we cannot drop any backwards compatibility code yet.

Still, let's merge this since faster CI runs are good (tm).

PR #1465, opened 2021-04-18
Use current_thread instead of currentThread method that was deprecated in Python 3.10

### LINKED ISSUE #1464: currentThread was deprecated in Python 3.10 in favour of current_thread
Ref : https://github.com/python/cpython/pull/25174

```
rg -t py -w 'currentThread|notifyAll|activeCount|isDaemon|setDaemon' 
debug_toolbar/utils.py
255:            thread = threading.currentThread()
262:            thread = threading.currentThread()

debug_toolbar/panels/logging.py
81:        self._records[threading.currentThread()] = records

debug_toolbar/panels/staticfiles.py
121:        self._paths[threading.currentThread()] = used_paths
```

tirkarthi:
Fixes #1464

### DIFF
--- a/debug_toolbar/panels/logging.py
+++ b/debug_toolbar/panels/logging.py
@@ -78,6 +78,6 @@ class LoggingPanel(Panel):
 
     def generate_stats(self, request, response):
         records = collector.get_collection()
-        self._records[threading.currentThread()] = records
+        self._records[threading.current_thread()] = records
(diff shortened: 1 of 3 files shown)


matthiask:
Thanks! The `current_thread` spelling was introduced in Python 2.6 a very long time ago, this is fine :)

PR #1467, opened 2021-04-23
Use default app config discovery

francoisfreitag:
Django 3.2 deprecated default_app_config.
https://docs.djangoproject.com/en/3.2/releases/3.2/#automatic-appconfig-discovery

### DIFF
--- a/debug_toolbar/__init__.py
+++ b/debug_toolbar/__init__.py
@@ -1,3 +1,5 @@
+import django
+
 __all__ = ["VERSION"]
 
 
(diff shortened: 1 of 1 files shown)


matthiask:
Looks good, thanks!

cnobile2012:
It looks like this is still an issue with the latest version of `debug_toolbar`.
This is the __init__.py

```
$ cat __init__.py 
__all__ = ["VERSION"]


# Do not use pkg_resources to find the version but set it here directly!
# see issue #1446
VERSION = "3.2.1"

# Code that discovers files or modules in INSTALLED_APPS imports this module.

urls = "debug_toolbar.toolbar", "djdt"

default_app_config = "debug_toolbar.apps.DebugToolbarConfig"
```

The last line should probably be:
```
if django.VERSION < (3, 2):
    default_app_config = 'rest_framework.apps.RestFrameworkConfig'
```

francoisfreitag:
The [last release was Apr 14th](https://github.com/jazzband/django-debug-toolbar/releases), this PR was merged on Apr 23rd.
This code will appear in the next release.

PR #1468, opened 2021-04-23
Use twine to check generated package and readme

francoisfreitag:
The setup.py check command is deprecated. Use twine check instead.
Rename the tox environment to packaging, it now checks the package.


matthiask:
Thanks!

PR #1472, opened 2021-05-14
Keep the toolbar handle in bounds

matthiask:


debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -210,23 +211,24 @@ const djdt = {
             e.classList.remove("djdt-active");
         });
     },
+    ensure_handle_visibility() {
+        const handle = document.getElementById("djDebugToolbarHandle");
+        // set handle position
+        const handleTop = Math.min(
+            localStorage.getItem("djdt.top") || 0,
+            window.innerHeight - handle.offsetWidth
```
tim-schilling, line 219: Do we need to cover the case in which the `handle.offsetWidth` is greater than `window.innerHeight`?

matthiask: Thanks! `offsetWidth` is about `74px` right now, I don't think we care for the case when the window height is less than a few hundred pixels.

tim-schilling: Good stuff!

tim-schilling (approved):

matthiask:
Thanks!

PR #1477, opened 2021-06-19
Update installation.rst

mmdmua:
Windows10 err


mmdmua:
chrome browser not working. edge browser ok

PR #1478, opened 2021-06-19
Add test coverage to ensure that SQL tracker wrappers are applied only once to database cursors

jayaddison:
This is an attempt to add test coverage for the behaviour discovered in #1239 and addressed in #1475.

The idea is to [`patch`](https://docs.python.org/3.6/library/unittest.mock.html#unittest.mock.patch) the [`state.Wrapper` method](https://github.com/jazzband/django-debug-toolbar/blob/6ec6bfe0ba5912d7fd0a2b98176a215847157696/debug_toolbar/panels/sql/tracking.py#L23-L37) that is called to add each wrapping layer, and count the number of calls for both chunked and non-chunked types of cursor (across all database provider implementations).

cc @saemideluxe

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -1,6 +1,7 @@
 import datetime
 import os
 import unittest
+from unittest.mock import patch
 
 import django
 from django.contrib.auth.models import User
@@ -11,6 +12,7 @@ from django.shortcuts import render
 from django.test.utils import override_settings
 
 from debug_toolbar import settings as dt_settings
+import debug_toolbar.panels.sql.tracking as sql_tracking
 
 from ..base import BaseTestCase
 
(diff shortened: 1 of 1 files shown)


tim-schilling:
Thanks for this @jayaddison. Did you happen to run these tests before 1475 was merged to verify that this would catch the duplicate errors?

jayaddison:
@tim-schilling Although I only learned about the problem _after_ a fix had been found (always nice to learn that a bug has already been solved), yep I did run this test coverage on a version of the code without the fix in place, to make sure that it failed before-the-fix and passes after-the-fix (albeit a small modification required, in honesty partly due to issues encountered with Python mock/patching.. see #1479).

matthiask (approved):

PR #1479, opened 2021-06-21
Extract common base cursor wrapper class

jayaddison:
This change follows on from #1475 and introduces a common base class for cursor wrappers.

This base class is substituted in to the conditional check that identifies whether a cursor is already wrapped by DDT's SQL cursor tracking.

This should be compatible with the test coverage offered in #1478.


matthiask:
Thanks!

jayaddison:
Thank you!

Issue #1483, opened 2021-07-11
Issues with history view functions


ashwch:
I've been trying to use the history views in a SPA, and I've come across the following issues:

- Refresh fails if `DebugToolbar._store` is still being updated in parallel as dict can't be mutated while iteration is happening on it.
- Switch functionality doesn't work on rows loaded after refresh due to missing signature.
- Old expired rows remain in FE and result in 500 error(`history_sidebar`) as the `store_id` is not present in `DebugToolbar` anymore

Please let me know if any of this is an intended behaviour.

PR #1484, opened 2021-07-11
Fixes and improvements to history views

ashwch:
- Allow refreshing while requests are still happening by
  using a list of dict.items()
- Fixed issue with `history_refresh`, it now returns SignedDataForm
  to allow Switch requests to work after refresh.
- Clean up old rows in FE from that have now expired when refresh
  happens.
- Fixed `history_sidebar` view to not raise 500 error when no toolbar
  is found for an expired `store_id`.
- Old rows that have now expired, display [EXPIRED] next to the
  Switch text.

Fixes https://github.com/jazzband/django-debug-toolbar/issues/1483

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/history.js
+++ b/debug_toolbar/static/debug_toolbar/js/history.js
@@ -15,7 +15,6 @@ $$.on(djDebug, "click", ".switchHistory", function (event) {
 
     ajaxForm(this).then(function (data) {
         djDebug.setAttribute("data-store-id", newStoreId);
-        console.log("New id is" + newStoreId);
         // Check if response is empty, it could be due to an expired store_id.
         if (Object.keys(data).length === 0) {
             const container = document.getElementById("djdtHistoryRequests");
--- a/debug_toolbar/static/debug_toolbar/js/history.js
+++ b/debug_toolbar/static/debug_toolbar/js/history.js
@@ -7,21 +7,31 @@ $$.on(djDebug, "click", ".switchHistory", function (event) {
     const newStoreId = this.dataset.storeId;
     const tbody = this.closest("tbody");
 
-    tbody
-        .querySelector(".djdt-highlighted")
-        .classList.remove("djdt-highlighted");
-    this.closest("tr").classList.add("djdt-highlighted");
+    const highlighted = tbody.querySelector(".djdt-highlighted");
+    if (highlighted) {
+        highlighted.classList.remove("djdt-highlighted");
+        this.closest("tr").classList.add("djdt-highlighted");
+    }
 
     ajaxForm(this).then(function (data) {
         djDebug.setAttribute("data-store-id", newStoreId);
-        Object.keys(data).forEach(function (panelId) {
-            const panel = document.getElementById(panelId);
-            if (panel) {
-                panel.outerHTML = data[panelId].content;
-                document.getElementById("djdt-" + panelId).outerHTML =
-                    data[panelId].button;
-            }
-        });
+        console.log("New id is" + newStoreId);
+        // Check if response is empty, it could be due to an expired store_id.
+        if (Object.keys(data).length === 0) {
+            const container = document.getElementById("djdtHistoryRequests");
+            container.querySelector(
+                'button[data-store-id="' + newStoreId + '"]'
+            ).innerHTML = "Switch [EXPIRED]";
+        } else {
+            Object.keys(data).forEach(function (panelId) {
+                const panel = document.getElementById(panelId);
+                if (panel) {
+                    panel.outerHTML = data[panelId].content;
+                    document.getElementById("djdt-" + panelId).outerHTML =
+                        data[panelId].button;
+                }
+            });
+        }
     });
 });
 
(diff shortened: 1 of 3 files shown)


matthiask:
I think this looks good. I'm not sure about the change from `HistoryStoreForm` to `SignedDataForm`. Shouldn't it be possible to remove the former now?

tim-schilling:
I'm a bit swamped this week, but I will try to review this on Thu/Fri.

debug_toolbar/static/debug_toolbar/js/history.js
```diff
@@ -7,34 +7,46 @@ $$.on(djDebug, "click", ".switchHistory", function (event) {
...
     const tbody = this.closest("tbody");
 
-    tbody
-        .querySelector(".djdt-highlighted")
-        .classList.remove("djdt-highlighted");
-    this.closest("tr").classList.add("djdt-highlighted");
+    const highlighted = tbody.querySelector(".djdt-highlighted");
+    if (highlighted) {
+        highlighted.classList.remove("djdt-highlighted");
+        this.closest("tr").classList.add("djdt-highlighted");
```
tim-schilling, line 13: I think this line needs to be pulled out of the if statement. Otherwise if nothing is highlighted from the beginning, the if statement can never be true.

debug_toolbar/static/debug_toolbar/js/history.js
```diff
@@ -7,34 +7,46 @@ $$.on(djDebug, "click", ".switchHistory", function (event) {
...
         djDebug.setAttribute("data-store-id", newStoreId);
-        Object.keys(data).forEach(function (panelId) {
-            const panel = document.getElementById(panelId);
-            if (panel) {
-                panel.outerHTML = data[panelId].content;
-                document.getElementById("djdt-" + panelId).outerHTML =
-                    data[panelId].button;
-            }
-        });
+        console.log("New id is" + newStoreId);
```
tim-schilling, line 18: Is this an artifact of development?

debug_toolbar/static/debug_toolbar/js/history.js
```diff
@@ -7,34 +7,46 @@ $$.on(djDebug, "click", ".switchHistory", function (event) {
...
 $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     const container = document.getElementById("djdtHistoryRequests");
     ajaxForm(this).then(function (data) {
+        // Remove existing rows first then re-populate with new data
+        container
+            .querySelectorAll("tr[data-store-id]")
+            .forEach(function (node) {
+                node.remove();
+            });
```
tim-schilling, line 47: Could you do `container.innerHTML = ''` rather than iterating over the nodes?

tim-schilling (changes requested):
Thank you for the PR, there's a lot of great stuff in here! Having the additional context of expired switches will be helpful.

I'm not sold on the need to switch away from the `HistoryStoreForm` though. If we don't need to use the `SignedDataForm`, then I'd prefer to not use that.

I'm also curious if it's possible to write a test to cover the case of parallel operations mutating the store and causing it to break. How did you encounter that case originally?

debug_toolbar/static/debug_toolbar/js/history.js
ashwch: I remember trying `container.innerHTML = ''` during development and running into issues, probably it was due to something else. It works now.

ashwch:
@tim-schilling Thanks for your review.

With the above changes the switch URLs(even after refresh) look like this:

image

If we use `HistoryStoreForm`, the switch buttons loaded by the initial render of `History` have similar signed URLs, but after hitting refresh we end up with 400 errors for any switch: `http://localhost:8000/__debug__/history_sidebar/?store_id=3be1d8503a0b46999cb4cb54efc1f2a1` as this results in invalid `signed_form` in `signed_data_view` decorator: `{'signed': ['This field is required.']}`

**Dict mutation test:** It's quite tricky to prove in a test. But here's an attempt in my fork https://github.com/ashwch/django-debug-toolbar/tree/dict_mutation_test: https://github.com/ashwch/django-debug-toolbar/blob/c0454b6c20e09b0db8d1d3eae67429cd7eb07bb3/tests/panels/test_history.py#L138.

The sleep time of `0.002` could be different based on the system, hence I personally don't like the test much.

```
make test TEST_ARGS='tests.panels.test_history.HistoryViewsTestCase.test_parallel_history_refresh'
```

tim-schilling (approved):
Thank you for the changes and further explanation. After pulling the PR and running the tests myself, it finally clicked. You're absolutely correct we should be wrapping the form should be a `SignedDataForm` instance, but we should still by supplying the `HistoryStoreForm().initial` rather than bypassing it. The purpose for this is if we ever change `HistoryStoreForm`'s shape in the future, that will be included. Otherwise we'd have to remember to also make the change here off of pure memory which I'm probably going to fail at.

I've made the change to your branch. I'm good with these changes, thank you for the PR!

matthiask:
Thanks, both of you! I think this looks good. Let's merge this.

PR #1488, opened 2021-07-28
Support for request-level urlconf overrides

NielkS:
Middlewares can override the default urlconf per request by setting a "urlconf" attribute to the inbound HttpRequest instance
(doc: https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.urlconf )
This change adds support for this kind of override

### DIFF
--- a/debug_toolbar/toolbar.py
+++ b/debug_toolbar/toolbar.py
@@ -146,7 +146,9 @@ class DebugToolbar:
         # The primary caller of this function is in the middleware which may
         # not have resolver_match set.
         try:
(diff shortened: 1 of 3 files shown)


tim-schilling (changes requested):
Thanks @NielkS for the PR! Can you add a test that verifies this functionality? Please also add a mention of this in the docs/changes.rst

tim-schilling:
Ah, looks like the linter failed. You can run that locally with `make style`

tim-schilling:
Rebased on main, added a test and updated changes documentation. Merging after CI passes.

tim-schilling:
@NielkS thank you for the PR!

PR #1493, opened 2021-08-12
Add warning for having django.contrib.postgres after debug_toolbar.

tim-schilling:
Having the INSTALLED_APPS misconfigured in this was can lead to HStoreFields
throwing can't adapt type 'dict' messages.

This closes PR #804

Given the issues we've had with folks installing the toolbar, I think adding more documentation to the installation section is no longer the right approach. Instead logging a warning to the user when the apps are misconfigured should catch the developer's attention and be corrected.

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -22,6 +22,8 @@ def check_middleware(app_configs, **kwargs):
     errors = []
     gzip_index = None
     debug_toolbar_indexes = []
+    postgres_app_index = None
+    debug_toolbar_app_index = None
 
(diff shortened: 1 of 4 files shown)


matthiask (approved):

matthiask:
The change looks good and makes sense to me (except for the test failures of course)

tim-schilling:
I realized I should actually try to reproduce these messages first. I may end up closing both PRs.

tim-schilling:
Yarp, I can't reproduce these messages on 2.2 or 3.2. I tried creating instances via the admin and using querysets that do some filtering. I'm going to close both PRs without any changes since the original PR is a few years old.

PR #1494, opened 2021-08-14
Update translations and revert xgettext aliasing.

tim-schilling:
Aliasing xgettext causes the usage not to be picked up by makemessages.
Only gettext and gettext_lazy work with aliasing.


matthiask:
Great find!

tim-schilling:
Do you know of any good way to solicit translators?

matthiask:
Good question ... :-/

tim-schilling:
I suspect transifex does do notifications, but probably a good idea to solicit for some help.

tim-schilling:
Notifications for folks who have submitted translations in the past*

PR #1495, opened 2021-08-24
Utilize pre-commit to help devs follow style guidelines

### LINKED ISSUE #1485: Setup pre-commit for repo
To help lower barriers to contributing we should integrate [pre-commit](https://pre-commit.com/) to the project. The main use for us would be to run the linter before a commit occurs to help prevent failed tests when PRs are opened. While this does change the setup time, it does conform to a larger python practice. That conformity should theoretically be more accessible when compared to knowing to run `make styles` before committing.

tim-schilling:
Closes #1485

@matthiask one thing that requires a close look is .pre-commit-config.yaml. Let me know if I should remove any that you think are overly strict. I picked from https://pre-commit.com/hooks.html and matched what we had. I did introduce doc8 since it seems like a positive change.


matthiask (approved):
I like it, especially the strictness of it :)

Regarding `F401`: You could add e.g. `__all__ = ["SQLPanel"]` to avoid the need to silence this particular error, but it's definitely fine as it is.

tim-schilling:
I'm glad I forced the specific flake8 warnings, TIL. I'll make that change and merge it in.

PR #1497, opened 2021-08-27
Changed cache monkey-patching for Django 3.2+

### LINKED ISSUE #1496: DJDT not showing low level caching correctly in Docker
Hi,
I working with DJDT for a few months.
I figure out the "not showing" issue in docker, but the data inside the cache panel is not showing properly.
The inner data, doesnt match the outer data **and** in the arguments column it shows addresses or something alike.

A proper data: (taken from an online guide)
image
(From: https://testdriven.io/blog/django-low-level-cache/#cache-backend)

What it shows inside a Docker:
ch

BTW: It also shows it like a view cache, but the view is not cached (Its a dynamic view)
My Configuration:

```

DEBUG_TOOLBAR_ENABLED = True
if DEBUG and DEBUG_TOOLBAR_ENABLED:
    INSTALLED_APPS.append('debug_toolbar')
    MIDDLEWARE.insert(0, 'debug_toolbar.middleware.DebugToolbarMiddleware')
    MIDDLEWARE.insert(1, 'django.middleware.cache.UpdateCacheMiddleware')
    MIDDLEWARE.append('django.middleware.cache.FetchFromCacheMiddleware')
    #INTERNAL_IPS.append('127.0.0.1')

    DEBUG_TOOLBAR_CONFIG = {
        'SHOW_TOOLBAR_CALLBACK': lambda request: False if request.is_ajax() else True,
    }

```

Does anyone have idea how to solve it ?
Where to start ?

tim-schilling:
Changed caching monkey-patching for Django3.2+ to iterate over existing
caches and patch them individually rather than attempting to patch
`django.core.caches` as a whole. The `middleware.cache` is still
being patched as a whole in order to attempt to catch any cache
usages before enable_instrumentation is called.

Fixes #1496

Big thanks to @LiorA1 for reporting the issue and creating a reproducible test case.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,6 +4,12 @@ Change log
 Next version
 ------------
 
+* Changed caching monkey-patching for Django3.2+ to iterate over existing
+  caches and patch them individually rather than attempting to patch
+  ``django.core.caches`` as a whole. The ``middleware.cache`` is still
+  being patched as a whole in order to attempt to catch any cache
+  usages before ``enable_instrumentation`` is called. Fixes #1496
+
 
 3.2.2 (2021-08-14)
 ------------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,11 +4,11 @@ Change log
 Next version
 ------------
 
-* Changed caching monkey-patching for Django3.2+ to iterate over existing
+* Changed cache monkey-patching for Django 3.2+ to iterate over existing
   caches and patch them individually rather than attempting to patch
   ``django.core.caches`` as a whole. The ``middleware.cache`` is still
   being patched as a whole in order to attempt to catch any cache
-  usages before ``enable_instrumentation`` is called. Fixes #1496
+  usages before ``enable_instrumentation`` is called.
 
 
 3.2.2 (2021-08-14)
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     ConnectionProxy = None
 
+import django
 from django.conf import settings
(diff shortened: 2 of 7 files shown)


tim-schilling:
> Shoot. This still needs to monkey-patch django.core.cache.caches for other caches that get created after enable_instrumentation is called.

That would require changing the settings which is a no-no. So I was right the first time.

tim-schilling:
The reduction in code coverage is because our report doesn't handle the coverage across different versions (I think).

docs/changes.rst
```diff
@@ -4,6 +4,12 @@ Change log
 Next version
 ------------
 
+* Changed caching monkey-patching for Django3.2+ to iterate over existing
```
matthiask, line 7:
```suggestion
* Changed cache monkey-patching for Django 3.2+ to iterate over existing
```

docs/changes.rst
```diff
@@ -4,6 +4,12 @@ Change log
...
+  caches and patch them individually rather than attempting to patch
+  ``django.core.caches`` as a whole. The ``middleware.cache`` is still
+  being patched as a whole in order to attempt to catch any cache
+  usages before ``enable_instrumentation`` is called. Fixes #1496
```
matthiask, line 11:
```suggestion
  usages before ``enable_instrumentation`` is called.
```

Do we want to start adding ticket references? It may be useful but I'd rather not have to do this additional work :)

matthiask (approved):
> The reduction in code coverage is because our report doesn't handle the coverage across different versions (I think).

I can't say for sure but I think you're right.

docs/changes.rst
tim-schilling: You're right.

Issue #542, opened 2014-02-09
Empty Context Processor


thebonzitree:
<p>I have a case where I sometimes add a variable to context via a context processor and sometimes it is an empty dict depending on some logic. In such a case Django Debug toolbar errors with <p>'NoneType' object is not iterable</p>

<p>I believe it the issue is in the "_request_context__init__" method in the panel.py file.  Seems we are checking for the context processors here but did not put in the logic for if none.  My fix was to simply change the line:</p>

<code>line 57.  updates.update(context)</code>

<p>to</p>

<code>if context:  updates.update(context)</code>

<p>Not sure if this has other implications but seems to work for now.  If it is not a bug and is my own misunderstanding please let me know.</p> <p>I figured I would let you guys know...keep up the good work.</p>

christopherpickering:
I have the same problem - it comes when a user has not authenticated yet. I patch my local install by adding
```py
if context:
    updates.update(context)
```

I'll put a pr for it.

christopherpickering:
I think I found what caused this in my case...

I had a custom context processor in my django config for templates. One of my processors was to get a few user props for logged in users, and was only returning a context for logged in users. It seems the a context processor must ALWAYS return, even if empty.

If I have a processor ```"context_processors.user",```

```py
def user(request):
    """Context processor for user information."""
    if not request.user.is_anonymous:
        return {
            "permissions": request.user.get_permissions(),
            "user": request.user,
            "favorites": request.user.get_starred_reports(),
        }
    # context processor must ALWAYS return a dict
    return {}
```

tim-schilling:
@matthiask There's a relatively recent ticket for adding a check to Django for context processors that return a non-dictionary value. I've seen this occur a few times now. I'm in favor of logging an warning/error in the console when this is encountered, but allow the toolbar to continue processing. This is a painful problem for developers and a source of support tickets for us.

matthiask:
Yes, I agree. It would definitely be an improvement if django-debug-toolbar wouldn't crash :)

PR #1502, opened 2021-09-16
only update if context exists

christopherpickering:
fix for #542

fixes this error that comes if a user is not authenticated yet.
image


christopherpickering:
closing... problem was in custom context processor. https://github.com/jazzband/django-debug-toolbar/issues/542#issuecomment-920894735

PR #1505, opened 2021-09-24
Update DE locale

Baltrunas:
msgid "Name" and msgid "Key" vas same translation as "Name", also it has negative effect for other modules


tim-schilling:
I'm not sure why I triggered the tests. We need to make this change via Transifex. Let me get that squared away.

matthiask:
I agree, this is a better translation.

(Maybe `pgettext` should be used more often...)

PR #1507, opened 2021-09-28
Possibly fix an infinite recursion crash by avoid str() inside repr()

matthiask:
See https://code.djangoproject.com/ticket/33134


tim-schilling:
Good catch.

PR #1523, opened 2021-11-04
Run CI tests weekly.

tim-schilling:
This helps identify issues with the library as
django and other dependencies release new versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,11 @@
 name: Test
 
-on: [push, pull_request]
+on:
+  push:
+  pull_request:
+  schedule:
+    # Run weekly on Saturday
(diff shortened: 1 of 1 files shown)


matthiask (approved):
That is an excellent idea! The MySQL failures are definitely unrelated, something changed in the Django@main version...

(If you wanted to avoid too much load on Github you could switch to a weekly schedule but it probably doesn't make a big difference.)

tim-schilling:
I did a quick look on the interwebs for a way to trigger based on other packages releases, but the best option is dependabot. I think we could do this, but I'll open an issue for discussion.

tim-schilling:
I also changed this to run weekly. It's more environmentally friendly and frankly we likely won't be fixing integration issues the day after the crop up.

PR #1349, opened 2020-09-27
Improve "Setting up URLconf" section in installation docs

jdufresne:
Remove the "if settings.DEBUG" guard. This guard is not necessary and
leads to several edge cases in local development setups.

This is not necessary as all views are guarded by the
require_show_toolbar decorator which uses the SHOW_TOOLBAR_CALLBACK
setting. This setting is already capable of enabling/disabling the
across the application when DEBUG is False. By default, when DEBUG is
False, these views return a 404 response.

By having the URLs always available, the MIDDLEWARE setting also doesn't
need special handling when DEBUG is False, simplifing the overall
configuration. An application should be able to reliably reverse URLs
regardless of the status of DEBUG.

Fixes #1035
Fixes #1043
Fixes #1332

PR #1528, opened 2021-11-15
Remove settings import from install instructions.

adamchainz:
The import of settings is no longer required after #1349

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -51,11 +51,10 @@ Setting up URLconf
 Add the Debug Toolbar's URLs to your project's URLconf::
 
     import debug_toolbar
-    from django.conf import settings
     from django.urls import include, path
 
     urlpatterns = [
-        ...
+        # ...
         path('__debug__/', include(debug_toolbar.urls)),
     ]
 


matthiask (commented):
Thanks! This makes the installation instructions more consistent.

Now that I look at the snippet again I worry a bit about potential security implications. The default `SHOW_TOOLBAR_CALLBACK` *does* make this safe but I don't know whether that's good defense in depth.

adamchainz:
I think it's sufficient protection. Asking users to put `if settings.DEBUG` in URLs is just “pushing the ‘if’ around”. I'm sure ~99% of users don't change `SHOW_TOOLBAR_CALLBACK`.

matthiask:
> Though we should add a check/warning that identifies when the callback has been changed from the default and the toolbar is running with DEBUG=False. Or something that covers that case.

I think we should add a big warning to the `SHOW_TOOLBAR_CALLBACK` docs and leave it at that. I'll submit a pull request right away. I merged this PR because the change itself isn't related to the security issue we're discussing.

Thanks @adamchainz !

adamchainz:
> I think we should add a big warning to the `SHOW_TOOLBAR_CALLBACK` docs and leave it at that. I'll submit a pull request right away.

+1 to that. That would be the same kind of thing we'd do for Django core: sensible defaults, with the ability to override if required with warnings.

PR #1534, opened 2021-11-24
Fix settings docs

adamchainz:
* Remove comment in ``debug_toolbar/settings.py`` that was invalidated in a8fa2cf467d8a8234b16e4bb9f45fbaa60597392.
* Fix documented default for ``DISABLE_PANELS``.


matthiask:
Thank you! That's great.

PR #1535, opened 2021-11-24
Use only pre-commit for style linters and fixers

adamchainz:
The method of using the `Makefile`, running under `tox`, duplicated the work of pre-commit and pre-commit.ci. Additionally tox did not use the same pinned versions and arguments as in `.pre-commit-config.yaml`, so there is the potential for a version clash to cause the two paths to conflict, e.g. when a new version of Black is released.

Sticking to only `pre-commit` seems sensible, saves some CI time, and allows some files to be removed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -221,4 +221,4 @@ jobs:
         python -m pip install --upgrade tox
 
     - name: Test with tox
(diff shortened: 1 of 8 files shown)


matthiask:
Thanks! I like this very much. I always forgot how to run the Makefile commands anyway...

PR #1537, opened 2021-12-04
Create urls module and remove import package from docs.

### LINKED ISSUE #1506: Include urls without importing package
I don't want to have the `django-debug-toolbar` package installed in my production environment, so I'm looking for alternatives to a slightly "dirty" import which is not at the top of the file (see [PEP-402](https://www.python.org/dev/peps/pep-0402/)):
```
if "debug_toolbar" in settings.INSTALLED_APPS:
    import debug_toolbar
    urlpatterns.append(url(r"^__debug__/", include(debug_toolbar.urls)))
```
But the obvious string-module-path
```
    urlpatterns.append(url(r"^__debug__/", include("debug_toolbar.urls")))
```
fails due to the error `ModuleNotFoundError: No module named 'debug_toolbar.urls'`.

However, I noticed that
```
urlpatterns.append(url(r"^__debug__/", include("debug_toolbar.toolbar")))
```
seems to work, so I wonder why this isn't documented. Are there any downsides to this approach, or might this stop to work in future versions of the django debug toolbar?

Thanks for any comments on this!

Django version: 3.2.7
Django Debug Toolbar version: 3.2.2

tim-schilling:
By creating a urls module, devs can include the urls with a string
path to the urls, include('debug_toolbar.urls').

Closes #1506

### DIFF
--- a/debug_toolbar/__init__.py
+++ b/debug_toolbar/__init__.py
@@ -1,5 +1,7 @@
 import django
 
+from debug_toolbar.urls import app_name
+
 __all__ = ["VERSION"]
 
 
@@ -9,7 +11,7 @@ VERSION = "3.2.2"
 
 # Code that discovers files or modules in INSTALLED_APPS imports this module.
 
-urls = "debug_toolbar.toolbar", "djdt"
+urls = "debug_toolbar.urls", app_name
 
 if django.VERSION < (3, 2):
     default_app_config = "debug_toolbar.apps.DebugToolbarConfig"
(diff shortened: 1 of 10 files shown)


tim-schilling:
I ran the tests with the previous urls definition where it imports the package, then uses `include(debug_toolbar.urls)` and confirmed that this will be compatible. @matthiask do you think we should formalize this in the tests?

matthiask:
Thanks, this looks great!

> I ran the tests with the previous urls definition where it imports the package, then uses `include(debug_toolbar.urls)` and confirmed that this will be compatible. @matthiask do you think we should formalize this in the tests?

If it's relatively easy to do then yes – this will mean less work when updating old projects; OTOH it's one of the smaller changes and django-upgrade helps with the more annoying changes so maybe it's fine to either fix it when it breaks later or just document it as possibly backwards incompatible.

tim-schilling:
Nah, it'll be easy. A second urls module then overriding the root URL conf
setting will do the trick.

PR #1539, opened 2021-12-08
Discourage insecure SHOW_TOOLBAR_CALLBACK values

matthiask:
A leaked SECRET_KEY combined with an insecure SHOW_TOOLBAR_CALLBACK
configuration may lead to remote execution of arbitrary SQL.

The toolbar shouldn't be installed in production environments anyway.


tim-schilling (approved):
Thanks @matthiask!

PR #1544, opened 2021-12-14
Avoid stripping additional characters from GIS SQL params.

### LINKED ISSUE #1543: Some String Has Truncated Before Model Save In  Releases Version==3.2.3
#### I found some string was truncated in this function

```python
def strip_GeomFromEWKB(param):
    if isinstance(param, str):
        return param.lstrip("ST_GeomFromEWKB('\\x").rstrip("'::bytea)")
    return param
```

#### Related Pull Request

- https://github.com/jazzband/django-debug-toolbar/pull/1426

####  Minimally Reproducible Test Case

```python
In [8]: "botpy".lstrip("ST_GeomFromEWKB('\\x").rstrip("'::bytea)")
Out[8]: 'botp'

In [9]: "Botpy".lstrip("ST_GeomFromEWKB('\\x").rstrip("'::bytea)")
Out[9]: 'tp'
```

tim-schilling:
If the param lead or ended with characters identified in the strip
calls, those characters would also be removed. Using split will
retain all of the characters in between the given strings.

When we drop support for < 3.9, we can switch to
the removeprefix and removesuffix functions.

Fixes #1543

This should add a test case that confirms the fix for 1543, but I'm unsure of what that would look like. If we think we absolutely need a test case for it, then perhaps we move it to `panels/sql/utils.py` so it can be in a unit test rather than an integration test.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -142,6 +142,73 @@ jobs:
       with:
         name: Python ${{ matrix.python-version }}
 
+  postgis:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 5
+      matrix:
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
+
+    services:
+      postgres:
+        image: postgis/postgis:14-3.1
+        env:
+          POSTGRES_DB: debug_toolbar
+          POSTGRES_USER: debug_toolbar
+          POSTGRES_PASSWORD: debug_toolbar
+        ports:
+        - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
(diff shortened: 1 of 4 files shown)


tim-schilling:
Ah, this needs to not run when it's not a GIS sql param.

matthiask:
Now that I read the code again I'm really unsure. We're modifying the SQL which is being executed instead of only logging it and maybe fixing it later when trying to analyze/explain it.

Is there a good reason for modifying the `params` list before calling `cursor.execute`, `cursor.callproc` etc. at all?

tim-schilling:
I'm not sure. Even though the tests indicate we test for postgis, it's not used. I'm working through setting it up locally to see if the tests pass.

tim-schilling:
@matthiask I'm going to revert #1426. I don't think it ever covered the change and the test is broken.

tim-schilling:
Replaced by #1547

PR #1547, opened 2021-12-15
Revert PR 1426 - PostGIS param stripping.

### LINKED ISSUE #1543: Some String Has Truncated Before Model Save In  Releases Version==3.2.3

tim-schilling:
The stripping logic is stripping general sql parameters and the test
written for it does not run. I'm not confident that the test actually
hits the code we expect it to run either.

Closes #1543
Reopens #423

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,7 +82,7 @@ jobs:
       max-parallel: 5
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
-        database: ['postgres', 'postgis']
+        database: [postgres, postgis]
 
     services:
       postgres:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,10 +82,11 @@ jobs:
       max-parallel: 5
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9', '3.10']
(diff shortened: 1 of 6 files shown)


tim-schilling:
I kept the postgis testing so things are easier to manage in the future if we reattempt #423. Though I was close to removing it all.

matthiask:
Thanks! For the record, I definitely agree with this decision. I don't know what I was thinking when merging the PR, probably not too much :/

tim-schilling:
Eh, it happens. I wouldn't lose sleep over it. I'm not sure there's a great answer to preventing this in the future other than more involvement on our parts when assessing PRs.

PR #1551, opened 2021-12-16
Drop support for Django 2.2

hramezani:


matthiask:
@hramezani Please try again, I pushed the wrong branch but I've reverted the change.

matthiask (commented):
Looks good already. There are still a few `django.VERSION` branches around which could be dropped as well, but we don't have to clean up everything at once.

hramezani:
@matthiask I will find them and remove them

PR #1556, opened 2021-12-18
Don't raise W006 warning when app loader is specified.

### LINKED ISSUE #1550: False positive for debug_toolbar.W006?
After upgrading to v3.2.3 I get the system warning W006, and an advisement to set _APP_DIRS=True_ for at least one of my template settings.

Shouldn't that check also be looking at the value of _loaders_ in order to avoid false warnings?

For example, the project in question had the following template settings:

```python
TEMPLATES = [{
    'BACKEND': 'django.template.backends.django.DjangoTemplates',
    'DIRS': [SITE_ROOT / 'templates'],
    'OPTIONS': {
        'builtins': [
            'django.templatetags.static',
            'common.templatetags.builtins',
        ],
        'context_processors': [
            'common.context_processors.constants',
            'django.template.context_processors.debug',
            'django.template.context_processors.request',
            'django.template.context_processors.media',
            'django.contrib.auth.context_processors.auth',
            'django.contrib.messages.context_processors.messages',
        ],
        'loaders': [
            'django.template.loaders.filesystem.Loader',
            'django.template.loaders.app_directories.Loader',
        ],
    },
}]
```
Django Debug Toolbar works wonderfully with these settings. My understanding is that the _app_directories_ loader is able to find the templates that the toolbar needs. The W0006 warning seems to be a false postive here (but to be honest, it's been a couple of years since I created this config).

As a quick experiment I tried just deleting the loaders option and adding the suggested APP_DIRS setting. It resulted in a key error for 'django' during deployment while trying to run migrate - so I've reverted back to v3.2.2 for now.

tim-schilling:
Fixes #1550

Thank you @leon-matthews for the report!

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,7 +5,7 @@ Next version
 ------------
 
 * Removed support for Django < 3.2.
-* Update check ``W006`` to look for
+* Updated check ``W006`` to look for
   ``django.template.loaders.app_directories.Loader``.
 
 3.2.4 (2021-12-15)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,6 +5,8 @@ Next version
 ------------
 
 * Removed support for Django < 3.2.
+* Update check ``W006`` to look for
+  ``django.template.loaders.app_directories.Loader``.
 
 3.2.4 (2021-12-15)
 ------------------
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -22,6 +22,24 @@ class DebugToolbarConfig(AppConfig):
         DebugToolbar.get_panel_classes()
 
 
+def check_template_config(config):
+    """
+    Checks if a template configuration is valid.
+
+    The toolbar requires either the toolbars to be unspecified or
+    ``django.template.loaders.app_directories.Loader`` to be
+    included in the loaders.
+    If custom loaders are specified, then APP_DIRS must be True.
+    """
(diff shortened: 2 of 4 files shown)


matthiask (approved):
Nice! A good side-effect of this is that it annoys people even more who use the toolbar with `DEBUG=False`.

docs/changes.rst
```diff
@@ -5,6 +5,8 @@ Next version
 ------------
 
 * Removed support for Django < 3.2.
+* Update check ``W006`` to look for
```
matthiask, line 8:
```suggestion
* Updated check ``W006`` to look for
```

tim-schilling:
> A good side-effect of this is that it annoys people even more who use the toolbar with DEBUG=False.

It does?

matthiask:
> > A good side-effect of this is that it annoys people even more who use the toolbar with DEBUG=False.
> 
> It does?

Possibly, if they are using the old pattern of adding the cached Loader when `DEBUG` is falsy in the settings. Or not? Maybe I'm not thinking clearly (again).

tim-schilling:
From my understanding, if the user specifies any loader, those are the loaders that are used. No additional loaders, included the cached loader will be used.

If they don't specify loaders, then Django will specify the filesystem loader, and if app_dirs is True, then the app loader. Then if debug is false it will use the cached loader with whichever are used.


Sorry, my brain isn't working at 100%. I'm probably being pedantic.

PR #1557, opened 2021-12-19
Remove the codecov badge, we're not using codecov anymore

matthiask:
Refs #1542

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@
 Django Debug Toolbar |latest-version|
 =====================================
 
-|jazzband| |build-status| |codecov| |docs| |python-support| |django-support|
+|jazzband| |build-status| |docs| |python-support| |django-support|
 
 .. |latest-version| image:: https://img.shields.io/pypi/v/django-debug-toolbar.svg
    :target: https://pypi.python.org/pypi/django-debug-toolbar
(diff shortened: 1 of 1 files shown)


pauloxnet:
I think it would be better to replace the coverage badge with another one related to GitHub action instead of simply remove it.

matthiask:
I don't think that's easily possible without e.g. committing the resulting coverage to the repository.

Something like https://pypi.org/project/coverage-badge/ or https://shields.io/category/coverage could probably be used to generate the badge, but we'd need to store the percentage somewhere.

pauloxnet:
We have htmlcov in artifacts don't we already store the coverage percentage anywhere with GitHub Actions?

tim-schilling:
I'm going to give this a try: https://dev.to/thejaredwilcurt/coverage-badge-with-github-actions-finally-59fa

tim-schilling:
Never mind. I don't have access to the repo settings. Do you @matthiask?

tim-schilling:
> We have htmlcov in artifacts don't we already store the coverage percentage anywhere with GitHub Actions?

Nope, the change uploads the report as an artifact, but I don't think the percentage isn't stored. I suppose we could use the failure threshold in the badge.

tim-schilling:
Welp I messed that up. I intended for it to be a suggestion not an actual commit.

matthiask:
> Never mind. I don't have access to the repo settings. Do you @matthiask?

No. I think that's roadies-only territory.

matthiask:
I've been thinking about this for a bit. What purpose does showing the coverage in a badge serve? I don't think it's very useful. Code coverage is at best an approximation of the quality of a testsuite. It tests that the code doesn't crash right away, but that's basically all it proves. Correctness isn't required. So, it may be useful when writing new tests but I don't think it's a badge of honor to be worn proudly. I think we don't lose anything when we get rid of it.

tim-schilling:
I don't feel strongly one way or the other here. I'm content to remove it. Before we do though, here's a benefit of the coverage badge that you didn't explicitly mention. Having the coverage percentage listed on the README, along with a passing tests badge increases the project's trustworthiness for being well maintained. That can improve a developer's confidence that this project is something that can be integrated and will make them more likely to use it.

pauloxnet:
I agree with @tim-schilling and I approved the change.

pauloxnet (approved):

matthiask:
Re showing that the project is well maintained: Yes, that's a good point. I more often look at the timeline of recent commits and the CI jobs list to determine this, but the coverage may be telling.

PR #1558, opened 2021-12-20
 Specify code coverage as a static value. 

matthiask:


matthiask:
@tim-schilling I added your suggestion here as a separate pull request, we can continue the discussion here.
 
cc @pauloxnet

pauloxnet (approved):

pauloxnet:
@matthiask can you resolve conflicts ?

pauloxnet:
@matthiask the PR is ready to merge. I see it's still in draft. There's a reason or you can simply  remove the draft tag ?

matthiask:
> @matthiask the PR is ready to merge. I see it's still in draft. There's a reason or you can simply remove the draft tag ?

I opened it as a draft because I wasn't sure whether it is a good idea to hardcode the coverage percentage. I'm marking it as ready for review now because I'm basically neutral, not opposed to this.

tim-schilling:
@matthiask what do you think about that [gist revision approach](https://dev.to/thejaredwilcurt/coverage-badge-with-github-actions-finally-59fa)? I'd prefer not to set it statically, but I will if it means the alternative is not having it.

matthiask:
> @matthiask what do you think about that [gist revision approach](https://dev.to/thejaredwilcurt/coverage-badge-with-github-actions-finally-59fa)? I'd prefer not to set it statically, but I will if it means the alternative is not having it.

This looks interesting! I suspect that we'd need Jannis' help for this because we don't have access to the repository settings?

I'd merge this pull request and open an issue to re-investigate later if there's a better solution, maybe directly offered by GitHub or someone.

tim-schilling (approved):
I added a comment indicating to change the readme's coverage badge link in case we don't replace this in the future.

PR #1559, opened 2021-12-20
Move settings reset logic to settings module.

adamchainz:
This makes it apply in every project using django-debug-toolbar, so packages or projects can reconfigure the toolbar during tests if required.

The receiver functions have also been combined and optimized the signature like https://github.com/django/django/pull/15185 .


matthiask:
Thanks!

PR #1560, opened 2021-12-20
Avoid installing middleware if Content-Encoding is set at all

adamchainz:
The previous check for "gzip" ignored the possibility of other `Content-Encoding` values, such as Brotli’s ``br``, which the middleware could also not touch. Instead, do not attempt to alter the content for *any* `Content-Encoding`.

I tested by installing django-brotli in the example app below the toolbar middleware and hitting it with:

```
curl http://localhost:8000/ -H 'Accept-Encoding: br'
```

Without the change, it would crash with:

```
  File "/.../debug_toolbar/middleware.py", line 87, in __call__
    content = response.content.decode(response.charset)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc7 in position 1: invalid continuation byte
```

After, it is fixed.

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -78,7 +78,7 @@ class DebugToolbarMiddleware:
         content_type = response.get("Content-Type", "").split(";")[0]
         if (
             getattr(response, "streaming", False)
-            or "gzip" in content_encoding
+            or content_encoding != ""
             or content_type not in _HTML_TYPES
         ):
             return response


tim-schilling (approved):

matthiask:
Thanks! It would be relatively easy to write a test for this but.... that's basically already a concern and not introduced by this proposal (following holacracy's governance rules)

adamchainz:
I am happy to find a test but couldn't pin down where instantly - if you can point me I will.

tim-schilling:
Duplicating [`test_middleware_response_injection`](https://github.com/jazzband/django-debug-toolbar/blob/main/tests/test_integration.py#L91) where it has `django.middleware.gzip.GZipMiddleware` applied before sending it to the toolbar's middleware would cover the regression case of it. Alternatively manipulating the `Content-Encoding` directly without the additional middleware.

adamchainz:
Thanks Tim, test PR in #1562.

PR #1563, opened 2021-12-20
Ignore all coverage files from Git

adamchainz:
Since #1542, coverage creates parallel files, with names like `.coverage.<hostname>.<pid>.<someothernumber>`, which need ignoring too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 *~
 .idea
 build
-.coverage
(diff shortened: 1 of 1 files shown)


adamchainz:
@tim-schilling

tim-schilling:
D'oh. I should have caught that myself. Thank you.

PR #1566, opened 2021-12-23
Fix #1565 Remove Python 3.6 due to EOL

### LINKED ISSUE #1565: Remove Python 3.6
Due to today's Python 3.6 end of life I propose to remove it from Django Debug Toolbar

pauloxnet:


matthiask:
Thanks! 

Django 3.2 supports the whole range from 3.6 to 3.10. Django 4.0 only supports 3.8 or better. Maybe we should wait a bit and drop Django<4 and Python<3.8 at the same time?

tim-schilling:
> Django 3.2 supports the whole range from 3.6 to 3.10. Django 4.0 only supports 3.8 or better. Maybe we should wait a bit and drop Django<4 and Python<3.8 at the same time?

I'm good with this unless some issue comes up for 3.6 that is a pain to solve before then. Effectively, soft support of 3.6 from now on.

matthiask:
Less CI runs is good for the environment so let's do this.

PR #1570, opened 2022-01-11
Implement CacheStatTracker.get_or_set

francoisfreitag:
Capture the get_or_set() calls to the cache, keeping the tracker
interface in line with django.core.cache.backends.BaseCache.


matthiask:
Thanks! That's excellent.

PR #1571, opened 2022-01-12
Optimize render_stacktrace()

adamchainz:
This function is called by the SQL panel to render each frame of each stack trace.

Previously it used a Django template, which is slow. By generating the HTML with pure Python code, we can avoid much of this overhead.

I tried this change on a demo app I built that has 96 total queries, due to N+1 queries. The queries execute in 13ms due to use of SQLite, so the whole view is a negligible concern. Nearly all the view runtime is the toolbar itself. Without this change, the runtime is ~1300ms; with the change it’s ~1100ms. That's a saving of **15%**.

I also checked the appearance of the generated HTML hasn’t changed.

### DIFF
--- a/debug_toolbar/utils.py
+++ b/debug_toolbar/utils.py
@@ -3,12 +3,12 @@ import os.path
 import re
 import sys
 from importlib import import_module
-from itertools import chain
+from pprint import pformat
 
 import django
 from django.core.exceptions import ImproperlyConfigured
 from django.template import Node
-from django.template.loader import render_to_string
+from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 
 from debug_toolbar import settings as dt_settings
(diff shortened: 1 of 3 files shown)


matthiask (approved):
Thanks! I like it. 

I think we had reports about `pformat` being slow in the past, maybe we should skip `pformat`-ing when `PRETTIFY_SQL` is `False` but that would be an additional change/improvement.

debug_toolbar/utils.py
```diff
@@ -69,26 +69,31 @@ def tidy_stacktrace(stack):
...
-    stacktrace = []
-    for frame in trace:
-        params = (v for v in chain(frame[0].rsplit(os.path.sep, 1), frame[1:]))
-        params_dict = {str(idx): v for idx, v in enumerate(params)}
-        try:
-            stacktrace.append(params_dict)
-        except KeyError:
-            # This frame doesn't have the expected format, so skip it and move
-            # on to the next one
-            continue
```
adamchainz, line 81: this didn't make sense - `list.append` cannot raise a `KeyError`

matthiask:
Thanks!

PR #1572, opened 2022-01-12
Remove unnecessary mock dependency

adamchainz:
Presumably left over from Python 2, since Python 3 added mock to the standard library.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,6 @@ envlist =
 deps =
     dj32: django~=3.2.9
     dj40: django~=4.0.0
-    sqlite: mock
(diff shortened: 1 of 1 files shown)


matthiask (approved):
`git log -Smock` doesn't find any uses of `mock` (excluding `unittest.mock`) at all so I suspect that maybe this has never been used at all?

adamchainz:
🤷

matthiask:
Thanks!

tim-schilling:
Nope, I added it for some reason and I can't tell what it was for. Likely a mistake.

PR #1573, opened 2022-01-12
Fix USE_TZ RemovedInDjango50Warning

adamchainz:
Fix this warning, which appears in both the docs build and Django 4.0 tests:

```
/.../python3.8/site-packages/django/conf/__init__.py:199: RemovedInDjango50Warning: The default value of USE_TZ will change from False to True in Django 5.0. Set USE_TZ to False in your project settings if you want to keep the current default behavior.
```


adamchainz:
Okay looks like the test suite needs adjustment...

tests/panels/test_sql.py
```diff
@@ -121,7 +121,13 @@ def test_param_conversion(self):
...
             .filter(group_count__gt=1)
         )
-        list(User.objects.filter(date_joined=datetime.datetime(2017, 12, 22, 16, 7, 1)))
+        list(
+            User.objects.filter(
+                date_joined=datetime.datetime(
+                    2017, 12, 22, 16, 7, 1, tzinfo=datetime.timezone.utc
+                )
+            )
+        )
```
adamchainz, line 130: change from naive datetime, which django will interpret with the default timezone, to aware

tests/test_forms.py
```diff
@@ -1,14 +1,14 @@
...
 from django.test import TestCase
 
 from debug_toolbar.forms import SignedDataForm
 
-SIGNATURE = "v02QBcJplEET6QXHNWejnRcmSENWlw6_RjxLTR7QG9g"
+SIGNATURE = "-WiogJKyy4E8Om00CrFSy0T6XHObwBa6Zb46u-vmeYE"
 
-DATA = {"value": "foo", "date": datetime(2020, 1, 1)}
-SIGNED_DATA = f'{{"date": "2020-01-01 00:00:00", "value": "foo"}}:{SIGNATURE}'
+DATA = {"value": "foo", "date": datetime(2020, 1, 1, tzinfo=timezone.utc)}
```
adamchainz, line 10: Again, this changes the naive datetime to aware, so that it isn't interpreted with the default timezone when converting to str

tests/panels/test_sql.py
```diff
@@ -121,26 +121,41 @@ def test_param_conversion(self):
...
-                tuple(q[1]["params"] for q in self.panel._queries),
-                ('["Foo", true, false]', "[10, 1]", '["2017-12-22 16:07:01"]'),
-            )
+            expected_bools = '["Foo"]'
+
+        if connection.vendor == "postgresql":
+            # PostgreSQL always includes timezone
+            expected_datetime = '["2017-12-22 16:07:01+00:00"]'
+        else:
+            expected_datetime = '["2017-12-22 16:07:01"]'
```
adamchainz, line 149: new formatting difference being triggered on PostgreSQL

matthiask:
Very nice, thanks!

PR #1574, opened 2022-01-12
Optimize SQL reformatting

adamchainz:
Add a few bits of caching:

1. Add sub-function to `parse_sql` with `lru_cache`, so that repeat calls with the same query are fast.
   This saves a lot of processing in N+1 situations.
2. Cache constructed filter stacks in `get_filter_stack()`.
   This avoids recreating all the various sqlparse objects for each query.
3. Pre-compile the simplification regex.
   The `re` module already uses an internal LRU cache of regexes, but this avoids recompiling if the regex ever drops out of that cache.

Building on top of #1571, this takes run time for the same tested view from ~1100ms to ~950ms, another ~15% saving.

### DIFF
--- a/debug_toolbar/panels/sql/utils.py
+++ b/debug_toolbar/panels/sql/utils.py
@@ -1,4 +1,5 @@
 import re
+from functools import lru_cache
 
 import sqlparse
 from django.utils.html import escape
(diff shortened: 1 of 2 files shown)


adamchainz:
After #1571 is merged, I'd like to rebase this and update the release note.

matthiask:
Thanks! That's awesome.

PR #1578, opened 2022-01-17
adding a green flash to new rows in the history table on refresh

gone:
I had this refactor left over from the observe ajax work and didn't want to throw it away

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -607,6 +607,18 @@
 #djDebug .djdt-highlighted {
     background-color: lightgrey;
 }
+@keyframes flash-new {
+    from {
+        background-color: green;
+    }
+    to {
+        background-color: inherit;
+    }
+}
+#djDebug .flash-new {
+    animation: flash-new 1s;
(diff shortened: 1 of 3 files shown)


tim-schilling:
I renamed `array` to `nodes` in `pluckData` since the argument should be a `NodeList` instance and added a docstring to try to clarify that for future developers. I also added a mention of the functionality change to the changes documentation. When CI passes, I'll merge it.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -607,6 +607,18 @@
...
+        background-color: green;
+    }
+    to {
+        background-color: inherit;
+    }
+}
+#djDebug .flash-new {
+    animation: flash-new 1s;
+}
+
```
matthiask, line 621: I think it would be a good idea to namespace this.

```suggestion
@keyframes djdt-flash-new {
    from {
        background-color: green;
    }
    to {
        background-color: inherit;
    }
}
#djDebug .flash-new {
    animation: djdt-flash-new 1s;
}

```

matthiask (approved):
LGTM. Thanks!

PR #1583, opened 2022-01-28
unsigning history forms

gone:
As discussed in #1577  Removing the signing feature of the history forms.

Arguments for the change: It doesn't really provide any security and makes working with them harder.

I can't personally think of a situation where an attacker has access to the debug toolbar + store_id + real database, but as is so often true with security it could be a not obvious exploit - please weigh in if this code is providing value!


tim-schilling (approved):
Thanks for making this a separate PR!

Issue #1581, opened 2022-01-27
Issue importing settings since 3.2.3


jaredahern:
Hello! I have a Django project in which I import the `PANELS_DEFAULTS` into the settings, modify it, and use it for `DEBUG_TOOLBAR_PANELS`. This simplifies the removal of any particular panel, as the defaults don't need to be hard-coded in our settings.

However, as of djdt 3.2.3, this no longer works. The following line in our `settings.py` file causes an `AppRegistryNotReady` error:
```python
from debug_toolbar.settings import PANELS_DEFAULTS
```
If we change it to `import debug_toolbar`, I see an `ImproperlyConfigured` error related to missing `SECRET_KEY`.

I think this is because as of 3.2.3, `debug_toolbar/__init__.py` imports `app_name`, which ultimately imports the whole package. And perhaps that's not good to do that in a `settings.py` file. It might be that we can restructure our `settings.py` to make this work, but at the moment I'm guessing not.

Would it be possible to restructure things such that `debug_toolbar.settings` could be imported without triggering the import of other submodules? Perhaps put `app_name` by itself in a module that itself doesn't import anything? I'm happy to task a stab at this if someone indicates that it would be acceptable.

Or perhaps is there a different approach to "modifying" the default panel settings?

matthiask:
I noticed too that somehow more of the toolbar was imported earlier than previously. 

I think there's no harm duplicating this string in both modules and keeping it synchronized. I submitted a PR.

PR #1588, opened 2022-02-20
Single source 'djdt' app name

living180:
Avoid needing to keep multiple constants in sync by defining an `APP_NAME` constant in `debug_toolbar.__init__` and referencing where it is needed.

### DIFF
--- a/debug_toolbar/__init__.py
+++ b/debug_toolbar/__init__.py
@@ -1,8 +1,10 @@
-__all__ = ["VERSION"]
+__all__ = ["APP_NAME", "VERSION"]
+
+APP_NAME = "djdt"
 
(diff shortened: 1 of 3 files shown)


matthiask:
Hmm. We stopped single-sourcing the value in ff0b974378ecf3e2d6f7f5a404d3c62d0e633912 because of early importing. But yeah, this way around may work. (The old code imported `urls` in `__init__`.)

living180:
Yeah, I hit #1581 as well, and looking at ff0b974, thought it would probably work to flip where the canonical version was stored.  All tests passed for me and it didn't cause the #1581 issue to recur in my case.

matthiask:
Thanks!

PR #1597, opened 2022-03-21
Fix and improve .assertValidHTML() test method

living180:
The `BaseTestCase.assertValidHTML()` helper method was not parsing the passed in `content` argument, but instead `self.panel.content`.  Fix this issue, and also remove the `msg` argument which was never supplied by can callers.  Finally, fix a test case failure exposed by changing the method to parse the supplied `content`.


matthiask:
Thank you!

Issue #1300, opened 2020-07-15
Debug toolbar's ThreadTrackingHandler mutes Daphne errors


JulienPalard:
In a "basic" setup, with `django/channels`, `django/daphne`, and `debug-toolbar` installed, and no specific configuration, Daphne logs are lost. I take daphne as an example, but any application or library not logging under the `django` namespace are lost the same (but could be displayed in the debug toolbar).

This is because we're relying, maybe too much, on the `lastResort` logger to log messages to stderr for any unconfigured logging namespace.

As `debug-toolbar` installs its `ThreadTrackingHandler` to the root logger, messages are considered handled by the `logging` module, which does not use the `lastResort` logger to log them to stdout: messages are not displayed (and while I'm in a websocket handler, I have no debug toolbar to display them to me).

medihack:
I also can confirm this. The issue is resolved by commenting out the `LoggingPanel`.

<pre>
DEBUG_TOOLBAR_PANELS = [
    "debug_toolbar.panels.versions.VersionsPanel",
    "debug_toolbar.panels.timer.TimerPanel",
    "debug_toolbar.panels.settings.SettingsPanel",
    "debug_toolbar.panels.headers.HeadersPanel",
    "debug_toolbar.panels.request.RequestPanel",
    "debug_toolbar.panels.sql.SQLPanel",
    "debug_toolbar.panels.staticfiles.StaticFilesPanel",
    "debug_toolbar.panels.templates.TemplatesPanel",
    "debug_toolbar.panels.cache.CachePanel",
    "debug_toolbar.panels.signals.SignalsPanel",
    # Does not work with Django Channels, see
    # https://github.com/django/channels/issues/1204
    # "debug_toolbar.panels.logging.LoggingPanel",
    "debug_toolbar.panels.redirects.RedirectsPanel",
    "debug_toolbar.panels.profiling.ProfilingPanel",
]
</pre>

Unfortunately, the following does NOT work:

<pre>
DEBUG_TOOLBAR_CONFIG = {
    "DISABLE_PANELS": [
        'debug_toolbar.panels.logging.LoggingPanel',
    ]
}
</pre>

tim-schilling:
I'm closing this as closed in #1603. @medihack @JulienPalard, if that's not the case, please let me know.

matthiask:
Thanks, I agree. Besides, if the bug is still there it would come up again.

Issue #1078, opened 2018-08-17
django-debug-toolbar is suppressing logging


ovalseven8:
Today, I noticed that there is some conflict between `django-debug-toolbar` and `huey`.

In general, when you run `manage.py run_huey` you get immeditately some output like
```
[2018-08-17 17:47:25,040] INFO:huey.consumer:MainThread:Huey consumer started with 1 thread, PID 2389
[2018-08-17 17:47:25,041] INFO:huey.consumer:MainThread:Scheduler runs every 1 second(s).
[2018-08-17 17:47:25,041] INFO:huey.consumer:MainThread:Periodic tasks are enabled.
[2018-08-17 17:47:25,041] INFO:huey.consumer:MainThread:UTC is enabled.
...
```

Now, add this to your settings file:
```
INSTALLED_APPS += [
    'debug_toolbar',
]

MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware', ]
```

And suddenly when you run `manage.py run_huey` again, you do not get any output at all:
```
python manage.py run_huey

```

## System information
Django 2.1
Python 3.7
Huey 1.10.2
Django Debug Toolbar 1.9.1

ovalseven8:
Oh, I've found the reason. Did not have anything to do with Django Debug toolbar, sorry.

ovalseven8:
Turns out there is an issue already: https://github.com/jazzband/django-debug-toolbar/issues/695#issuecomment-414113249

Issue #695, opened 2015-03-05
Enabling debug_toolbar seems to break logging from management commands


cancan101:
When I add `debug_toolbar` to my `INSTALLED_APPS`, calls to `logging` `logger` seems no longer work from within a management command.

tim-schilling:
Can you please provide us with a test case?

cancan101:
https://www.dropbox.com/s/oifjoqqxsrizzew/debug_toolbar_issue.zip?dl=0
the command is `check_log`. You can see that toggling the `debug_toolbar` app in the settings.py file changes the log behavior.

aaugustin:
This may be related to 7d29b40c70dbc1f0996b9342254f32705eba1b67.

gfairchild:
I'm not sure if this is related to this issue (I suspect it is), but logging seems completely broken for me with DDT 1.3.0. I have numerous `logging.debug` calls scattered throughout my code, and DDT is no longer picking them up. I reverted to 1.2.2, and logging resumes working.

I'm using Python 3.4.x and Django 1.7.7.

If you want me to create a new issue, let me know.

tim-schilling:
@gfairchild Do you know what your `root.logging.level` is set to?

gfairchild:
@tim-schilling, I'm sorry, but I googled around and can't find any references to `root.logging.level`. What is that, and how do I report my value to you?

gfairchild:
Actually, it looks like `logging.getLogger().getEffectiveLevel()` gets that value. I missed a SO thread on it. That function returns 0, which tells me my logger's root logging level is `NOTSET`.

tim-schilling:
Sorry about that. I meant ```logging.root.level```. I typed it out wrong. I'll try to recreate your scenario.

gfairchild:
I just updated to Django 1.8, and logging still isn't working, so I'm guessing this is a DDT bug.

dkrnl:
Yes, i catch similar bug: "No messages logged" in logging panel.
After update from Django 1.7 (logging working well) to 1.8 version.

tim-schilling:
@gfairchild @dkrnl Can you try adding the following code to your settings file?

```
import logging
logging.root.setLevel(logging.NOTSET)
```

gfairchild:
That seems to have fixed the issue!

tim-schilling:
Glad it worked for you! That problem is actually different than the
original issue which I'm still looking into.

[...]

PR #1603, opened 2022-04-02
Preserve logs that LoggingPanel would previously overwrite

riwatt:
This might solve the root cause of issues reported in https://github.com/jazzband/django-debug-toolbar/issues/695 https://github.com/jazzband/django-debug-toolbar/issues/1078 and https://github.com/jazzband/django-debug-toolbar/issues/1300

-----

Summary:
LoggingPanel upon its import, changes the Python interpreter's root logger to log into a handler that collects its records for the LoggingPanel's display. However for users that don't explicitly configure the root logger this looks like the Debug Toolbar suppresses their logs that previously went to standard error.
With this commit, the logs will still keep going to standard error even if the LoggingPanel is installed, but only if the root logger has not been explicitly configured with a handler yet.
This behavior will make it so that installing DDT won't break logging neither for users that already have logging customizations nor for users that have an out-of-the-box django default logging setup.

Also: change LOGGING config in tests/settings.py because now DDT's own test suite would show more logs without an explicit config.

### DIFF
--- a/debug_toolbar/panels/logging.py
+++ b/debug_toolbar/panels/logging.py
@@ -45,6 +45,24 @@ class ThreadTrackingHandler(logging.Handler):
         self.collector.collect(record)
 
 
+# Preserve Python's fallback log mechanism before adding ThreadTrackingHandler.
+
+# If the root logger has no handlers attached then everything that reaches it goes
+# to the "handler of last resort".
+# So a Django app that doesn't explicitly configure the root logger actually logs
+# through logging.lastResort.
+# However, logging.lastResort is not used after ThreadTrackingHandler gets added to
+# the root logger below. This means that users who have LoggingPanel enabled might
+# find their logs are gone from their app as soon as they install DDT.
+# Explicitly adding logging.lastResort to logging.root's handler sidesteps this
+# potential confusion.
+# Note that if root has already been configured, or logging.lastResort has been
+# removed, then the configuration is unchanged, so users who configured their
+# logging aren't exposed to the opposite confusion of seeing extra log lines from
+# their app.
+if not logging.root.hasHandlers() and logging.lastResort is not None:
+    logging.root.addHandler(logging.lastResort)
+
 # We don't use enable/disable_instrumentation because logging is global.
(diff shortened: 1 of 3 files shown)


matthiask (approved):
Thanks! Looks good.

I was wondering for a moment if it would be possible to use `lastResort` for the toolbar but that would probably be more brittle and less obvious.

I'm always worrying a bit about module-level configuration because of the dependency on import ordering. This will be fine though since Django initializes logging long before the toolbar is initialized for the first time in the middleware.

tim-schilling:
Is it possible to test this in the test suite? Otherwise this looks good, but needs verification.

riwatt:
> Is it possible to test this in the test suite? Otherwise this looks good, but needs verification.

Sorry I didn't mention it, but I verified this with a minimal django app by varying its logging configs.
Would it be sufficient if I describe that repro here?
Or maybe I can add something to the example app that can show before and after behavior?

Fully covering this within the test suite might be difficult because of the above mentioned concerns of the original code overriding global behavior at import time. I will look into it though, but let me know if repro is enough.
Thanks for the quick reviews!

matthiask:
I agree with the sentiment that adding a unittest which fully covers this may be hard to do. Maybe it's feasible to add a regression test which emits a logging message without any configured loggers and which verifies that the output appears on `sys.stderr`? (maybe mock `stderr` if necessary) I think this should cover the expected default behavior and would be sufficient.

If that proves hard to do then I think a description would be fine too. We're all volunteering here :)

riwatt:
@matthiask I rebased this branch so its first commit is now a regression test.
If you run the testsuite against that commit only then it will fail, but then you can see that the final commit fixes it.
I also added my original reproducer app into this repo: https://github.com/riwatt/ddjt_logging. The readme there describes the steps I used to test the fix.

----

Re: the precommit.ci failure is because of https://github.com/psf/black/issues/2964
I ran it locally with black==22.3.0 without any issues.

matthiask:
Thanks! Let's merge this, everything makes a lot of sense to me.

PR #1605, opened 2022-04-19
Reduce SQL panel overhead

living180:
The existing `SQLPanel.generate_stats()` method can be rather expensive if there are a lot of SQL queries.  This PR defers that expense by moving that code to a new `generate_context()` method, which only gets called if the panel's `.content` attribute is accessed.  Thus requests in which the user does not need to see the SQL Panel content get faster.

It also moves some code from the `NormalCursorWrapper._record()` function to the `SQLPanel.generate_context()` method to reduce the runtime overhead of recording SQL queries.

### DIFF
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -289,7 +289,5 @@ class SQLPanel(Panel):
         )
 
     def generate_server_timing(self, request, response):
-        stats = self.get_stats()
-        title = "SQL {} queries".format(len(stats.get("queries", [])))
-        value = stats.get("sql_time", 0)
-        self.record_server_timing("sql_time", title, value)
+        title = f"SQL {len(self._queries)} queries"
+        self.record_server_timing("sql_time", title, self._sql_time)
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -1,3 +1,4 @@
+import json
 import uuid
 from collections import defaultdict
 from copy import copy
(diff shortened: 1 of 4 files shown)


tim-schilling:
>  Thus requests in which the user does not need to see the SQL Panel content get faster.

How do we determine which requests the user doesn't need to see the SQL Panel content for?

living180:
> How do we determine which requests the user doesn't need to see the SQL Panel content for?

We don't have to.  All the information is still there, but the heavy computations are deferred until the the first time the render_panel() view is called, which actually displays the panel content.

tim-schilling:
There are a lot of good things in this PR that aren't all related to the performance improvements. I'd be happy to merge those in separate PRs.

Regarding the performance improvements: which part(s) of `generate_stats` are expensive? Can we move those out to something that occurs on render? The concern is that the SQLPanel won't have the same flow as the other panels. This makes maintenance a bit more cumbersome and would impact any third party panels that inherit from `SQLPanel`. I'm currently a 0 on this. I'd be a 1 if we stuck to `record_stats` storing the information that's then used by the overridden `content` property / a new render function which would do the expensive computations.

@matthiask what are your thoughts?

living180:
Thanks for the feedback.  I've had some thoughts on the design so I'm planning to rework this PR a little bit and then hopefully post an updated version for review this week.

Regarding the overall performance of `.generate_stats()`, I have not done any profiling to see what parts are expensive, but was rather trying to defer the cost until it was actually needed.  But trying to improve the overall performance is probably something worth looking into in addition to deferring unnecessary work until rendering the content.  I do know that prettifying the SQL can be expensive, especially for queries with large numbers of terms.  I have posted PRs to fix a couple of instances of quadratic behavior in sqlparse, but they have not yet been merged.  I suspect that sqlparse isn't the only contributor to the rendering cost, but without doing some profiling work I cannot say definitively.

matthiask:
Superseeded by changes which have been merged in the meantime. Thanks!

PR #1610, opened 2022-04-28
Version 3.3.0

tim-schilling:
* Track calls to :py:meth:`django.core.caches.cache.get_or_set`.
* Removed support for Django < 3.2.
* Updated check ``W006`` to look for
  ``django.template.loaders.app_directories.Loader``.
* Reset settings when overridden in tests. Packages or projects using
  django-debug-toolbar can now use Django’s test settings tools, like
  ``@override_settings``, to reconfigure the toolbar during tests.
* Optimize rendering of SQL panel, saving about 30% of its run time.
* New records in history panel will flash green.
* Automatically update History panel on AJAX requests from client.

First time contributors:

- bellini666
- riwatt

Contributors to the release:

- adamchainz
- living180
- gone
- matthiask
- francoisfreitag
- pauloxnet

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,7 +44,7 @@ Here's a screenshot of the toolbar in action:
 In addition to the built-in panels, a number of third-party panels are
 contributed by the community.
 
-The current stable version of the Debug Toolbar is 3.2.4. It works on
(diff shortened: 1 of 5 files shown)


matthiask:
Thanks!

PR #1611, opened 2022-05-01
Update to 'debug_toolbar.middleware.show_toolbar'

richard-scott:
I seem to remember reading somewhere that you wanted to be able to check if the REMOTE_ADDR was in a network range... I needed that too, so coded this up. The 'ipaddress' module is a standard module with Python3 and was introduced in v3.3.

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -4,6 +4,7 @@ Debug Toolbar middleware
 
 import re
 from functools import lru_cache
+from ipaddress import IPv4Address, IPv4Network, AddressValueError
 
 from django.conf import settings
 from django.utils.module_loading import import_string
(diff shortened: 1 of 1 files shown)


tim-schilling:
I'm on board with this, but I'd prefer to see the convenience methods used to support IPV6 as well. I'm wondering if we do that if we should add a system check that `INTERNAL_IPS` is not valid so we can avoid the try except in a loop. If we do that we could get to:

```python
# Short circuit on debug
if not settings.DEBUG:
    return False
# Check IP is in range
try:
    ip = ipaddress.ip_address(request.META.get('REMOTE_ADDR'))
except ValueError:
    ip = None
return ip and any(ip in ipaddress.ip_network(internal_ip) for internal_ip in settings.INTERNAL_IPS)
```

PR #1619, opened 2022-05-10
Get PostgreSQL transaction tracking working

living180:
There were bugs at multiple levels that prevented PostgreSQL transactions from being tracked properly.  This PR hopefully fixes all of those, plus adds a test for this functionality.

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -158,7 +158,7 @@ class NormalCursorWrapper(BaseCursorWrapper):
 
             alias = self.db.alias
             conn = self.db.connection
-            vendor = getattr(conn, "vendor", "unknown")
+            vendor = self.db.vendor
 
             # Sql might be an object (such as psycopg Composed).
             # For logging purposes, make sure it's str.
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -61,6 +61,7 @@ def wrap_cursor(connection, panel):
 def unwrap_cursor(connection):
     if hasattr(connection, "_djdt_cursor"):
         del connection._djdt_cursor
+        del connection._djdt_chunked_cursor
         del connection.cursor
         del connection.chunked_cursor
 
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -156,7 +156,7 @@ class NormalCursorWrapper(BaseCursorWrapper):
                 pass  # object not JSON serializable
             template_info = get_template_info()
 
-            alias = getattr(self.db, "alias", "default")
+            alias = self.db.alias
             conn = self.db.connection
             vendor = getattr(conn, "vendor", "unknown")
 
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -60,10 +60,22 @@ def wrap_cursor(connection, panel):
 
 def unwrap_cursor(connection):
     if hasattr(connection, "_djdt_cursor"):
+        # Sometimes the cursor()/chunked_cursor() methods of the DatabaseWrapper
+        # instance are already monkey patched before wrap_cursor() is called.  (In
+        # particular, Django's SimpleTestCase monkey patches those methods for any
+        # disallowed databases to raise an exception if they are accessed.)  Thus only
+        # delete our monkey patch if the method we saved is the same as the class
+        # method.  Otherwise, restore the prior monkey patch from our saved method.
+        if connection._djdt_cursor == connection.__class__.cursor:
+            del connection.cursor
+        else:
+            connection.cursor = connection._djdt_cursor
         del connection._djdt_cursor
+        if connection._djdt_chunked_cursor == connection.__class__.chunked_cursor:
+            del connection.chunked_cursor
+        else:
+            connection.chunked_cursor = connection._djdt_chunked_cursor
         del connection._djdt_chunked_cursor
-        del connection.cursor
-        del connection.chunked_cursor
 
 
 class BaseCursorWrapper:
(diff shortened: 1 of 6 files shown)


debug_toolbar/panels/sql/tracking.py
```diff
@@ -87,10 +100,10 @@ class NormalCursorWrapper(BaseCursorWrapper):
     Wraps a cursor and logs queries.
     """
 
-    def __init__(self, cursor, db, logger):
+    def __init__(self, cursor, connection, logger):
         self.cursor = cursor
         # Instance of a BaseDatabaseWrapper subclass
-        self.db = db
+        self.connection = connection
```
tim-schilling, line 106: Please keep the name `db` rather than `connection`. It should match the [cursor signature for Django](https://github.com/django/django/blob/main/django/db/backends/utils.py#L15)

debug_toolbar/panels/sql/tracking.py
```diff
@@ -60,10 +60,16 @@ def chunked_cursor(*args, **kwargs):
...
     if hasattr(connection, "_djdt_cursor"):
+        if connection._djdt_cursor == connection.__class__.cursor:
+            del connection.cursor
+        else:
+            connection.cursor = connection._djdt_cursor
         del connection._djdt_cursor
+        if connection._djdt_chunked_cursor == connection.__class__.chunked_cursor:
+            del connection.chunked_cursor
+        else:
+            connection.chunked_cursor = connection._djdt_chunked_cursor
```
tim-schilling, line 77: How can we test that this is working as expected?

Additionally, adding summarizing the commit message would be helpful to future developers.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -213,7 +213,6 @@ def _record(self, method, sql, params):
                         "trans_id": trans_id,
                         "trans_status": final_trans_status,
                         "iso_level": iso_level,
-                        "encoding": conn.encoding,
```
tim-schilling, line 185: @matthiask are we going to count this as a backwards incompatibility for any panel that might inherit from `SQLPanel`? It's been around since Pre v1.0, but I don't know of many things that inherit from `SQLPanel` or would use this.

tim-schilling (changes requested):
Thank you for the well written pull request. There's a lot of good information in the individual commits. I think summarizing some of that into comments would help future devs understand what things are doing and why, but I'm content without them for the most part.

debug_toolbar/panels/sql/tracking.py
living180: I don't think there is a backwards compatibility issue here since none of these attributes were ever being recorded, since prior to this PR the code was looking up the `vendor` attribute on the underlying DB connection rather than Django's `DatabaseWrapper`.  Thus none of the PostgreSQL-specific code was ever being run.

debug_toolbar/panels/sql/tracking.py
living180: Well, I had to implement it to get the multi-DB tests working :slightly_smiling_face:.  Not sure if it needs more testing than that...

debug_toolbar/panels/sql/tracking.py
living180: Sounds good.  Dropped that entire renaming commit.

tests/panels/test_sql.py
```diff
@@ -506,3 +506,97 @@ def test_nested_template_information(self):
...
+
+        query = self.panel._queries[0]
+        self.assertEqual(query[0], "default")
+
+        query = self.panel._queries[-1]
+        self.assertEqual(query[0], "replica")
+
+    @unittest.skipUnless(
+        connection.vendor == "postgresql", "Test valid only on PostgreSQL"
+    )
```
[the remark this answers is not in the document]
living180: So I changed the test to run on all databases, and for anything other than PostgreSQL, it ensures that no transaction information was recorded.

tests/panels/test_sql.py
```diff
@@ -506,3 +506,97 @@ def test_nested_template_information(self):
...
+
+        response = self.panel.process_request(self.request)
+        self.panel.generate_stats(self.request, response)
+
+        self.assertEqual(len(self.panel._queries), 6)
+
+        query = self.panel._queries[0]
+        self.assertEqual(query[0], "default")
+        self.assertIsNotNone(query[1]["trans_id"])
+        self.assertTrue(query[1].get("starts_trans", False))
```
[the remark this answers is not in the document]
living180: Updated the test to avoid the use of `.get()` in favor of explicit indexing or `in` tests.

living180:
Thanks so much for the review.  I hope I have adequately addressed the points you raised.  I added comments to the parts that I felt were non-obvious from the code.  If there is anything still unclear, either from the comments that I added or that I have not yet commented, please let me know.

debug_toolbar/panels/sql/tracking.py
tim-schilling: That's fair, thank you!

debug_toolbar/panels/sql/tracking.py
matthiask: I'm fine with this change. The variable is completely unused inside django-debug-toolbar. The variable has been added in 72271115974463f30a2f9f8755e42acc69e204c3 11 years ago but I'm not sure it has ever been used anywhere.

matthiask (approved):
It's a lot to review. I think the code looks good (better than before), tests pass and I've tested it a bit in a local environment. 

I am fine with merging this but I'll wait for a moment to see if @tim-schilling has some reservations.

debug_toolbar/panels/sql/tracking.py
tim-schilling: That works for me. I wanted to make sure we made an active decision rather than letting it go silently. Thank you both!

tim-schilling (approved):
Thank you for the additional comments. The code makes more sense to me and is in a much better spot. This is excellent work!

living180:
Thanks @matthiask and @tim-schilling for the reviews!  One question for future reference - would it be helpful if I included relevant change log entries with future PRs?

matthiask:
> Thanks @matthiask and @tim-schilling for the reviews! One question for future reference - would it be helpful if I included relevant change log entries with future PRs?

Sure! This would be helpful.

PR #1622, opened 2022-05-18
Remove a couple of archived third-party repos

cclauss:


matthiask:
Thank you! I think it would be even better to remove those two panels; django-haystack-panel had its last commit in 2014, djdt-flamegraph doesn't look like it's maintained to actively either https://github.com/blopker/djdt-flamegraph/commits/master

matthiask:
Thanks!

PR #1624, opened 2022-05-18
Fix #1621: Do not crash when encountering unexpected data in the request

### LINKED ISSUE #1621: Middleware breaks raising APIException from ViewSet
When raising a subclass of `APIException` (for example `ValidationError` or `ParseError`) from a `ViewSet`, I get a very cryptic error message that suggests that some sort function is broken. When I disable the middleware, everything works as expected.

Minimal code example:

```
# views.py
class MyViewSet(viewsets.GenericViewSet):
    ...

    @action(methods=['post'], detail=False, url_path='bulk-create')
    def bulk_create(self, request, *args, **kwargs):
        # Some assertion about data structure        
        if True:
            raise ParseError("Expected the data to be in list format.")

        # Actually process incoming data
        ...
```

Error log:

```
2022-05-17T12:23:45.366216553Z   File "/opt/venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
2022-05-17T12:23:45.366221178Z     response = get_response(request)
2022-05-17T12:23:45.366223345Z   File "/opt/venv/lib/python3.9/site-packages/debug_toolbar/middleware.py", line 67, in __call__
2022-05-17T12:23:45.366225470Z     panel.generate_stats(request, response)
2022-05-17T12:23:45.366238220Z   File "/opt/venv/lib/python3.9/site-packages/debug_toolbar/panels/request.py", line 30, in generate_stats
2022-05-17T12:23:45.366240470Z     "post": get_sorted_request_variable(request.POST),
2022-05-17T12:23:45.366346803Z   File "/opt/venv/lib/python3.9/site-packages/debug_toolbar/utils.py", line 227, in get_sorted_request_variable
2022-05-17T12:23:45.366352511Z     return [(k, variable.getlist(k)) for k in sorted(variable)]
2022-05-17T12:23:45.366354803Z TypeError: '<' not supported between instances of 'dict' and 'dict'
```

I have also posted this question on StackOverflow: https://stackoverflow.com/questions/72274329/cannot-raise-apiexception-from-django-viewset

Thank you for your help in advance.

matthiask:


debug_toolbar/utils.py
```diff
@@ -229,14 +229,17 @@ def getframeinfo(frame, context=1):
...
-        return [(k, variable.get(k)) for k in sorted(variable)]
-    else:
-        return [(k, variable.getlist(k)) for k in sorted(variable)]
+    try:
+        if isinstance(variable, dict):
+            return [(k, variable.get(k)) for k in sorted(variable)]
+        else:
+            return [(k, variable.getlist(k)) for k in sorted(variable)]
+    except TypeError:
+        return [(f"<{source}>", variable)]
```
tim-schilling, line 242: Since the wrapping call already has this associated with the attribute on the request (get, post, files), I don't think we need to use the source. I think using an empty string or something constant will be more appropriate.

matthiask: Thanks, that's a good point. I have pushed an update but I haven't tested the update (yet)

matthiask:
image

Here's how this could look. I broke `get_sorted_request_variable` on purpose to show cookies and query parameters as a raw block instead of as a table for the screenshot, but the code should still behave the same except if encountering some malformed data in request variables.

matthiask:
@FinnGu I'd appreciate your input and/or testing if you have a way to do this.

FinnGu:
> @FinnGu I'd appreciate your input and/or testing if you have a way to do this.

Alright, setting up some tests on my end.

FinnGu:
@matthiask I just recently started using the DDT, so I don't know about all of its features. But for my use case, everything works flawlessly! I am now able to raise `ParseError`s, `ValidationError`s and basically all other exceptions that I am used to, without the DDT interfering with DRF ❤️

matthiask:
Thanks! Good to hear.

PR #1625, opened 2022-05-18
Replace OrderedDict

living180:
Replace most usages of `OrderedDict` with plain Python dicts since as of Python 3.7, dicts are guaranteed to maintain insertion order.  Also make a couple of small improvements related to sorting dicts.

### DIFF
--- a/debug_toolbar/forms.py
+++ b/debug_toolbar/forms.py
@@ -47,7 +47,6 @@ class SignedDataForm(forms.Form):
 
     @classmethod
     def sign(cls, data):
-        items = sorted(data.items(), key=lambda item: item[0])
         return signing.Signer(salt=cls.salt).sign(
-            json.dumps({key: force_str(value) for key, value in items})
+            json.dumps({key: force_str(value) for key, value in data.items()})
         )
--- a/tests/test_forms.py
+++ b/tests/test_forms.py
@@ -7,7 +7,7 @@ from debug_toolbar.forms import SignedDataForm
 
 SIGNATURE = "-WiogJKyy4E8Om00CrFSy0T6XHObwBa6Zb46u-vmeYE"
 
-DATA = {"value": "foo", "date": datetime(2020, 1, 1, tzinfo=timezone.utc)}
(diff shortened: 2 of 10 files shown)


debug_toolbar/forms.py
```diff
@@ -47,7 +47,6 @@ def verified_data(self):
 
     @classmethod
     def sign(cls, data):
-        items = sorted(data.items(), key=lambda item: item[0])
         return signing.Signer(salt=cls.salt).sign(
-            json.dumps({key: force_str(value) for key, value in items})
+            json.dumps({key: force_str(value) for key, value in data.items()})
```
matthiask, line 51: I was a bit worried about this. I thought that maybe the string was hashed and that changing the order of keys may lead to invalid hashes but this doesn't seem to be the case since we're not hashing, only signing and unsigning.

tests/test_forms.py
```diff
@@ -5,10 +5,10 @@
 
 from debug_toolbar.forms import SignedDataForm
 
-SIGNATURE = "-WiogJKyy4E8Om00CrFSy0T6XHObwBa6Zb46u-vmeYE"
+SIGNATURE = "O7cv5oYzP3ZgjSBrJwdidTUQ5kBWI7oD1xn8fb2vK-w"
 
 DATA = {"value": "foo", "date": datetime(2020, 1, 1, tzinfo=timezone.utc)}
-SIGNED_DATA = f'{{"date": "2020-01-01 00:00:00+00:00", "value": "foo"}}:{SIGNATURE}'
+SIGNED_DATA = f'{{"value": "foo", "date": "2020-01-01 00:00:00+00:00"}}:{SIGNATURE}'
```
matthiask, line 11: Detail: I'd prefer it if you reversed the order of `value` and `date` in the dictionary below; this would neatly show that the signature stays the same.

matthiask (approved):

tests/test_forms.py
living180: Done.

matthiask:
Thanks!

PR #1626, opened 2022-05-23
Cache panel work

living180:
This PR re-architects `CachePanel` call recording.  Using signals for monitoring cache calls gives incorrect results in the face of concurrency.  If two requests are being processed concurrently in different threads, they will store each other's cache calls because both panels will be subscribed to the same signal.  This PR updates the code to record the cache calls directly via a method on the panel instance rather than going through signals.

Additionally, it reworks the `enable_instrumentation()` mechanism to monkey patch methods on the cache instances directly instead of replacing the cache instances with wrapper classes.  This should eliminate the corner cases mentioned in the (now-removed) `disable_instrumentation()` comments.

One change I made as part of the preparatory work for these changes is adding a `ready()` class method to the panel API, to be called from the `DebugToolbarConfig.ready()` method.  This provides a clean way for panels to perform any initialization or instrumentation that needs to be done unconditionally for the panel regardless of whether it is enabled for a particular request.  This in theory could cause backwards compatibility issues if a third-party panel already had a `ready` method or attribute.  I did check all the third-party panels listed in the Debug Toolbar documentation, and none of them would be affected by this change.  However, if there are still backwards compatibility concerns about the addition of this method, I can rework this part of the PR.

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -17,9 +17,10 @@ class DebugToolbarConfig(AppConfig):
     def ready(self):
         from debug_toolbar.toolbar import DebugToolbar
 
-        # Import the panels when the app is ready. This allows panels
-        # like CachePanel to enable the instrumentation immediately.
-        DebugToolbar.get_panel_classes()
+        # Import the panels when the app is ready and call their ready() methods.  This
+        # allows panels like CachePanel to enable their instrumentation immediately.
+        for cls in DebugToolbar.get_panel_classes():
+            cls.ready()
 
 
 def check_template_config(config):
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ classifiers =
 [options]
 python_requires = >=3.7
 install_requires =
+    asgiref >= 3.3.2, < 4
     Django >= 3.2
     sqlparse >= 0.2.0
 packages = find:
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -114,6 +114,19 @@ class Panel:
         """
         return []
 
+    # Panel early initialization
+
+    @classmethod
+    def ready(cls):
+        """
+        Perform early initialization for the panel.
+
+        This should only include initialization or instrumentation that needs to
+        be done unconditionally for the panel regardless of whether it is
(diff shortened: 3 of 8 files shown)


living180:
FYI, I think this should also fix #799 (though that was not my goal when I started working on these changes).

debug_toolbar/apps.py
```diff
@@ -17,9 +17,10 @@ class DebugToolbarConfig(AppConfig):
     def ready(self):
         from debug_toolbar.toolbar import DebugToolbar
 
-        # Import the panels when the app is ready. This allows panels
-        # like CachePanel to enable the instrumentation immediately.
-        DebugToolbar.get_panel_classes()
+        # Import the panels when the app is ready and call their ready() methods.  This
+        # alows panels like CachePanel to enable their instrumentation immediately.
```
matthiask, line 21:
```suggestion
        # allows panels like CachePanel to enable their instrumentation immediately.
```

setup.cfg
```diff
@@ -31,6 +31,7 @@ classifiers =
 [options]
 python_requires = >=3.7
 install_requires =
+    asgiref >= 3.4.1
```
matthiask, line 34: Would `asgiref >= 3.3.2` be fine as well or do we need the newer version? Django 3.2.x specifies `asgiref >= 3.3.2 <4` in `setup.cfg` and I'd prefer to reuse this.

matthiask (approved):
I like it, thanks!

debug_toolbar/signals.py
```diff
@@ -0,0 +1,4 @@
+from django.dispatch import Signal
+
+# Signal used internally for testing
+toolbar_created = Signal()
```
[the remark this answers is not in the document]
living180: I've made it a class attribute on the `DebugToolbar` class under the name `_created`.  Hopefully that will lower its profile?

setup.cfg
living180: Either should work fine I think - I just had to pick one and went with what the latest Django used.  Updated to the same as Django 3.2.x.

debug_toolbar/apps.py
living180: Good catch, thanks!

matthiask:
Thank you!

PR #1629, opened 2022-05-24
Fix cache panel miss counting

living180:
The cache panel was not counting misses for the `get_many()` cache method, because it assumed that all keys would be present in the returned dict (with a value of `None` if not present in the cache) while in reality only keys present in the cache are present in the returned dict.  Correct the miss counting logic, and add a test for tracking hits and misses.


tim-schilling (approved):
Makes sense

tim-schilling:
@living180 looks like using github's conflict resolver didn't go well. I'll get that cleaned up.

PR #1630, opened 2022-05-29
Stacktrace cleanups

living180:
Remove some obsolete compatibility code from the stack trace recording code.

### DIFF
--- a/debug_toolbar/utils.py
+++ b/debug_toolbar/utils.py
@@ -47,12 +47,11 @@ def omit_path(path):
 
 def tidy_stacktrace(stack):
     """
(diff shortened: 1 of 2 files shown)


matthiask:
Thank you! Makes sense.

I haven't researched if the `coding=` line is still supported in Python 3 but even if it were it surely wouldn't be widely used.

living180:
> I haven't researched if the `coding=` line is still supported in Python 3 but even if it were it surely wouldn't be widely used.

From the documentation, it looks like it still is:  https://docs.python.org/3.10/howto/unicode.html#unicode-literals-in-python-source-code

PR #1631, opened 2022-05-29
Stack trace rework

living180:
Add a new stack recording mechanism which includes caching on a per-request basis to reduce expensive file system operations.  Update the cache and SQL panels to use this new mechanism to record stack traces.

The caching is implemented on top of `asgiref.local.Local` to ensure that requests being handled in different thread/async task contexts have independent caches.

This PR supersedes #1606.

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -10,6 +10,7 @@ from django.utils.module_loading import import_string
 
 from debug_toolbar import settings as dt_settings
 from debug_toolbar.toolbar import DebugToolbar
+from debug_toolbar.utils import clear_stack_trace_caches
 
 _HTML_TYPES = ("text/html", "application/xhtml+xml")
 
@@ -56,6 +57,7 @@ class DebugToolbarMiddleware:
             # Run panels like Django middleware.
             response = toolbar.process_request(request)
         finally:
+            clear_stack_trace_caches()
             # Deactivate instrumentation ie. monkey-unpatch. This must run
             # regardless of the response. Keep 'return' clauses below.
             for panel in reversed(toolbar.enabled_panels):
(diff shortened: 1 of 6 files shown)


matthiask (approved):
The code makes a lot of sense to me. Thanks! 

This should make it possible to delete the old `get_stack`, `tidy_stacktrace` functionality? I don't think we should treat those utilities as API of django-debug-toolbar (which would mean that we'd have to deprecate them and keep them around for some time) but we should instead delete them here and now. This would improve the diffstat, too -- it shows mostly added lines until now.

I haven't checked out the code locally, the approval is based on reading the code and a passing testsuite.

living180:
> This should make it possible to delete the old `get_stack`, `tidy_stacktrace` functionality?

I'm a -1 on deleting them.  I know for certain that there are multiple third-party panels that use them, even though they aren't part of the documented API, and deleting them would break those panels.  I'm happy to add deprecation warnings to them, however.

living180:
And then a third-party panel that wanted to support multiple debug toolbar versions without triggering deprecation warnings could use a snippet like the following:

```
try:
    from debug_toolbar.utils import get_stack_trace
except ImportError:
    from debug_toolbar.utils import get_stack, tidy_stacktrace

    def get_stack_trace():
        return tidy_stacktrace(reversed(get_stack))
```

matthiask:
I see.

The problem is that these functions do not seem to be called in any tests anymore and breakage/bit rot will happen sooner rather than later. Therefore we should definitely deprecate those functions.

I think it would be nice to add the snippet to the release notes and warn everyone that the old functionality will be removed in the release after the next. If you know of third party panels using these functions giving them a heads up would be extra nice. They could even help testing the relevant code :)

living180:
Just pushed an update that adds deprecation warnings to `get_stack()` and `tidy_stacktrace()`, and also adds a test to ensure they get run.  Also added a note about the deprecation of these functions to the change log.

matthiask:
Thanks! Let's move ahead with this.

PR #1635, opened 2022-05-31
Reimplement HIDE_IN_STACKTRACES machinery

living180:
The pre-existing `HIDE_IN_STACKTRACES` machinery was based on file system paths, but was rather inefficient because it required a file system operation (`os.path.realpath()`) for each stack frame.  In addition, it would not work properly for namespace packages since they can have multiple file system hierarchies.

The new implementation avoids using paths altogether and instead uses the `__name__` variable found in each frame's `f_globals` attribute to determine the associated module name, and compares that directly with the list of module names provided in the `HIDE_IN_STACKTRACES` setting.

### DIFF
--- a/debug_toolbar/utils.py
+++ b/debug_toolbar/utils.py
@@ -3,11 +3,9 @@ import linecache
 import os.path
 import sys
 import warnings
-from importlib import import_module
 from pprint import pformat
 
 from asgiref.local import Local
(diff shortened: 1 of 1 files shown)


matthiask (approved):
Thanks! Looks good. 

I think I'd prefer both `get_stack_trace` methods to use keyword-only arguments at that point in time. It should make future refactorings easier, I hope.

PR #1636, opened 2022-05-31
Rework similary/duplicate query grouping

living180:
The existing code for tracking and grouping similar and duplicate queries was complicated, hard to understand (nested dict comprehensions), and inefficient (multiple nested loops, using exceptions for flow control, calling key-computation functions twice for each query).  Tweak the data structures to allow significant simplification of the code, and pull duplicated functionality into a shared function.

### DIFF
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -48,6 +48,18 @@ def get_transaction_status_display(vendor, level):
     return choices.get(level)
 
 
+def _similar_query_key(query):
+    return query["raw_sql"]
+
+
+def _duplicate_query_key(query):
+    raw_params = () if query["raw_params"] is None else tuple(query["raw_params"])
+    # saferepr() avoids problems because of unhashable types
+    # (e.g. lists) when used as dictionary keys.
+    # https://github.com/jazzband/django-debug-toolbar/issues/1091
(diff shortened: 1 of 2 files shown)


matthiask:
I'm surprised. I thought there was a good reason for using the `pprint` module, because of unhashable types and other problems with data structures (I could be wrong, it has been a long time). LGTM.

debug_toolbar/panels/sql/panel.py
```diff
@@ -48,6 +47,33 @@ def get_transaction_status_display(vendor, level):
...
+    # https://github.com/jazzband/django-debug-toolbar/issues/1091
+    return (query["raw_sql"], repr(raw_params))
+
+
+def _process_query_groups(query_groups, databases, colors, name):
+    counts = defaultdict(int)
+    for (alias, key), query_group in query_groups.items():
+        count = len(query_group)
+        # Queries are similar / duplicates only if there are at least 2 of them.
+        if count > 2:
```
slyapustin, line 67: @living180 Should it be 
```python
if count > 1:
```
as per comment:
> if there are at least 2 of them.

living180: Good catch! My refactor accidentally changed a `>=` to a `>`.  Fixed, and added some tests which would catch this.

PR #1644, opened 2022-06-23
django-debug-toolbar 3.5

matthiask:
I think it's time to release 3.5; no major version bump despite all these changes because I'm somewhat sure that we didn't break anything.


tim-schilling:
Thanks @matthiask for releasing this. And big thank you @living180 for the major contributions in this release!

PR #1655, opened 2022-08-02
Make Selenium tests pass, hopefully

matthiask:


tim-schilling:
Looks good. I would have removed the old version of the tests in tox using selenium, but not a big deal.

matthiask:
Thanks! I thought it's fine to run selenium tests with two Django versions but you make me reconsider since we're not testing Django but django-debug-toolbar.

PR #1657, opened 2022-08-03
Fixed signature of views decorated with signed_data_view

spookylukey:
This fixes a nuisance when using django-debug-toolbar with [django-urlconfchecks](https://github.com/AliSayyah/django-urlconfchecks/) which type checks URLconfs, and possibly other tools. Otherwise it has little effect.

Basically, the `@functools.wraps()` call within the `signed_data_view` decorator makes views like `sql_explain` appear to have an external signature of `sql_explain(request, verified_data)` to tools like [inspect.signature](https://docs.python.org/3/library/inspect.html#inspect.signature). If this was really the case, then the function would fail when used from the urlconf, because the `verified_data` parameter is not passed. 

So, django-urlconfchecks raises an error for these views. 

In reality `signed_data_view` is providing that argument, and so changing the effective signature of the function.

With the patch, the signature is changed to `sql_explain(request, *args, **kwargs)`, which is not as accurate as it could be, but at least django-urlconfchecks now just issues a warning (about `*args` / `**kwargs` preventing signature checking), and not an error.

An alternative fix is to remove the `@functools.wraps` call entirely in the implementation of `signed_data_view`, but that would potentially produce more problems.

A test is added for this.

### DIFF
--- a/debug_toolbar/decorators.py
+++ b/debug_toolbar/decorators.py
@@ -32,4 +32,8 @@ def signed_data_view(view):
             )
         return HttpResponseBadRequest("Invalid signature")
 
+    # We have changed the signature, so need to remove `__wrapped__` to
+    # avoid confusing type checking tools
+    del inner.__wrapped__
+
     return inner
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,8 +1,10 @@
+import inspect
 import unittest
 
 from django.test import override_settings
 
 import debug_toolbar.utils
+from debug_toolbar.decorators import signed_data_view
 from debug_toolbar.utils import (
     get_name_from_obj,
     get_stack,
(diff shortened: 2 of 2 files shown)


matthiask (approved):
Thank you! Looks good to me. I checked the `functools.update_wrapper` implementation and I think this makes sense. I don't know much about typechecking though so I'll wait for a moment before merging.

tim-schilling:
I'm less sure about this. We're removing the attribute that's used to fetch the wrapped function. I can imagine that this will cause weirder issues than `django-urlconfchecks` down the line having a problem with this decorator pattern.

I'd prefer to see the debug toolbars url patterns excluded from evaluation or for us to rework this decorator so that it doesn't pass the verified data as an argument. Or pass it as a part of `kwargs` and hide it as an explicit keyword argument.

tim-schilling:
@spookylukey can you confirm that #1658 resolves the issue with `django-urlconfchecks` please?

spookylukey:
I did some research and found a package `makefun` to help create decorators that produce 100% correct signatures. With it, you would write something like:

```python
@makefun.wraps(view, remove_args="verified_data") 
```
instead of `@functools.wraps`

https://smarie.github.io/python-makefun/#easier-edits

However, this adds a dependency and seems like overkill, unless django-debug-toolbar were itself adopting django-urlconfchecks for its own usage.

tim-schilling:
Ah, that's pretty neat but I'd prefer to avoid adding additional dependencies to the project.

spookylukey:
> @spookylukey can you confirm that #1658 resolves the issue with `django-urlconfchecks` please?

Yes, I've checked and that works.

tim-schilling:
Closing in favor of #1658

PR #1658, opened 2022-08-04
Remove signed_data_view decorator to support url type checking.

tim-schilling:
The package django-urlconfchecks was erroring on the toolbar's
URLs because signed_data_view was injecting an extra parameter
for the views. The alternative to the decorator isn't terrible
so let's use that instead. Eventually this can be shortened
with the walrus operator when py37 support is dropped.

This is an alternative to #1657


matthiask (approved):
Thanks, I like it.

PR #1669, opened 2022-09-10
Profiling panel improvements

tim-schilling:
* Added Profiling panel setting ``PROFILER_THRESHOLD_RATIO`` to give users
  better control over how many function calls are included. A higher value
  will include more data, but increase render time.
* Update Profiling panel to include try to always include user code. This
  code is more important to developers than dependency code.
* Highlight the project function calls in the profiling panel.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -270,7 +270,7 @@ Panel options
   render of the profiling panel, but will exclude data.
 
   This value is used to determine the threshold of cumulative time to include
-  the nested functions. The threshold is calculated by the root calls
+  the nested functions. The threshold is calculated by the root calls'
   cumulative time divided by this ratio.
 
 * ``SHOW_TEMPLATE_CONTEXT``
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -261,7 +261,7 @@ Panel options
 
 * ``PROFILER_THRESHOLD_RATIO``
 
-  Default: ``10``
+  Default: ``8``
 
   Panel: profiling
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -259,6 +259,20 @@ Panel options
   This setting affects the depth of function calls in the profiler's
   analysis.
 
+* ``PROFILER_THRESHOLD_RATIO``
+
+  Default: ``10``
+
+  Panel: profiling
+
+  This setting affects the which calls are included in the profile. A higher
+  value will include more function calls. A lower value will result in a faster
+  render of the profiling panel, but will exclude data.
+
+  This value is used to determine the threshold of cumulative time to include
+  the nested functions. The threshold is calculated by the root calls
+  cumulative time divided by this ratio.
+
 * ``SHOW_TEMPLATE_CONTEXT``
 
   Default: ``True``
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -250,6 +250,18 @@ Panel options
     WHERE "auth_user"."username" = '''test_username'''
     LIMIT 21
 
+* ``PROFILER_CAPTURE_PROJECT_CODE``
+
+  Default: ``True``
+
+  Panel: profiling
+
+  When enabled this setting will include all project function calls in the
+  panel. Project code is defined as files in the path defined at
+  ``settings.BASE_DIR``. If you install dependencies under
+  ``settings.BASE_DIR`` in a directory other than ``sites-packages`` or
+  ``dist-packages`` you may need to disable this setting.
+
 * ``PROFILER_MAX_DEPTH``
 
   Default: ``10``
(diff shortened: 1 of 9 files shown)


docs/panels.rst
```diff
@@ -130,6 +130,11 @@ Profiling information for the processing of the request.
 This panel is included but inactive by default. You can activate it by default
 with the ``DISABLE_PANELS`` configuration option.
 
+The panel will include all function calls made by your project if your using
```
matthiask, line 133:
```suggestion
The panel will include all function calls made by your project if you're using
```

docs/configuration.rst
```diff
@@ -259,6 +259,20 @@ Panel options
   This setting affects the depth of function calls in the profiler's
   analysis.
 
+* ``PROFILER_THRESHOLD_RATIO``
+
+  Default: ``10``
```
matthiask, line 264:
```suggestion
  Default: ``8``
```

I have no idea what this value even means :-) but the code says that the default is 8.

docs/configuration.rst
```diff
@@ -259,6 +259,20 @@ Panel options
...
+
+  Panel: profiling
+
+  This setting affects the which calls are included in the profile. A higher
+  value will include more function calls. A lower value will result in a faster
+  render of the profiling panel, but will exclude data.
+
+  This value is used to determine the threshold of cumulative time to include
+  the nested functions. The threshold is calculated by the root calls
```
matthiask, line 273:
```suggestion
  the nested functions. The threshold is calculated by the root calls'
```

docs/panels.rst
```diff
@@ -130,6 +130,11 @@ Profiling information for the processing of the request.
...
+the setting ``settings.BASE_DIR`` to point to your project's root directory.
+If a function is in a file within that directory and does not include
+``"/site-packages/"`` in the path, it will be included.
```
matthiask, line 136: I had some strange occurrences of broken Python installations when packages ended up in `dist-packages` as well. I couldn't really explain the difference -- all I know is that `site-packages` signify that everything is working well, `dist-packages` signify that many things will end up broken.

matthiask (approved):
Looks good. (I think.)

docs/panels.rst
tim-schilling: We could check for that path here too.

matthiask: I think it would be prudent to do so, both  `dist-packages` and `site-packages` are very improbable to exist outside virtualenvs.

tim-schilling:
So a little context here @matthiask, as I was writing up my tutorial I wrote this view that iterated over 50k posts to calculate how many were created in the last 30, 90 and 180 days. I had a function that was called for each iteration, but this was immediately present in the profiling panel's information. So my initial thought was to change the ` / 8 ` calculation. Then I realized that we really should show any project code regardless of what it does. I ended up not needing the threshold control, but I left the change in case it's helpful to future developers. I'm happy to remove it if you feel we should.

Additionally, do you think we should have a killswitch preventing the `or func.is_project_code()` bypass? I'm thinking maybe we get something wrong with site-packages or dist-packages and there is too much to the profiling stats so it doesn't return.

tim-schilling:
@matthiask is it worth adding another setting to prevent this inclusion?

tim-schilling:
@matthiask can you review the latest changes? If it's still good, we can squash merge it.

matthiask (approved):
Thank you! 

> Additionally, do you think we should have a killswitch preventing the or func.is_project_code() bypass? I'm thinking maybe we get something wrong with site-packages or dist-packages and there is too much to the profiling stats so it doesn't return.

I'm a bit confused by the setting. The default was to skip some frames and now you want to add the option to include the projects' code no matter what? `PROFILER_CAPTURE_PROJECT_CODE = False` sounds to me as if the profiler wouldn't capture the projects' code at all. Naming is always borderline bikeshedding, but maybe we could avoid finding a good name for the setting by not making this configurable at all for now? Or maybe it's just me. I definitely don't have any requested changes and the code looks sound to me.

tim-schilling:
>  The default was to skip some frames and now you want to add the option to include the projects' code no matter what?

Correct. The system currently guesses at what frames it should include. My pitch is that we should include all of the project's code in the profiler. The killswitch I was mentioning ended up being `PROFILER_CAPTURE_PROJECT_CODE`. If that's set to False, it reverts back to how the profiling panel works today which includes frames depending on the timing ratio.

> Naming is always borderline bikeshedding, but maybe we could avoid finding a good name for the setting by not making this configurable at all for now?

That's a fair comment. I feel strongly enough about having it be configurable (I can explain if you want me to) that I'd push for us to come up with a better name or merge it as is.

matthiask:
> That's a fair comment. I feel strongly enough about having it be configurable (I can explain if you want me to) that I'd push for us to come up with a better name or merge it as is.

`PROFILER_SKIP_PROJECT_CODE_FILTERING` or something, maybe? If this doesn't sound better to you feel free to merge as-is :) Thank you!

matthiask:
Let's move ahead. I want to try this out with a current project :)

tim-schilling:
Sorry, it was a long weekend for me. I think we can do another version release after #1673 is merged. Probably a minor version bump?

matthiask:
Nice! And yes, let's make a release after that.

PR #1670, opened 2022-09-11
Fix JS linting error from pre-commit.

tim-schilling:
prettier v3 changed the default of trailing comma to all which seems to be problematic for ESLint.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,6 +42,8 @@ repos:
     hooks:
     -   id: prettier
         types_or: [javascript, css]
+        args:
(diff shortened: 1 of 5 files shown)


tim-schilling:
I did a little digging for eslint configurations, but couldn't find an easy answer.

matthiask:
https://www.npmjs.com/package/eslint-config-prettier would deactivate conflicting ESLint rules, but I think this fix is fine.

tim-schilling:
Hmm, I take that back. I think I like the change from this PR a bit more since it doesn't require use of `additional_dependencies` in the pre-commit yaml.

Issue #1668, opened 2022-08-31
'<' not supported between instances of 'int' and 'str'


OdifYltsaeb:
If session keys are not all strings, then this line (https://github.com/jazzband/django-debug-toolbar/blob/6a34e007e773d1243e3c7ba0b0ee07e4b45514ba/debug_toolbar/panels/request.py#L67) generate a TypeError:

```
TypeError at /myurl

'<' not supported between instances of 'int' and 'str'
```
because:
```
>>> l = ['a', 1, 'x', 22]
>>> sorted(l)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: '<' not supported between instances of 'int' and 'str'

```

matthiask:
How can session keys NOT all be strings? Are you using the `PickleSerializer` by chance? It is deprecated and will be removed in Django 5.0.

OdifYltsaeb:
Not sure. Going through someone else's code, and Yes I know that the Django session guidelines say they should be all strings.. but apparently, someone found a way to bypass it :)

So I don't know... perhaps it makes sense to ignore this report, because it appears someone is trying to use Django the way it is not supposed to be used.

matthiask:
Thank you! 

django-debug-toolbar shouldn't crash anyway, but this makes it a normal priority issue for me. @tim-schilling Maybe something for the Djangocon sprints as well, if nobody gets to it earlier?

tim-schilling:
I'm thinking we sort the session by strings, then anything else goes in at a random order. I think trying to stringify the keys could have averse effects. For example, if someone put a model as the key for the session, it would then call the __str__ function which could be accessing other properties, generating DB requests.

ritiksoni00:
Hy, is anybody working on this issue? @tim-schilling

matthiask:
@ritiksoni00 The djangocon US sprints will take place in October, see https://2022.djangocon.us/sprints/

AFAIK nobody is working on this issue currently. Feel free to jump in!

ritiksoni00:
Thanks, Matthias for pointing this out but I already got the _virtual_(😞) ticket for the conf.  

for this task according to my understanding, I need to convert all the items into `str` right? or I understand it wrong.

matthiask:
My comment from above still applies :) I think the code should just bail out and skip sorting if it fails.

ritiksoni00:
[...]

PR #1673, opened 2022-09-18
added functionality to keep unsort the session dict

ritiksoni00:
#1668

### DIFF
--- a/tests/panels/test_request.py
+++ b/tests/panels/test_request.py
@@ -123,3 +123,15 @@ class RequestPanelTestCase(BaseTestCase):
         self.panel.generate_stats(self.request, response)
         panel_stats = self.panel.get_stats()
         self.assertEqual(panel_stats["session"], data)
+        
+        self.request.session = {
+            "b": "b-value",
+            "a": "a-value",
+        }
+        data = {
+            "list": [("a", "a-value"), ("b", "b-value")]
+        }
+        response = self.panel.process_request(self.request)
+        self.panel.generate_stats(self.request, response)
+        panel_stats = self.panel.get_stats()
+        self.assertEqual(panel_stats["session"], data)        
--- a/tests/panels/test_request.py
+++ b/tests/panels/test_request.py
@@ -123,15 +123,13 @@ class RequestPanelTestCase(BaseTestCase):
         self.panel.generate_stats(self.request, response)
         panel_stats = self.panel.get_stats()
         self.assertEqual(panel_stats["session"], data)
-        
+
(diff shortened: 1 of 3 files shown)


tests/panels/test_request.py
```diff
@@ -104,3 +104,22 @@ def test_namespaced_url(self):
...
+            "data": ["foo", "bar", 1],
+            (2, 3): "tuple_key",
+        }
+        data = {
+            "list": [(1, "value"), ("data", ["foo", "bar", 1]), ((2, 3), "tuple_key")]
+        }
+        response = self.panel.process_request(self.request)
+        self.panel.generate_stats(self.request, response)
+        panel_stats = self.panel.get_stats()
+        self.assertEqual(panel_stats["session"], data)
```
matthiask, line 125:
```suggestion
        self.assertEqual(panel_stats["session"], data)
        
        self.request.session = {
            "b": "b-value",
            "a": "a-value",
        }
        data = {
            "list": [("a", "a-value"), ("b", "b-value")]
        }
        response = self.panel.process_request(self.request)
        self.panel.generate_stats(self.request, response)
        panel_stats = self.panel.get_stats()
        self.assertEqual(panel_stats["session"], data)        
```

matthiask (commented):
Looks good! I think you're only verifying the failing case though, maybe add a second test where the session contains only string keys (which can actually be sorted)?

tests/panels/test_request.py
ritiksoni00: i guess I should also test that in a new test

matthiask: I think it's fine to keep both in a single test (as suggested)

tim-schilling: @ritiksoni00 you can either copy this code into your local environment and commit or use github's "Commit suggestion" button to create a commit on the PR. When we merge it, we'll squash the commits into one.

tim-schilling (changes requested):
Thank you for opening a PR! Please add a mention of this change in `docs/changes.rst`. That helps reduce effort in releasing new versions.

tim-schilling (approved):

tim-schilling:
Thank you for the work on this @ritiksoni00!

ritiksoni00:
> Thank you for opening a PR! Please add a mention of this change in `docs/changes.rst`. That helps reduce effort in releasing new versions.

sorry, tim for not responding.. !

tim-schilling:
No worries, thank you for the PR!

PR #1676, opened 2022-09-25
fix: Simplify logic for Panel.enabled property

adamantike:
With this change, having a panel status set in the cookies exits earlier, as it isn't needed to retrieve the Django settings and panel name to calculate the default value.

### DIFF
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -20,7 +20,12 @@ class Panel:
         return self.__class__.__name__
 
     @property
-    def enabled(self):
+    def enabled(self) -> bool:
+        # The user's cookies should override the default value
+        cookie_value = self.toolbar.request.COOKIES.get("djdt" + self.panel_id)
+        if cookie_value is not None:
+            return cookie_value == "on"
+
         # Check to see if settings has a default value for it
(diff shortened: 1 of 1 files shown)


debug_toolbar/panels/__init__.py
```diff
@@ -20,24 +20,23 @@ def panel_id(self):
         return self.__class__.__name__
 
     @property
-    def enabled(self):
+    def enabled(self) -> bool:
+        # The user's cookies should override the default value
+        cookie_value = self.toolbar.request.COOKIES.get("djdt" + self.panel_id)
```
tim-schilling, line 25: If a user has enabled a panel, then changes their settings file to disable it, wouldn't this override the customized settings?

adamantike: If I understand correctly, that's the same behavior  we have today, as the `default` value is only used when a cookie value is not found in the request.

This shouldn't change how a panel status is decided, just to return faster when the cookie is set.

tim-schilling: You're correct. `DebugToolbar.get_panel_classes` is what I was worried about. I wanted to make sure that a developer was still able to stop using a panel even if the cookies are set.

tim-schilling (approved):

tim-schilling:
Thank you for the PR!

PR #1685, opened 2022-10-21
Auto-update History panel with JavaScript fetch requests.

### LINKED ISSUE #1650: Extend auto-update check to the fetch API
The [fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) and [XHR](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) are different. The example app uses `fetch` so we can't trigger the auto update functionality without a change. Realistically we should instrument both.

tim-schilling:
The fetch and XHR requests don't share internals in JavaScript. We should automatically capture both sets of requests as they occur.

Closes #1650

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -261,6 +261,15 @@ const djdt = {
             document.getElementById("djDebug").dataset.sidebarUrl;
         const slowjax = debounce(ajax, 200);
 
+        function handleAjaxResponse(storeId) {
+            storeId = encodeURIComponent(storeId);
+            const dest = `${sidebar_url}?store_id=${storeId}`;
+            slowjax(dest).then(function (data) {
+                replaceToolbarState(storeId, data);
+            });
+        }
+
+        // Patch XHR / traditional AJAX requests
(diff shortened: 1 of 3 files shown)


debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -270,16 +279,25 @@ const djdt = {
...
-                    });
+                    handleAjaxResponse(this.getResponseHeader("djdt-store-id"));
                 }
             });
             origOpen.apply(this, arguments);
         };
+
+        const origFetch = window.fetch;
+        window.fetch = function (url, options) {
+            const promise = origFetch(url, options);
```
matthiask, line 290: It probably makes no sense and it shouldn't happen anyway but it might be more cautious to use `origFetch.apply(this, arguments)` here (similar to the XHR case above)

Apart from that: Looks good to me, thanks!

tim-schilling: Will do!

tim-schilling: Done. I'll squash the commits after approval.

matthiask (approved):
Looks great!

PR #1686, opened 2022-10-21
Support rerendering the toolbar on HTMX boosted pages.

tim-schilling:
This reworks the JavaScript integration to put most event handlers on document.body. This means we'll have slightly slower performance, but it's easier to handle re-rendering the toolbar when the DOM has been replaced.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -2,7 +2,7 @@ import { $$, ajax, replaceToolbarState, debounce } from "./utils.js";
 
 function onKeyDown(event) {
     if (event.keyCode === 27) {
-        djdt.hide_one_level();
+        djdt.hideOneLevel();
     }
 }
 
@@ -30,9 +30,9 @@ const djdt = {
                 const panelId = this.className;
                 const current = document.getElementById(panelId);
                 if ($$.visible(current)) {
-                    djdt.hide_panels();
+                    djdt.hidePanels();
                 } else {
-                    djdt.hide_panels();
+                    djdt.hidePanels();
 
                     $$.show(current);
                     this.parentElement.classList.add("djdt-active");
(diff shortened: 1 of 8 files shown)


tim-schilling:
@knyghty can you verify that this resolves the issue on your project?

docs/installation.rst
```diff
@@ -199,3 +199,24 @@ The Debug Toolbar currently doesn't support Django Channels or async projects.
...
+HTMX
+^^^^
+
+If you're using `HTMX`_ to `boost a page`_ you will need to add the following
+event handler to your code:
+
+.. code-block:: javascript
+
+    if (typeof htmx !== "undefined") {
+        htmx.on("htmx:afterSettle", function(detail) {
```
tim-schilling, line 213: This should check for the existence of `djdt`. People are likely to paste this in and use it on production.

matthiask: I don't know htmx that well but does it ever happen that `window.htmx` is undefined on such pages? 

Maybe `const h = window.htmx, d = window.djdt; if (h && d) { ... }`? This is all a matter of taste and a very personal opinion (not as a DJDT maintainer!) of course but I prefer `window.<var>` to the somewhat magic global `<var>`.

debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -322,13 +328,17 @@ const djdt = {
...
+        return document.getElementById("djDebug");
+    },
 };
 window.djdt = {
     show_toolbar: djdt.show_toolbar,
     hide_toolbar: djdt.hide_toolbar,
     init: djdt.init,
     close: djdt.hide_one_level,
     cookie: djdt.cookie,
+    get_debug_element: djdt.get_debug_element,
```
matthiask, line 341: I'm a bit sad that we're using a mix of snake_case and camelCase in this file; camelCase seems to be more idiomatic when writing javascript since almost all of the JS runtime itself uses camelCase (especially the DOM) but consistency is more important in the end.

The more important question is maybe if `get_debug_element` really has to be exposed here? Can't the method "just" be an internal helper?

tim-schilling: You're right. It can be internal. I'll switch it to camelCase. And then have another commit changing everything else to camelCase so we can stay consistent moving forward.

docs/installation.rst
tim-schilling: I can make that change.

knyghty:
@tim-schilling once I use the snippet from the example it works fine, yes 👍🏻

docs/installation.rst
tim-schilling: I decided against creating constants and went with `window.htmx`. I did realize that we can't check `window.djdt` there since it may not be loaded in yet. So I pushed that check into the event handler, then wrapped it all with `{% if debug %}` to better signal that this should only be used when the toolbar is installed.

matthiask (approved):
Thanks!

PR #1687, opened 2022-10-21
Add divide by zero protection in timer.js 

### LINKED ISSUE #1416: Error: <rect> attribute width: Expected length, "NaN%".
In the loadEvent line, the width NaN in the Time tab

Screenshot_20201202_125950
Screenshot_20201202_125928

SmailBestybay:
Fixes #1416


tim-schilling:
@matthiask I'm leaving this review to you. We created it during the sprints.

matthiask:
The fallback values are a bit confusing to me (zero in `getLeft`, 100 - getLeft() for the denominator, and then zero again)

The code still does the same thing in the happy path and divisions by zero shouldn't happen anymore so let's merge this. 

Thanks!

PR #1689, opened 2022-10-22
Include panel scripts in content when RENDER_PANELS=True

matthiask:
It seems to me that panel scripts aren't included at all when using `RENDER_PANELS=True` at the moment. This would probably be a bug. I don't use `RENDER_PANELS` anywhere and therefore cannot really test if this change would be correct.

I found this while trying to write a test for the timer panel (because I wanted a green build, re. code coverage)

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/timer.js
+++ b/debug_toolbar/static/debug_toolbar/js/timer.js
@@ -1,7 +1,6 @@
 import { $$ } from "./utils.js";
 
 function insertBrowserTiming() {
-    console.log(["inserted"]);
     const timingOffset = performance.timing.navigationStart,
         timingEnd = performance.timing.loadEventEnd,
         totalTime = timingEnd - timingOffset;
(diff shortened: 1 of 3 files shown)


matthiask:
Addendum: My gut feeling is that this change is correct, I lean towards merging but I really want a second opinion.

tim-schilling:
RENDER_PANELS is used when setting up an asgi app with channels. We should extend the example app to support testing that case, but not necessarily now. I can run this PR against a project of mine that uses ASGI next week if you can wait till then.

matthiask:
Sure! No pressure at all. I already added a `template_source` test here https://github.com/jazzband/django-debug-toolbar/commit/a8fa1d07f52d5480cc9d62e2c06fdf11231a2171 , the build is green already.

tim-schilling:
Alright this should be good now. I spotted a bug where the "inserted" statement I removed (when it was still in the code) wasn't being printed out. Making it a module script seems to have resolved the issue. I'll leave the last review, squash and merge to you if you don't have any questions.

tim-schilling (approved):

matthiask:
No questions, thanks for checking and for improving the test!

PR #1690, opened 2022-10-24
Update package metadata

ofek:
## Background

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://peps.python.org/pep-0517/)/[PEP 660](https://peps.python.org/pep-0660/)) and the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)). As a result, the execution of `setup.py` files is now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

related: https://github.com/jazzband/help/issues/313

## Summary of changes

This implements [PEP 621](https://peps.python.org/pep-0621/), obviating the need for `setup.py` (see note), `setup.cfg`, and `MANIFEST.in`. The build backend `hatchling` (of which I am a maintainer in the [PyPA](https://github.com/pypa)) is now used as that is the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Hatchling is available on all the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hatchling/default.nix), [Alpine Linux](https://github.com/alpinelinux/aports/blob/master/community/py3-hatchling/APKBUILD), [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-hatchling/Makefile)/[OpenBSD](https://github.com/openbsd/ports/blob/master/devel/py-hatchling/Makefile), [Gentoo Linux](https://github.com/gentoo/gentoo/tree/master/dev-python/hatchling), [MacPorts](https://github.com/macports/macports-ports/blob/master/python/py-hatchling/Portfile), [OpenEmbedded](https://github.com/openembedded/openembedded-core/commit/846e806181f1349be29cbce78c5041735dfd7e6f), [Spack](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-hatchling/package.py), [MSYS2](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-hatchling/PKGBUILD), etc.

## Notes

- The version is now sourced directly from `debug_toolbar/__init__.py`, removing a release step
- The source distributions on PyPI are erroneously shipping a build artifact `*.egg-info` from `python setup.py develop`; this is now fixed
- Let me know if I should remove `setup.py`, I followed https://github.com/urllib3/urllib3/blob/main/setup.py
- I updated the pin for the `pypa/gh-action-pypi-publish` Action https://github.com/pypa/gh-action-pypi-publish#-master-branch-sunset-


matthiask (commented):
Thank you! I like the general direction. I have yet to start converting any packages to the new format so I'm not able to judge if this is the best way to do it. 

I have never heard about hatch or hatchling before. I see that it is a pypa repository, and this makes me feel warm and fuzzy inside :) so that's good. But I've certainly seen alternative build systems being used. Why should we use hatchling? (As you see I do not have a strong opinion here myself.)

I remember a discussion re. using something like this in Django proper. It was judged to be too early. I think django-debug-toolbar may be a good place to shake out the last (ha!) bugs but it still is a relatively high-profile project for Django users and I am wary to break things. Is the "new way" stable enough?

I don't know about leaving `setup.py` in there. GitHub seems to require it. But, if you leave it in there, you should probably add a `sys.exit(...)` above the commentary that the lines below will never be executed, otherwise they may be :-)

Thanks!

tim-schilling:
I don't have a strong opinion one way or another for packaging, but it seems like hatch is mature enough for us to use. I would suggest us running a test with test.pypi.org first. I've made a request to the owner of the django debug toolbar package on test.pypi.org to add me as an owner. If I'm able to get those permissions, I'll add you as well @matthiask.

ofek:
> Why should we use hatchling?

Not a complete list but here's from a recent comment -

Hatchling:

- Supports static analysis tools for editable installations by default whereas the new setuptools [does not](https://github.com/microsoft/pylance-release/issues/3407) i.e. it requires enabling an option for IDEs to work.
- Uses [Git-style glob patterns](https://git-scm.com/docs/gitignore#_pattern_format) rather than stdlib globs to better match user expectations.
- Allows easy and explicit control of what gets shipped in the `sdist` [target](https://hatch.pypa.io/latest/config/build/) whereas with setuptools it's an objectively confusing mix of conditional wheel options and a [`MANIFEST.in`](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html) file.

> I don't know about leaving `setup.py` in there.

Can I remove it?

matthiask:
> Hatchling: [...]

Thanks!

> Can I remove it?

I think it probably makes more sense to copy urllib3 there as well? See https://github.com/community/community/discussions/6456

ofek:
Done!

ofek:
@hugovk 

> One suggestion: shall we add https://github.com/tox-dev/pyproject-fmt/ to pre-commit to consistently format `pyproject.toml`?

Can we do that in a separate PR?

tim-schilling:
@matthiask It'll take at least one more week until I have ownership of the test pypi repo. Do you want to do a release with the code before this change, then this after we test the release process on test.pypi.org? Or batch it all together?

matthiask:
@tim-schilling I think it's fine to batch it. We can always push another release or revert in the worst case.

ofek:
another just merged https://github.com/jazzband/prettytable/pull/212

docs/contributing.rst
```diff
@@ -167,12 +167,12 @@ The release itself requires the following steps:
...
 
 #. Bump version numbers in ``docs/changes.rst``, ``docs/conf.py``,
-   ``README.rst``, ``debug_toolbar/__init__.py`` and ``setup.cfg``.
+   ``README.rst``, and ``debug_toolbar/__init__.py``.
    Add the release date to ``docs/changes.rst``. Commit.
 
 #. Tag the new version.
 
-#. ``python setup.py sdist bdist_wheel upload``.
+#. ``tox packaging``.
```
tim-schilling, line 175: Shouldn't this be `tox -e packaging`?

ofek: Fixed!

tim-schilling (approved):
Alright I'm good with this. Thank you for the PR! @matthiask I don't feel strongly enough about testing with test.pypi.org to stop us from releasing this.

matthiask:
Me neither, let's merge it then. Thanks all!

pauloxnet:
Great. Can we create a recipe with steps to apply this update in other jazzband packages ?

ofek:
First step: https://hatch.pypa.io/latest/intro/#existing-project

hugovk:
> @hugovk
> 
> > One suggestion: shall we add https://github.com/tox-dev/pyproject-fmt/ to pre-commit to consistently format `pyproject.toml`?
> 
> Can we do that in a separate PR?

Please see PR https://github.com/jazzband/django-debug-toolbar/pull/1700.

PR #1697, opened 2022-11-02
Create one-time mouseup listener for each mousedown

scuml:
Move the creation on the mouseup event handler to inside the mousedown and add {once: true} so that is is removed after each use.  This prevents multiple mouseup events from getting attached to `document` (which breaks mouseup behavior) when using libraries that update page content without reloading the page like Hotwire Turbo, htmx, and Unpoly.


tim-schilling (approved):

tim-schilling:
Thanks for the PR and cleaning up my mess!

PR #1698, opened 2022-11-02
Fix highlighting on history panel

scuml:
`#djDebug .djdt-panelContent tbody > tr:nth-child(odd)` is overriding `#djDebug .djdt-highlighted` on the history panel so every odd row will not take on a highlighted appearance.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -621,7 +621,7 @@
 #djDebug .djdt-max-height-100 {
     max-height: 100%;
 }
-#djDebug .djdt-highlighted{
+#djDebug .djdt-highlighted {
     background-color: lightgrey;
 }
 #djDebug tr.djdt-highlighted.djdt-profile-row {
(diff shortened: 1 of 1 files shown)


tim-schilling:
Thanks for the PR!

I'm not a fan of using `!important` we should be able to design our css in a way that avoids that since it's pretty small. Can you give us the scenario that produces the effect this fixes?

And as a side note, the commit message here should contain the description you used in the PR. That description provides a fair amount of context of why the change was made and would be beneficial in our git history.

scuml:
Yeah - not a fan of !important myself, but thought this might be a good use of it.  Check out the history panel - every odd row for me does not highlight when `switch` is pressed.

Screen Shot 2022-11-02 at 3 12 40 PM

scuml:
It looks like our options are `!important` or the awkward...
```
#djDebug .djdt-highlighted, #djDebug .djdt-panelContent .djdt-highlighted:nth-child(odd){
    background-color: lightgrey;
}
```

tim-schilling:
Hmm. Yeah, I'm not sure. It's not terrible with a linebreak:

```css
#djDebug .djdt-highlighted,
#djDebug .djdt-panelContent .djdt-highlighted:nth-child(odd){
    background-color: lightgrey;
}
```

Another option is to use the `:not()` operator, `#djDebug .djdt-panelContent tbody > tr:nth-child(odd):not(.djdt-highlighted)`

scuml:
Good thought on `:not()` - that seems more sensible.

tim-schilling (approved):

tim-schilling:
Thank you for the PR! This is a nice fix to have.

PR #1703, opened 2022-11-09
Allow Language Selection on the Toolbar

### LINKED ISSUE #991: Toolbar language selection
django-debug-toolbar picks up the main apps locale to determine it's language. I'd really like to override this via a setting.

The app I'm working on is in Spanish, but it doesn't really make sense to have the toolbar in spanish; even as a native spanish speaker living in a spanish-speaking country, all the words are weird to me because we use the english words in our day to day work (even when talking between ourselves), and we always use development software in english.

(TBH, I've never seen a developer work with tools configured to use spanish, since all online docs, etc are in english, so all the entire developer subculture has kinda adopted english terms).

leandrodesouzadev:
closes #991

Added a new optional setting `TOOLBAR_LANGUAGE` that can be set by a user to define in which language the toolbar is displayed.
Modified the `render_toolbar` method from `DebugToolbar` class to use the `django.utils.translations.override`, so the language is set either: To the defined language on the `TOOLBAR_LANGUAGE` or the current language (No breaking changes). After the template is rendered, the language is set back to the previous.

Added two tests cases:
One that checks if the toolbar is rendered accordingly when this setting was not defined, the default behavior. 
Other that checks if the toolbar was rendered to the set language.

Updated documentation on `configuration.rst` on how to set this value.

### DIFF
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -40,6 +40,7 @@ CONFIG_DEFAULTS = {
     "SKIP_TEMPLATE_PREFIXES": ("django/forms/widgets/", "admin/widgets/"),
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
+    "TOOLBAR_LANGUAGE": None,
 }
 
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -149,7 +149,18 @@ Toolbar options
   the request doesn't originate from the toolbar itself, EG that
   ``is_toolbar_request`` is false for a given request.
 
+* ``TOOLBAR_LANGUAGE``
 
+  Default: ``None``
+
+  The language used to render the toolbar. If no value is supplied, then the
+  application's current language will be used. This setting can be used to
+  render the toolbar in a different language than what the application is
+  rendered in. For example, if you wish to use English for development,
+  but want to render your application in French, you would set this to
+  ``"en-us"`` and `settings.LANGUAGE_CODE`_ to ``"fr"``.
+
+.. _settings.LANGUAGE_CODE: https://docs.djangoproject.com/en/stable/ref/settings/#std-setting-LANGUAGE_CODE
 
 Panel options
 ~~~~~~~~~~~~~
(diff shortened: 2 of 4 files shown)


docs/configuration.rst
```diff
@@ -149,6 +149,14 @@ Toolbar options
...
   ``is_toolbar_request`` is false for a given request.
 
+* ``TOOLBAR_LANGUAGE``
+
+  Default: ``None``
+
+  The language that will be used to render the toolbar. If no value is supplied
+  then the current language will be used. After the toolbar is rendered, the
+  language is set back to the previous one. A common case is to set this to
+  ``en-us`` since most developers use english on projects.
```
tim-schilling, line 159: This is pretty good. I think it can be improved if we explain the scenario that would be useful though. What do you think?

> The language used to render the toolbar. If no value is supplied, then the application's current language will be used. This setting can be used to render the toolbar in a different language than what the application is rendered in. For example, if you wish to use English for development, but want to render your application in French, you would set this to ` ``"en-us"`` ` and ` (``LANGUAGE_CODE``)[https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-LANGUAGE_CODE] ` to `fr`.

debug_toolbar/settings.py
```diff
@@ -40,6 +40,7 @@
     "SKIP_TEMPLATE_PREFIXES": ("django/forms/widgets/", "admin/widgets/"),
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
+    "TOOLBAR_LANGUAGE": None,
```
tim-schilling, line 43: I'm torn between having this be LANGUAGE or TOOLBAR_LANGUAGE. Does anyone have a strong opinions?

docs/configuration.rst
leandrodesouzadev: That's great!
English is not my primary language, so the text might be have a lot of spelling problems :laughing:

leandrodesouzadev: I had pushed a new commit with this change.

debug_toolbar/settings.py
matthiask: I think both are fine, or even `LANGUAGE_CODE` (same as `settings.LANGUAGE_CODE`); the `TOOLBAR_` prefix is a bit redundant since all toolbar settings mostly affect the toolbar.

No strong opinions :)

tim-schilling (approved):
I'm good with this as is.

tim-schilling:
Squashed your commits.

Issue #1705, opened 2022-11-11
Add Type hints to the codebase


leandrodesouzadev:
Is there any plans to add type hints to the codebase?
As a newcomer to the project, i feel like there's a lot of places that could use type hints to better describe functions.
I'm willing to do the job! :mechanical_arm:

tim-schilling:
I don't have major any qualms against it. I think they're useful for method signatures and class members. However, I'm not committing getting to a point where we are going to say we need type hints on everything.

tim-schilling:
@leandrodesouzadev looks like you have the go-ahead on this. I'd suggest doing it piecemeal so you don't end up with too many merge conflicts from long living branches. And so you don't burn yourself out. Thank you!

leandrodesouzadev:
Looks great, baby steps all the way.
Maybe one module per pull request? How that sounds to you?
Monday I will start working on this.

tim-schilling:
That sounds good! Let us know if you need any help.

tim-schilling:
@leandrodesouzadev circling back to this. How are things going?

leandrodesouzadev:
I haven't touched on the code base for a while now.
I will try to do some other module or two this week and open the pull requests

ataylor32:
I would like to see this addressed too. Currently, I'm getting this error with mypy:

```
error: Skipping analyzing "debug_toolbar": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```

tim-schilling:
@ataylor32 is this something you would have time to help out with?

ataylor32:
@tim-schilling Possibly, but I don't have experience with creating stubs. I'll have to look into it.

PR #1707, opened 2022-11-14
Type hints: `utils` module

leandrodesouzadev:
typing: Adds type hints to the `utils` module. A `stubs` file was added with type definitions.

Partial implementation of #1705.


debug_toolbar/utils.py
```diff
@@ -174,7 +177,7 @@ def get_name_from_obj(obj):
     return name
 
 
-def getframeinfo(frame, context=1):
+def getframeinfo(frame: Any, context=1) -> inspect.Traceback:
```
tim-schilling, line 180: `context` was missed here.

debug_toolbar/utils.py
```diff
@@ -131,7 +132,7 @@ def get_template_info():
     return template_info
 
 
-def get_template_context(node, context, context_lines=3):
+def get_template_context(node, context, context_lines=3) -> Dict[str, Any]:
```
tim-schilling, line 135: The parameters here were missed.

tim-schilling (commented):
It looks like `get_sorted_request_variable` needs a type hint as well.

debug_toolbar/utils.py
leandrodesouzadev: Just pushed a new commit.

debug_toolbar/utils.py
leandrodesouzadev: Done.

debug_toolbar/stubs.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any, List, NamedTuple, Optional, Tuple
+
+from django import template as dj_template
```
tim-schilling, line 3: @matthiask Admittedly I haven't done typing in a project of my own that requires a stubs file. I'm unsure how I feel about a single stub file without docs or some way to determine if they're used someplace. However, I think I'm okay with us using this until we find it doesn't work.

Thoughts?

tim-schilling (approved):
I'm good with these changes, but want to see if we need a discussion regarding stubs.py before merging.

matthiask:
I'm only wondering if the module should be renamed to `_stubs` to make it clear that it's not supposed to be used from outside django-debug-toolbar (not even by third party panels). At the very least I'd expect a comment stating that these stubs are purely internal.

Apart from that I have to admit that I haven't used typing on real world projects yet. I have played around with mypy a bit but that's all, so I can't really say if these changes make sense or not. FWIW I approve this pull request as well :)

tim-schilling:
@leandrodesouzadev could you make the changes Matthias outlined? I think they're worthwhile.

leandrodesouzadev:
@tim-schilling I just pushed a new commit with the changes outlined by @matthiask.

matthiask:
Let's get starting.

matthiask:
Thanks!

sshishov:
You can just check guys how the types are added into other libraries

PR #1712, opened 2022-12-03
Add a pull request template for future changes.

tim-schilling:
This should help reduce maintenance overhead when drafting new releases. The test checkbox item is simply to have the developer consider how to test their change. I'm less sure that will result in significant change.


matthiask:
Thanks, I like it! It's a good idea either way, even if it only raises awareness of the changelog.

PR #1717, opened 2022-12-07
Use correct language in panels

amureki:
# Description

Greetings fellows,

PR #1703 introduced a bug where SQLPanel (and probably others) is rendered in `LANGUAGE_CODE` language, while it should use `TOOLBAR_LANGUAGE` if provided.
This PR fixes panel's content language.

I am not sure what else could be missed, I am also open to other implementations.

Best,
Rust

# Checklist:

### DIFF
--- a/debug_toolbar/decorators.py
+++ b/debug_toolbar/decorators.py
@@ -1,6 +1,9 @@
 import functools
 
 from django.http import Http404
+from django.utils.translation import get_language, override as language_override
+
+from debug_toolbar import settings as dt_settings
 
 
 def require_show_toolbar(view):
@@ -13,5 +16,18 @@ def require_show_toolbar(view):
             raise Http404
 
         return view(request, *args, **kwargs)
+
+    return inner
+
+
+def render_with_toolbar_language(view):
+    """Force any rendering within the view to use the toolbar's language."""
+
+    @functools.wraps(view)
+    def inner(request, *args, **kwargs):
+
+        lang = dt_settings.get_config()["TOOLBAR_LANGUAGE"] or get_language()
+        with language_override(lang):
+            return view(request, *args, **kwargs)
 
     return inner
--- a/debug_toolbar/panels/history/views.py
+++ b/debug_toolbar/panels/history/views.py
@@ -1,12 +1,13 @@
 from django.http import HttpResponseBadRequest, JsonResponse
 from django.template.loader import render_to_string
 
-from debug_toolbar.decorators import require_show_toolbar
+from debug_toolbar.decorators import render_with_toolbar_language, require_show_toolbar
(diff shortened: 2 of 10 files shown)


tim-schilling (approved):
This is a good catch. Thank you.

tim-schilling:
@amureki did you happen to test loading a previous request via the history panel to confirm that the language is correct then?

amureki:
> @amureki did you happen to test loading a previous request via the history panel to confirm that the language is correct then?

Honestly, I didn't, was not aware of such specifics. But I can try to add a test for it tomorrow, if I will find a good reference.

tim-schilling:
Sorry, I didn't mean a unit test, but rather testing with the application as a whole. So when you discovered the bug that this PR fixes, did you also use the history panel?

amureki:
@tim-schilling ah, I see what you mean, sorry Tim. I just quickly checked, history panel had also this bug (table titles were in website language, not toolbar language), and it is fixed by this patch. 👍

tim-schilling:
Perfect, thank you!

amureki:
@tim-schilling oh wait. It does switch back to the `LANGUAGE_CODE` when you switch to the old requests. Sorry, I never used this panel TBH.

tim-schilling:
@amureki this is a more complicated change than expected. I'm going to push a change shortly that should address more of the issues.

tim-schilling:
@matthiask can you review the changes in this PR when you're available?

matthiask (approved):
I'm not completely sure if it wouldn't be better to send the toolbar's language as an additional query parameter to the individual views, because now you could still get a toolbar with mixed languages by changing the language preference after the initial page load (by going into the browser settings)

But that may also be a new requirement (or enhancement). FWIW I think the change is fine. I'm not *that* happy with the aliased imports (`dt_settings`) but that's the way it's done in the toolbar source so that's fine with me.

tim-schilling:
> I'm not that happy with the aliased imports (dt_settings) but that's the way it's done in the toolbar source so that's fine with me.

So we're on the same page, what would be your preference?

matthiask:
@tim-schilling I opened a new issue for this at #1720 , let's continue the discussion about aliases there.

matthiask:
@amureki Do you have an opinion regarding the following comment?

> I'm not completely sure if it wouldn't be better to send the toolbar's language as an additional query parameter to the individual views, because now you could still get a toolbar with mixed languages by changing the language preference after the initial page load (by going into the browser settings)
>
> But that may also be a new requirement (or enhancement). FWIW I think the change is fine. [...]

amureki:
@matthiask I am generally happy with @tim-schilling solution for this problem. In my opinion, it solves it for the majority of the usage cases. I'd be happy to report any further issues I'll experience with the languages, but we definitely can move on with merging this PR.

PR #1719, opened 2022-12-09
Drop support for Python 3.7

matthiask:
# Description

This pull request drops support for Python 3.7. Older versions of the toolbar will still be usable by people still using 3.7. This also makes us use less CI resources, something which is a big plus for me.

Also, see https://pythonspeed.com/articles/stop-using-python-3.7/

# Checklist:

### DIFF
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -45,7 +45,7 @@ CONFIG_DEFAULTS = {
 }
 
 
-@lru_cache()
+@lru_cache
 def get_config():
     USER_CONFIG = getattr(settings, "DEBUG_TOOLBAR_CONFIG", {})
     CONFIG = CONFIG_DEFAULTS.copy()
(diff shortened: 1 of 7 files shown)


matthiask:
@tim-schilling I added a removal of Django 4.0 to this since the end of the extended support is close and our CI matrix is already really big. Everything should still work with 4.0 but we don't spend CI time verifying it over and over.

pyproject.toml
```diff
@@ -27,13 +27,13 @@ classifiers = [
     "Framework :: Django :: 3.2",
     "Framework :: Django :: 4.0",
     "Framework :: Django :: 4.1",
+    # "Framework :: Django :: 4.2",
```
matthiask, line 30: The identifier doesn't seem to be available yet.

debug_toolbar/settings.py
```diff
@@ -70,7 +70,7 @@ def get_config():
 ]
 
 
-@lru_cache()
+@lru_cache
```
matthiask, line 73: Automatically changed by `pyupgrade`.

tim-schilling:
Another option to dropping 4.0 from the test suite entirely is to only test it with sqlite with the latest version of python it supports.

matthiask:
Good idea. I left `py310-dj40-sqlite` in there.

PR #1732, opened 2023-01-20
Remove the logging panel as discussed

### LINKED ISSUE #695: Enabling debug_toolbar seems to break logging from management commands

### LINKED ISSUE #906: Object leak in logging panel when using threads
I recently tracked down a memory leak in our app, which turned out to be caused by the debug toolbar.

Our Django app launches and destroys a thread in the context of a view (to talk to a websocket server). When that thread writes to logs, debug_toolbar creates a reference to the Thread instance  which it never clears. Disabling DEBUG makes no difference.

Essentially [this call](https://github.com/jazzband/django-debug-toolbar/blob/e7a9ed0c3de1c8635f7e8c83349cb77cef42489f/debug_toolbar/panels/logging.py#L83) only clears the view thread, and not any sub-threads that may have been created in the view.

I think you might want to make `_records` in `LoggingPanel` a weak key dictionary, if I have understood the code correctly...

### LINKED ISSUE #1119: object browser in log messages
When I log an object using the message I get is generic and not very helpful. 

For example if I call `logging.debug(request)` I get `<rest_framework.request.Request object at 0x7ff3e20ad0b8>`.

Is there a way to make the object browse-able. I was thinking something like a [+] button that shows all the properties of the object and their values.

### LINKED ISSUE #1681: Overly loud tests in the development environment
We recently went through an upgrade of Django and Python, including the supporting tools, and suddenly our tests started outputting lots of extra information (that was not there before), when run in the development environment. After countless hours trying to find the source of the output, I finally discovered that DDT now adds a `lastResort` handler to the root logger. This change, implemented in #1603, is fairly recent (merged 6 months ago) and, unfortunately, is not documented in the changelog.
The difference between the development environment and the CI testing environment is, of course, that in the CI flow DDT is not installed.
How can DDT be updated so as to avoid all the verbose output being printed, when tests are run?

### LINKED ISSUE #1683: Memory leak from logging panel, even when debug_toolbar middleware is not active
If `debug_toolbar` is ever imported (e.g. to add the URL) then it will set up its collection log handler (https://github.com/jazzband/django-debug-toolbar/blob/main/debug_toolbar/panels/logging.py#L69)

This happens even if the middleware is not active.

This leads to high memory usage by collecting and retaining all log entries from the root logger.

This is obviously desirable for request handling to display the logs in the toolbar, where the middleware is active, but we encountered this with a django command consuming ever more memory as it ran.

We removed it from `INSTALLED_APPS` and `MIDDLEWARE` using an ENV var to exclude it, but we had not removed it from `urlpatterns` and so `debug_toolbar.panels.logging` was still being loaded and was adding its collection logger.

Similarly, if the logger panel is in `DISABLE_PANELS` the logger is still added. It is only if the panel is not present in `DEBUG_TOOLBAR_PANELS` that it is never loaded and the collection logger is not set up.

We now exclude debug\_toolbar from our `urlpatterns` too, however it would be good if debug\_toolbar only set up this collection logger when it is actually active, rather than always adding it when the module is loaded.

---

This doesn't appear to be #906, which related explicitly to threading, but has been referenced and worked around by reef-technologies/cookiecutter-rt-django#81 in a similar way to our workaround.

matthiask:
Closes #1683, closes #1681, closes #1119, closes #906, closes #695.

# Description

Sorry for rushing ahead. I looked at the ticket tracker again and found all those open issues that could be closed by this change. I had to do it 😅


tim-schilling:
Love it! Though we should do what you originally suggested and replace the logging panel class with a stub class that does nothing, but logs a warning to be removed from the config. That way when someone upgrades we don't break their project.

debug_toolbar/settings.py
```diff
@@ -76,6 +76,14 @@ def get_panels():
...
     except AttributeError:
         PANELS = PANELS_DEFAULTS
+
+    logging_panel = "debug_toolbar.panels.logging.LoggingPanel"
+    if logging_panel in PANELS:
+        PANELS = [panel for panel in PANELS if panel != logging_panel]
+        warnings.warn(
+            f"Please remove {logging_panel} from your DEBUG_TOOLBAR_PANELS setting.",
+            DeprecationWarning,
+        )
```
matthiask, line 86: @tim-schilling Have you seen this block? I have to admit I haven't tested it properly but shouldn't this be sufficient? 

(It wouldn't work if people imported the panel themselves but that wasn't documented anyway.)

tim-schilling (approved):
100% agreement with this. We should probably update the mo / po translation files with this. I forget which one is first off the top of my head.

matthiask:
First .po (text) then .mo (compiled); you did the Transifex sync last time, I have never managed to get it working (but haven't spent much time doing it)

It's not critical since we do not have any new strings to translate though.

tim-schilling:
Gotcha, thank you for the clarification. I do think you should regenerate the po files. The updated po files would have the logging related translated strings removed, which if we have any translators currently working, would reduce their workload. I'm pretty sure transifex will pick up the change once it's merged into automatically. I'll handle the syncing of the mo files back into the repo later.

`python -m django makemessages --no-obsolete` should do the trick.

matthiask:
`cd debug_toolbar && ../.tox/py310-djmain-sqlite/bin/django-admin makemessages -a --no-obsolete` did it.

I pushed the change; the CI is green, let's merge it :)

tim-schilling:
Sounds good to me!

Shriukan33:
Thanks !

tim-schilling:
I realized I probably should include what we actually discussed on our call.

I originally proposed changing our logging panel to no longer try to capture logs as soon as the panel was imported. Instead it would have only started monitoring when the panel was initialized. This would allow for us to keep the old version of the panel around if needed.

Matthias proposed simply disabling the panel by default. This would have reduced the number of issues created, but wouldn't actually close the issues.

Eventually we realized that we weren't sure if the logging panel even worked. Some exploratory attempts didn't show any thing. After discussing the purpose of the logging panel, we decided that because logs could be configured and monitored elsewhere (the terminal), there was less benefit in presenting the information as a part of the request in the toolbar.

Hopefully this helps us keep track of our decision making in the future.

matthiask:
Thanks for writing it down!

goetzb:
Thanks for the summary - I'm surprised by this change, I liked to use the panel sometimes for quick debugging of views. I'd have preferred to at least keep it around as an optional panel that's off by default but could still be activated easily.

matthiask:
I totally understand. For the record: The general idea of a logging panel hasn't been rejected at all and if someone were to propose adding a new logging panel without the issues of the old one we'd certainly consider it and quite certainly merge it!

PR #1737, opened 2023-02-06
added support and test cases for psycopg3

### LINKED ISSUE #1736: Add Support for psycopg3
With the release of django 4.2a we can use psycopg3. However [django-debug-toolbar](https://github.com/jazzband/django-debug-toolbar) is not compatible with it.

nofalx:
# Description

Add support for psycopg3 while still keeping support for psycopg2

Fixes #1736

# Checklist:

- [x] Fix test `tests.panels.test_sql.SQLPanelMultiDBTestCase.test_transaction_status`

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -133,10 +133,10 @@ class NormalCursorWrapper(BaseCursorWrapper):
     def _decode(self, param):
         if PostgresJson and isinstance(param, PostgresJson):
             # psycopg3
-            if hasattr(param, 'obj'):
+            if hasattr(param, "obj"):
                 return param.dumps(param.obj)
             # psycopg2
-            if hasattr(param, 'adapted'):
+            if hasattr(param, "adapted"):
                 return param.dumps(param.adapted)
 
         # If a sequence type, decode each element separately
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -10,6 +10,7 @@ from debug_toolbar.utils import get_stack_trace, get_template_info
 
 try:
     import psycopg
+
     PostgresJson = psycopg.types.json.Jsonb
     STATUS_IN_TRANSACTION = psycopg.pq.TransactionStatus.INTRANS
 except ImportError:
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -132,7 +132,12 @@ class NormalCursorWrapper(BaseCursorWrapper):
 
     def _decode(self, param):
         if PostgresJson and isinstance(param, PostgresJson):
-            return param.dumps(param.obj)
+            # psycopg3
+            if hasattr(param, 'obj'):
+                return param.dumps(param.obj)
+            # psycopg2
+            if hasattr(param, 'adapted'):
+                return param.dumps(param.adapted)
 
         # If a sequence type, decode each element separately
         if isinstance(param, (tuple, list)):
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -155,7 +155,7 @@ class NormalCursorWrapper(BaseCursorWrapper):
         if vendor == "postgresql":
             # The underlying DB connection (as opposed to Django's wrapper)
             conn = self.db.connection
-            initial_conn_status = conn.info.status
+            initial_conn_status = conn.info.transaction_status
 
         start_time = time()
         try:
@@ -214,7 +214,7 @@ class NormalCursorWrapper(BaseCursorWrapper):
                 # case where Django can start a transaction before the first query
                 # executes, so in that case logger.current_transaction_id() will
                 # generate a new transaction ID since one does not already exist.
-                final_conn_status = conn.info.status
+                final_conn_status = conn.info.transaction_status
                 if final_conn_status == STATUS_IN_TRANSACTION:
                     if initial_conn_status == STATUS_IN_TRANSACTION:
                         trans_id = self.logger.current_transaction_id(alias)
@@ -223,6 +223,7 @@ class NormalCursorWrapper(BaseCursorWrapper):
                 else:
                     trans_id = None
 
+                print(trans_id)
                 params.update(
                     {
                         "trans_id": trans_id,
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -224,7 +224,6 @@ class NormalCursorWrapper(BaseCursorWrapper):
                 else:
                     trans_id = None
 
-                print(trans_id)
                 params.update(
                     {
                         "trans_id": trans_id,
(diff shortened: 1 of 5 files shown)


tests/panels/test_sql.py
```diff
@@ -225,14 +225,33 @@ def test_json_param_conversion(self):
...
+                    "SELECT * FROM tests_postgresjson WHERE field ->> 'key' = ANY(%s)",
+                    [["a", "b'"]],
+                )
+            )
+        else:
+            list(
+                PostgresJSON.objects.raw(
+                    "SELECT * FROM tests_postgresjson WHERE field ->> 'key' IN %s",
+                    [("a", "b'")],
+                )
```
matthiask, line 253: Since this test is only run on PostgreSQL anyway I think you could keep just the psycopg3 branch since it should work with psycopg2 as well.

matthiask (commented):
Thanks for starting this! Most of the code looks good but I haven't given it a very close reading yet.

tests/panels/test_sql.py
nofalx: kept the psycopg3 version

pauloxnet:
there's typos in tile and description: it's 'psycopg' and not 'psycorg'

debug_toolbar/panels/sql/tracking.py
```diff
@@ -205,7 +214,7 @@ def _record(self, method, sql, params):
                 # case where Django can start a transaction before the first query
                 # executes, so in that case logger.current_transaction_id() will
                 # generate a new transaction ID since one does not already exist.
-                final_conn_status = conn.status
```
nofalx, line 208: Seems like the old logic was based on connection status, 
https://www.psycopg.org/docs/extensions.html#connection-status-constants

however according to the new api in psycopg3 `connection.info.status` we only have 2 states
https://www.psycopg.org/psycopg3/docs/api/pq.html#psycopg.pq.ConnStatus

Since `STATUS_IN_TRANSACTION` is never in any of the options of `connection.info.status` the test `tests.panels.test_sql.SQLPanelMultiDBTestCase.test_transaction_status` will always fail. How should we based the logic for the new states of `psycopg3`? should we rely on `conn.info.transaction_status` instead of `connection.info.status`

tox.ini
```diff
@@ -3,17 +3,18 @@ isolated_build = true
 envlist =
     docs
     packaging
-    py{38,39,310}-dj{32,41,42}-{sqlite,postgresql,postgis,mysql}
+    py{38,39,310}-dj{32,41,42}-{sqlite,legacy_postgresql,postgis,mysql}
     py{310}-dj{40}-{sqlite}
-    py{310,311}-dj{41,42,main}-{sqlite,postgresql,postgis,mysql}
+    py{310,311}-dj{41,42,main}-{sqlite,legacy_postgresql,postgresql,postgis,mysql}
```
matthiask, line 8:
```suggestion
    py{310,311}-dj{41}-{sqlite,legacy_postgresql,postgis,mysql}
    py{310,311}-dj{42,main}-{sqlite,legacy_postgresql,postgresql,postgis,mysql}```

Untested, but should probably fix the problem where you're still running tests with Django 4.1/psycopg3.

File: tox.ini [hunk 4]
```diff
@@ -3,17 +3,19 @@ isolated_build = true
 envlist =
     docs
     packaging
-    py{38,39,310}-dj{32,41,42}-{sqlite,postgresql,postgis,mysql}
+    py{38,39,310}-dj{32,40,41}-{sqlite,legacy_postgresql,postgis,mysql}
     py{310}-dj{40}-{sqlite}
-    py{310,311}-dj{41,42,main}-{sqlite,postgresql,postgis,mysql}
+    py{310,311}-dj{40,41}-{sqlite,legacy_postgresql,postgis,mysql}
```
matthiask, line 8: Now you're adding back many Django 4.0 test runs, those have been removed on purpose in https://github.com/jazzband/django-debug-toolbar/pull/1719. That's not better.

nofalx: my bad, should be good this time

tests/panels/test_sql.py
```diff
@@ -225,12 +225,16 @@ def test_json_param_conversion(self):
...
+        https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#you-cannot-use-in-s-with-a-tuple
+        """
         self.assertEqual(len(self.panel._queries), 0)
 
         list(
             PostgresJSON.objects.raw(
-                "SELECT * FROM tests_postgresjson WHERE field ->> 'key' IN %s",
-                [("a", "b'")],
+                "SELECT * FROM tests_postgresjson WHERE field ->> 'key' = ANY(%s)",
+                [["a", "b'"]],
```
tim-schilling, line 237: I'm not sure this test should change. It should be updated to expectedly fail with psycopg3.

matthiask: Ah sorry, I suggested only keeping the new version since that one would work both with psycopg2 and psycopg3. If you want to keep the old version around as well and just run one of them depending on the installed psycopg version that's fine too.

tim-schilling: I think I'm more comfortable with that since this is a regression test (https://github.com/jazzband/django-debug-toolbar/pull/1722) and the updated version avoids that test case altogether

matthiask: I wasn't sufficiently aware of that. I agree then. Sorry @nofalx

tox.ini
nofalx: It seems that github actions still tries to run python3.8 with django 4.2a
https://github.com/jazzband/django-debug-toolbar/actions/runs/4113633936/jobs/7102476809#step:9:26

can you help adjust the test to prevent that?

tests/panels/test_sql.py
```diff
@@ -225,14 +225,32 @@ def test_json_param_conversion(self):
...
-            PostgresJSON.objects.raw(
-                "SELECT * FROM tests_postgresjson WHERE field ->> 'key' IN %s",
-                [("a", "b'")],
+        if is_psycopg3:
+            list(
+                PostgresJSON.objects.raw(
+                    "SELECT * FROM tests_postgresjson WHERE field ->> 'key' = ANY(%s)",
+                    [["a", "b'"]],
+                )
+            )
```
tim-schilling, line 246: Sorry, we had a misunderstanding. This block does not fulfill the purpose of the test. If psycopg3 doesn't support tuples as a parameter, this test should fail expectedly. However, it doesn't look like there's a `expectedFailureIf` decorator, so maybe we update the `skipUnless` to include confirming it's psycopg2?

nofalx: was not sure how to check if we have psycopg3

so i relied on the method used with django to detect psycopg version
https://github.com/django/django/blob/main/django/db/backends/postgresql/base.py#L24

and combined it with `self.skipTest`

let me know if there is a better way to skip the test

tim-schilling: @nofalx can you enable maintainers to edit the PR? I'd like to push a change. It should be on the right side near the notifications button. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

nofalx:
closed in favor for https://github.com/jazzband/django-debug-toolbar/pull/1739 as the option "Allow edits by maintainers
" is not allowed here

PR #1752, opened 2023-04-02
SQL query formatting improvements

living180:
# Description

Various improvements to the code which formats SQL statements for the SQL panel.  The majority are performance improvements, with an overall 35% reduction in formatting time on my system.  However, there is also an improvement in indentation for queries which use the `CASE` keyword, and better simplification of queries generated by `.count()` querysets.

# Checklist:

### DIFF
--- a/debug_toolbar/panels/sql/utils.py
+++ b/debug_toolbar/panels/sql/utils.py
@@ -1,8 +1,8 @@
 import re
 from functools import lru_cache
+from html import escape
 
 import sqlparse
-from django.utils.html import escape
 from sqlparse import tokens as T
 
 from debug_toolbar import settings as dt_settings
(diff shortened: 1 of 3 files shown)


matthiask (approved):
Thanks for refactoring the way settings are modified! That's an excellent change.

The code amount increase makes me a bit sad, but I think NOT hiding `count(*)` queries etc. is definitely a net positive. 

Until now we haven't used `sqlparse`'s `insert_(before|after)` methods. They are documented https://sqlparse.readthedocs.io/en/latest/analyzing/?highlight=insert_before#sqlparse.sql.TokenList.insert_before but I find the `idx` manipulation and those methods a bit disquieting. The test suite seems to cover the relevant cases, we are depending on `sqlparse` anyway and it has been quite stable over the years and that's sufficient for me, but I want to wait a bit if anyone else has some reservations about this change before I'll merge it. 

Thanks!

matthiask:
Thanks!

PR #1753, opened 2023-04-02
Allow per-DB port specification for tox

living180:
As currently written, tox.ini allows specifying the DB port to use to connect to the test DB using a `DB_PORT` environment variable.  However, it uses the same variable for the mysql, postgresql, and postgis environments.  Thus if specifying a custom `DB_PORT` environment variable when running the full combination of tox environments, the mysql, postgresql, and postgis environments will all attempt to use the same DB port, which is unlikely to result in happiness.

Fix by checking for three new environment variables, `POSTGRESQL_DB_PORT`, `POSTGIS_DB_PORT`, and `MYSQL_DB_PORT` in the respective environments, falling back to `DB_PORT` (and then the hard-coded default) for backwards compatibility.

The motivation for this change is that I wanted to run postgresql within a container listening on a non-standard port for testing, but didn't have a way to run all tox over all the environments while setting the port for postgresql/postgis without also affecting the port for mysql.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -20,6 +20,10 @@ Pending
   is rendered, so that the correct values will be displayed in the rendered
   stack trace, as they may have changed between the time the stack trace was
   captured and when it is rendered.
+* Enabled per-DB port specification for tox.  Tox will now check the
+  ``MYSQL_DB_PORT``, ``POSTGRESQL_DB_PORT``, and ``POSTGIS_DB_PORT``
+  environment variables to override the DB port for mysql, postgresql, and
+  postgis environments, respectively.
 
 3.8.1 (2022-12-03)
 ------------------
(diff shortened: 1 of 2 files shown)


matthiask (commented):
I'm not sure I understand the tox documentation fully but maybe factors or something could help with this; if a database environment is a factor (is it?) couldn't you run tox once per DB environment without having to list all individual combinations explicitly?

matthiask:
The following commands seem to produce the correct list of environments for the given databases:

`tox -f postgresql -f postgis -l`
`tox -f mysql -l`

living180:
Thanks for those commands.  Unfortunately they only work with tox 4.x, and I have to stay with tox 3.x for the moment due to some other projects.  I'd still like to be able to just run the full test suite with just a single tox invocation, although using the `TOX_SKIP_ENV` environment variable I might be able to select the appropriate environments similar to using factors as you suggested.  I still need to see why my changes fail on tox 4.x, as they were based on tox 4.x documentation and work fine for tox 3.x.

living180:
So my changes in this PR didn't work with tox 4.x due to https://github.com/tox-dev/tox/issues/2972.  Closing.

PR #1759, opened 2023-04-09
Use the new STORAGES setting in Django 4.2

### LINKED ISSUE #1758: The STATICFILES_STORAGE setting is deprecated in Django 4.2
The _Static files_ panel produces the following warning in Django 4.2

`RemovedInDjango51Warning: The STATICFILES_STORAGE setting is deprecated. Use STORAGES instead.`

This is due to the use of the `django.core.files.storage.get_storage_class()` function which has been deprecated.

I will submit a patch and pull request to resolve this issue.

radwon:
In Django 4.2 the `django.core.files.storage.get_storage_class()` function is deprecated as well as the `STATICFILES_STORAGE` setting in favor of `STORAGES["staticfiles"]`.

Use `django.core.files.storage.storages` to get the configured storage class for static files instead.

For Django versions prior to 4.2 keep using the `django.core.files.storage.get_storage_class()` function for backwards compatibility.

Fixes #1758

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -3,7 +3,6 @@ from os.path import join, normpath
 from django.conf import settings
 from django.contrib.staticfiles import finders, storage
 from django.core.checks import Warning
-from django.core.files.storage import get_storage_class
 from django.utils.functional import LazyObject
 from django.utils.translation import gettext_lazy as _, ngettext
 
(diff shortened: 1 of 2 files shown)


matthiask (approved):
I think this looks good. I also think that the existing tests already cover "enough" of the behavior so that new tests aren't strictly necessary. A note in `docs/changes.rst` would be great, though!

radwon:
I've added a note in `docs/changes.rst` to include the update.

matthiask:
Thanks!

PR #1760, opened 2023-04-10
Add theming support

### LINKED ISSUE #1757: Make font families configurable
Currently it's not that simple to configure font families due to the way `toolbar.css` is written. It would be great if at least the font families were configurable via CSS variables (similar to how the Django admin site does it: https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#theming-support). The existing `css` block already allows us to inject custom CSS, so we could use the same technique if the style sheet was using variables.

This is mostly to be able to make the look of the toolbar consistent across different systems.

GergelyKalmar:
# Description

This PR adds theming support to the Django debug toolbar that works the same way as https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#theming-support.

Fixes #1757.

# Checklist:

- [x] I have added the relevant tests for this change. (Existing tests should be sufficient as this is just a simple refactoring at this point.)

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -1,3 +1,38 @@
+/* Varible definitions */
+:root {
+    --font-family-primary: /* Same as in Django admin/css/base.css */
+        -apple-system,
+        BlinkMacSystemFont,
+        "Segoe UI",
+        system-ui,
+        Roboto,
+        "Helvetica Neue",
+        Arial,
+        sans-serif,
+        "Apple Color Emoji",
+        "Segoe UI Emoji",
+        "Segoe UI Symbol",
+        "Noto Color Emoji";
+    --font-family-monospace:  /* Same as in Django admin/css/base.css */
+        ui-monospace,
+        Menlo,
+        Monaco,
+        "Cascadia Mono",
+        "Segoe UI Mono",
+        "Roboto Mono",
+        "Oxygen Mono",
+        "Ubuntu Monospace",
+        "Source Code Pro",
+        "Fira Mono",
+        "Droid Sans Mono",
+        "Courier New",
+        monospace,
+        "Apple Color Emoji",
+        "Segoe UI Emoji",
+        "Segoe UI Symbol",
+        "Noto Color Emoji";
+}
+
 /* Debug Toolbar CSS Reset, adapted from Eric Meyer's CSS Reset */
 #djDebug {
     color: #000;
@@ -77,9 +112,7 @@
     color: #000;
     vertical-align: baseline;
     background-color: transparent;
-    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", system-ui,
-        Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji",
-        "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+    font-family: var(--font-family-primary);
     text-align: left;
     text-shadow: none;
     white-space: normal;
@@ -179,7 +212,7 @@
 
 #djDebug #djDebugToolbar li.djdt-active:before {
     content: "▶";
-    font-family: sans-serif;
+    font-family: var(--font-family-primary);
     position: absolute;
     left: 0;
     top: 50%;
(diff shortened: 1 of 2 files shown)


GergelyKalmar:
@matthiask Any chance to get this reviewed, merged and released? The failing tests seems to be unrelated flakiness.

tim-schilling:
Thank you for the PR. Given the context that this is a development tool, who is this particular change meant for and what problem(s) is it solving for them?

GergelyKalmar:
There are more details in the related issue: https://github.com/jazzband/django-debug-toolbar/issues/1757.

In my case I'm developing a library that pre-configures Django Debug Toolbar in a certain way. The tests include screenshots with the toolbar open, therefore it would be important to ensure that the font family is consistent (to ensure that the screenshots are the same regardless of the developer's operating system).

I acknowledge that this might be a relatively niche use case, but I think adding this feature should not hurt nor should it cause unnecessary maintenance burden. We're bringing the toolbar closer to how Django handles this as well (e.g. for the admin site). Additionally, this feature could be beneficial for plugin developers as well (in case they want to add screenshots to their tests).

tim-schilling:
Thank you for the explanation! I think this PR needs to include the documentation on how to customize/utilize the theming support, along with explanations on what can be included in that theme (I assume that's a shorter list than what can't be).

GergelyKalmar:
Alright, I'll extend the documentation, however, it will be pretty limited with only the font families at the moment.

GergelyKalmar:
Documentation is extended now with a section on theming support. As the process is basically the same as for the Django admin site (see https://docs.djangoproject.com/en/4.2/ref/contrib/admin/#theming-support) I also documented it very similarly.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -1,3 +1,17 @@
+/* Varible definitions */
```
matthiask, line 1:
```suggestion
/* Variable definitions */
```

Typo

debug_toolbar/templates/debug_toolbar/base.html
```diff
@@ -3,6 +3,7 @@
 <link rel="stylesheet" href="{% static 'debug_toolbar/css/print.css' %}" media="print">
 <link rel="stylesheet" href="{% static 'debug_toolbar/css/toolbar.css' %}">
 {% endblock %}
+{% block extrastyle %}{% endblock %}
```
matthiask, line 6:
```suggestion
```

docs/configuration.rst
```diff
@@ -338,3 +338,26 @@ Here's what a slightly customized toolbar configuration might look like::
...
+The debug toolbar uses CSS variables to define fonts. This allows changing
+fonts without having to override many individual CSS rules. For example, if
+you preferred Roboto instead of the default list of fonts you could add a
+**debug_toolbar/base.html** template override to your project:
+
+.. code-block:: django
+
+    {% extends 'debug_toolbar/base.html' %}
+
+    {% block extrastyle %}{{ block.super }}
```
matthiask, line 353:
```suggestion
    {% block css %}{{ block.super }}
```

matthiask (approved):
Thank you! 

I found a typo; that should be fixed. The removal of the `extrastyle` block is a stylistic choice; you don't have to follow that. I don't see why the additional block is necessary when you're using `{{ block.super }}` anyway, but that's just me.

Now that I took a closer look I worry minimally that we use a generic sounding CSS property for the font family list. I'm quite sure it's safe enough to try but I wanted to raise the question if we shouldn't use `--djdt`-prefixed properties. I think I'm slightly against it since for example `--font-family-monospace` really should be a monospace font and we cannot hinder people from shooting themselves in the foot (or maybe rather applying lipstick to their foot or something)

matthiask:
The spell check failed:

```
writing output... [ 44%] configuration
configuration.rst:343: : Spell check: Theming: Theming support.
configuration.rst:346: : Spell check: Roboto: you preferred Roboto instead of the default list of fonts you could add a.
Writing /home/runner/work/django-debug-toolbar/django-debug-toolbar/docs/_build/spelling/configuration.spelling
writing output... [ 55%] contributing
```

The psycopg3 failures are not related and can be ignored for this change, but you should fix the spell check test. Thanks!

tim-schilling:
Thank you @GergelyKalmar! Good work all around.

GergelyKalmar:
Thank you all for the support! Glad to see this go through.

PR #1763, opened 2023-04-13
add Redis history

ivanyshchuk:
# Description

this information as well.

Fixes # (issue)

# Checklist:


matthiask:
Thank you! If I understand it correctly you want to persist the toolbar's state in the cache instead of the local memory, correct? We have been discussing approaches here: https://github.com/jazzband/django-debug-toolbar/pull/1432

matthiask:
I'm going to reject this as it is currently since we have discussed a slightly different approach which we want to pursue.

Also, tests are failing and there hasn't been any response to my comments. 

It's not a rejection of the feature, just a rejection of this particular pull request. Thanks!

PR #1765, opened 2023-04-22
D9pouces wsgi.multiprocess 1

d9pouces:
# Description

If you apply Django middlewares on a HttpRequest without "wsgi.multiprocess" in META, an exception is raised by DJT. 
This simple patch avoids this bug.

Fixes # (issue)

# Checklist:

### DIFF
--- a/debug_toolbar/toolbar.py
+++ b/debug_toolbar/toolbar.py
@@ -95,9 +95,13 @@ class DebugToolbar:
 
         If False, the panels will be loaded via Ajax.
         """
-        render_panels = self.config["RENDER_PANELS"]
-        if render_panels is None:
-            render_panels = self.request.META.get("wsgi.multiprocess", False)
+        if (render_panels := self.config["RENDER_PANELS"]) is None:
+            # If wsgi.multiprocess isn't in the headers, then it's likely
+            # being served by ASGI. This type of set up is most likely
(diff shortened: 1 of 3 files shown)


tim-schilling:
Thank you for the PR. I agree that this change is straightforward and a +1 from me.

Could you please add a test for this, add a line to the changelog and include a description of how this error occurs within Django Channels? All of that will help the toolbar be easier to maintain and more robust moving forward.

tim-schilling:
@d9pouces could you provide us with the use case that causes this bug?

tim-schilling:
@matthiask I think `should_render_panels` should return True if `wsgi.multiprocess` is not in the headers. The reason being that an asgi application is more likely to be running in an async or similar process that's not going to be compatible.

matthiask:
@tim-schilling I checked, PEP-3333 specifies that `wsgi.multiprocess` must exist when running under WSGI. https://peps.python.org/pep-3333/

You're saying that the better fallback, for example when running under ASGI, would be to render panels right away? I think I'm in agreement.

tim-schilling:
Yes, that's what I'm saying. A person can still find a way to get this to work via the `RENDER_PANELS` setting.

I'll push my changes to this branch over lunch today. The other change is including a comment and some tests for this particular method.

PR #1770, opened 2023-05-08
Don't try to undo cache method monkey patching

### LINKED ISSUE #1769: After upgrading to Django 4.2 LTS, I receive AttributeError on each page load. Using DDT 4.0.0
```
E0506 14:25:32.246 django.request:241 Internal Server Error: /dashboard/
  File "C:\Users\Owner\.virtualenvs\premind-6P0nJig8\lib\site-packages\django\core\handlers\exception.py", line 55, in inner
  File "C:\Users\Owner\.virtualenvs\premind-6P0nJig8\lib\site-packages\sentry_sdk\integrations\django\middleware.py", line 176, in __call__
    return f(*args, **kwargs)
  File "C:\Users\Owner\.virtualenvs\premind-6P0nJig8\lib\site-packages\debug_toolbar\middleware.py", line 64, in __call__
  File "C:\Users\Owner\.virtualenvs\premind-6P0nJig8\lib\site-packages\debug_toolbar\panels\cache.py", line 217, in disable_instrumentation
    self._unmonkey_patch_cache(cache)
  File "C:\Users\Owner\.virtualenvs\premind-6P0nJig8\lib\site-packages\debug_toolbar\panels\cache.py", line 171, in _unmonkey_patch_cache
    if original_method.__func__ == getattr(cache.__class__, name):
AttributeError: 'function' object has no attribute '__func__'
```

Commenting `debug_toolbar.panels.cache.CachePanel` allows DDT to continue to work.
```python
# settings.py
MIDDLEWARE = [
    "django.middleware.gzip.GZipMiddleware",
]

if DEBUG_TOOLBAR:
    INSTALLED_APPS += ["debug_toolbar"]
    MIDDLEWARE += ["debug_toolbar.middleware.DebugToolbarMiddleware"]
    DEBUG_TOOLBAR_PANELS = [
        "debug_toolbar.panels.history.HistoryPanel",
        "debug_toolbar.panels.versions.VersionsPanel",
        "debug_toolbar.panels.timer.TimerPanel",
        "debug_toolbar.panels.settings.SettingsPanel",
        "debug_toolbar.panels.headers.HeadersPanel",
        "debug_toolbar.panels.request.RequestPanel",
        "debug_toolbar.panels.sql.SQLPanel",
        # 'debug_toolbar.panels.staticfiles.StaticFilesPanel',  # breaks if minio is enabled
        "debug_toolbar.panels.templates.TemplatesPanel",
        # "debug_toolbar.panels.cache.CachePanel",  # commenting this line allows DDT to continue to work
        "debug_toolbar.panels.logging.LoggingPanel",
    ]
```
Please let me know if you need any more information in order to solve this issue

living180:
# Description

Trying to undo the monkey patch of cache methods in the `CachePanel.disable_instrumentation()` method is fragile in the presence of other code which may also monkey patch the same methods (such as Sentry's Django integration), and there are theoretically situations where it is actually impossible to do correctly.  Thus once a cache has been monkey-patched, leave it that way, and instead rely on checking in the patched methods to see if recording needs to happen.  This is done via a `_djdt_panel` attribute which is set to the current panel in the `enable_instrumentation()` method and then set to `None` in the `disable_instrumentation()` method.

Fixes #1769

# Checklist:

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -30,6 +30,27 @@ WRAPPED_CACHE_METHODS = [
 ]
 
 
+def _monkey_patch_method(cache, name):
+    original_method = getattr(cache, name)
+
+    @functools.wraps(original_method)
+    def wrapper(*args, **kwargs):
+        panel = cache._djdt_panel
+        if panel is None:
+            return original_method(*args, **kwargs)
+        else:
+            return panel._record_call(cache, name, original_method, args, kwargs)
+
+    setattr(cache, name, wrapper)
+
+
+def _monkey_patch_cache(cache):
+    if not hasattr(cache, "_djdt_patched"):
+        for name in WRAPPED_CACHE_METHODS:
+            _monkey_patch_method(cache, name)
+        cache._djdt_patched = True
+
+
 class CachePanel(Panel):
     """
     Panel that displays the cache statistics.
@@ -72,7 +93,8 @@ class CachePanel(Panel):
                 cache = original_method(self, alias)
                 panel = cls.current_instance()
                 if panel is not None:
-                    panel._monkey_patch_cache(cache)
+                    _monkey_patch_cache(cache)
+                    cache._djdt_panel = panel
                 return cache
 
             CacheHandler.create_connection = wrapper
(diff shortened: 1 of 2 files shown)


matthiask (approved):
I think it's a good idea. I haven't tested it out, it would certainly be helpful if the change were tested by those who reported crashes in the other issue. 

Sorry for the bikeshedding, but maybe `_djdt_recording` was a more fitting name than `_djdt_panel`; or maybe I misunderstand the intention of the variable?

I also think we should generally move towards using `from time import perf_counter` instead of `time.time()` to measure short durations; my understanding is that `time.time` could be affected by updating the system clock. It's extremely unlikely to matter of course, but `perf_counter` would be more correct. And it's also what Django does, see https://github.com/django/django/blob/1586a09b7949bbb7b0d84cb74ce1cadc25cbb355/django/test/utils.py#L934-L941

living180:
> Sorry for the bikeshedding, but maybe `_djdt_recording` was a more fitting name than `_djdt_panel`; or maybe I misunderstand the intention of the variable?

So the type of the variable actually changed:  previously it was a `bool`, now it is an instance of `CachePanel` if instrumentation is enabled, or `None` otherwise.  This is necessary because with this PR, once a cache has been monkey patched, it stays that way, but each request creates a new `CachePanel` (previously the cache was being monkey patched for each request specifically for the associated `CachePanel`).

living180:
> I also think we should generally move towards using `from time import perf_counter` instead of `time.time()` to measure short durations; my understanding is that `time.time` could be affected by updating the system clock. It's extremely unlikely to matter of course, but `perf_counter` would be more correct. And it's also what Django does, see https://github.com/django/django/blob/1586a09b7949bbb7b0d84cb74ce1cadc25cbb355/django/test/utils.py#L934-L941

Totally agree, though that is really orthogonal to this change.  I can look at making a new PR switching to `perf_counter` across the codebase (I'm pretty sure the SQL panel would also need to be updated, and perhaps others).

matthiask:
> So the type of the variable actually changed: previously it was a `bool`, now it is an instance of `CachePanel` if instrumentation is enabled, or `None` otherwise. This is necessary because with this PR, once a cache has been monkey patched, it stays that way, but each request creates a new `CachePanel` (previously the cache was being monkey patched for each request specifically for the associated `CachePanel`).

Ah of course, thank you!

PR #1772, opened 2023-05-09
Fix #1711 Improve installation page formats

pauloxnet:
# Description

Improve installation page formats

Fixes #1711


matthiask:
Thank you!

PR #1773, opened 2023-05-09
Fix psycopg3 tests

living180:
# Description

Several tests (such as `SQLPanelTestCase.test_cursor_wrapper_singleton`) are written to ensure that only a single cursor wrapper is instantiated during a test.  However, this fails when using psycopg3, since the `.last_executed_query()` call in `NormalCursorWrapper._record()` ends up creating an additional cursor (via [the `mogrify()` function](https://github.com/django/django/blob/fea17b76150688056d78818ea1ef47f1122dc165/django/db/backends/postgresql/psycopg_any.py#L21)).  To avoid this, use a `._djdt_in_record` attribute on the database wrapper.  Make the `NormalCursorWrapper._record()` method set `._djdt_in_record` to `True` on entry and reset it to `False` on exit.  Then in the overridden database wrapper `.cursor()` and `.chunked_cursor()` methods, check the `._djdt_in_record` attribute and return the original cursor without wrapping if the attribute is `True`.

# Checklist:

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -144,6 +144,21 @@ class NormalCursorWrapper(BaseCursorWrapper):
         except UnicodeDecodeError:
             return "(encoded string)"
 
+    def _last_executed_query(self, sql, params):
+        """Get the last executed query from the connection."""
+        # Django's psycopg3 backend creates a new cursor in its implementation of the
+        # .last_executed_query() method.  To avoid wrapping that cursor, temporarily set
+        # the DatabaseWrapper's ._djdt_logger attribute to None.  This will cause the
+        # monkey-patched .cursor() and .chunked_cursor() methods to skip the wrapping
+        # process during the .last_executed_query() call.
+        self.db._djdt_logger = None
+        try:
+            return self.db.ops.last_executed_query(
+                self.cursor, sql, self._quote_params(params)
+            )
+        finally:
+            self.db._djdt_logger = self.logger
+
     def _record(self, method, sql, params):
         alias = self.db.alias
         vendor = self.db.vendor
@@ -176,9 +191,7 @@ class NormalCursorWrapper(BaseCursorWrapper):
             params = {
                 "vendor": vendor,
                 "alias": alias,
-                "sql": self.db.ops.last_executed_query(
-                    self.cursor, sql, self._quote_params(params)
-                ),
+                "sql": self._last_executed_query(sql, params),
                 "duration": duration,
                 "raw_sql": sql,
                 "params": _params,
@@ -186,7 +199,9 @@ class NormalCursorWrapper(BaseCursorWrapper):
                 "stacktrace": get_stack_trace(skip=2),
                 "start_time": start_time,
                 "stop_time": stop_time,
-                "is_slow": duration > dt_settings.get_config()["SQL_WARNING_THRESHOLD"],
+                "is_slow": (
+                    duration > dt_settings.get_config()["SQL_WARNING_THRESHOLD"]
+                ),
                 "is_select": sql.lower().strip().startswith("select"),
                 "template_info": template_info,
             }
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -63,7 +63,6 @@ def wrap_cursor(connection, panel):
 
         connection.cursor = cursor
         connection.chunked_cursor = chunked_cursor
-        return cursor
 
 
 def unwrap_cursor(connection):
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -3,6 +3,7 @@ import datetime
 import json
 from time import time
 
+import django.test.testcases
 from django.utils.encoding import force_str
 
 from debug_toolbar import settings as dt_settings
@@ -31,10 +32,15 @@ class SQLQueryTriggered(Exception):
     """Thrown when template panel triggers a query"""
 
 
-def wrap_cursor(connection, panel):
+def wrap_cursor(connection):
+    # If running a Django SimpleTestCase, which isn't allowed to access the database,
+    # don't perform any monkey patching.
+    if isinstance(connection.cursor, django.test.testcases._DatabaseFailure):
+        return
     if not hasattr(connection, "_djdt_cursor"):
         connection._djdt_cursor = connection.cursor
         connection._djdt_chunked_cursor = connection.chunked_cursor
+        connection._djdt_logger = None
 
         def cursor(*args, **kwargs):
             # Per the DB API cursor() does not accept any arguments. There's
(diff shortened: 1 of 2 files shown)


tim-schilling:
Thank you for diving into this. I kept trying to make time over the weekend, but got sidetracked each time.

matthiask (commented):
(Btw, https://github.com/jazzband/django-debug-toolbar/pull/1773/files?w=1 is much easier to review.)

Thanks for diving into this! Tests are finally green again, so that's definitely a good sign.

I have to admit: It's not obvious that the code does the right thing. I'm really wondering if there isn't a scenario where it's possible to pass some object to `cursor.execute` which fetches its own cursor during the preparation or execution stage of the SQL query in psycopg. Are we sure that there are no legitimate uses of instantiating a new cursor inside `execute`, `executemany` and friends? Because if that's the case we should probably change the assertion depending on whether we're using psycopg2 or psycopg3.

(I did some research but I didn't find an answer to this question. I'd certainly be unsure again when revisiting this code in a few months or years.)

tim-schilling:
@living180 and @matthiask I took a look at this. Originally I was just going to swap out the decorated function with another ContextVar. Then I started digging into the issue and felt like we could reasonably re-use `allow_sql` here if we allow `ExceptionCursorWrapper` to access `connection` and call `self.close()`. This was achievable (for 4.2 at least) by utilizing Django's `CursorWrapper` class. And rather than writing an exception case for those two members, I swapped it for an explicit deny list of fields.

I didn't test for other pythons, and djangos. Only 4.2 with psycopg3 and sqlite, we'll see what CI says.

tim-schilling:
Hot dog! Let me know what you think. It's pretty late for me right now.

living180:
Hey @tim-schilling I definitely like the approach of subclassing Django's `CursorWrapper` instead of proxying.  A couple of suggestions though:  I would rename our `BaseCursorWrapper` to something like `DjDTCursorWrapper` since it is no longer the actual base class.  Second, I think your original idea of using an allow list in `ExceptionCursorWrapper` to permit `.connection` and `.close()` are a better approach than a deny list.  With the allow list approach, if an additional attribute is accessed that needs to be allowed, the result will be a very visible exception that will immediately show up in testing and can be addressed.  However, with the deny list approach, if there happens to be a `CursorWrapper` method that does access the DB that isn't denied, it could silently allow the `TemplatesPanel._store_template_info()` method to access the DB when it shouldn't, and likely wouldn't be noticed for much longer.

That said, I still think it might be better to simply not wrap the cursor created as a result of calling `.last_executed_query()` instead of trying to force `ExceptionCursorWrapper` to fit that case, using a similar approach to my original commit.

living180:
So after looking at the code a bit more, I tried a slightly different approach that builds on no longer undoing the SQL panel monkey patching, similar to the change that I just made to the cache panel:  https://github.com/jazzband/django-debug-toolbar/compare/main...living180:django-debug-toolbar:psycopg3_tests_v2?w=1.

tim-schilling:
I don't agree with the earlier decision with the cache panel and now this to not undo the monkey patching on the functionality. If we want to go that route, we should work through the justification to removing the "disable_instrumentation" method.

To be clear, I'm fine changing how the toolbar works internally, but would like to see that decision to change the toolbar's future made via a discussion.

tim-schilling:
@living180 Can you make this work via a `ContextVar`. My justification is that it's easier to reason about in that context than manipulating a private attribute across various functions. `self.db._djdt_in_record` isn't obvious that `self.db` our wrapped cursor. If you disagree, please let me know why you feel an attribute on the function is the better approach.

I'd also be curious about your thoughts on moving the setting of that context var as close as possible to `last_executed_query()` since we know that's what triggers it.

matthiask:
> I don't agree with the earlier decision with the cache panel and now this to not undo the monkey patching on the functionality. If we want to go that route, we should work through the justification to removing the "disable_instrumentation" method.
> 
> To be clear, I'm fine changing how the toolbar works internally, but would like to see that decision to change the toolbar's future made via a discussion.

I'm sorry for merging the change too quickly then.

For the record: I don't think removing the `disable_instrumentation` method would be a good idea. Let's not get sidetracked here (too much) and continue in another issue if there's much discuss -- I'm happy to participate. 

I want to say this though: Reversing some types of monkey patching is brittle, especially when some other code also applies its own monkey patches. I'm sure that the performance hit of an additional method call (added by the instrumentation) is negligible and only really harms those who are already putting themselves in harms way by installing the toolbar in production environments. It's better in the long run to keep the amount of patching back and forth low.

tim-schilling:
You're right. I was extrapolating too much, going from undoing the monkey patching to removing `disable_instrumentation`.

It would be nice to find a re-usable pattern/abstraction for this type of logic in these monkey-patched panels.

living180:
> I don't agree with the earlier decision with the cache panel and now this to not undo the monkey patching on the functionality. If we want to go that route, we should work through the justification to removing the "disable_instrumentation" method.

I'd love to hear more about your concerns.  The reason I went with the approach I used with the cache panel was based on @matthiask's comment:  https://github.com/jazzband/django-debug-toolbar/issues/1769#issuecomment-1538248434.  And indeed, after looking at how that code worked previously, I realized that in the general case it is not possible to safely undo monkey patching in general:  consider if another piece of code monkey patches the cache _after_ our monkey patch is applied.  Then if we try to undo our monkey patch, it will also inadvertently undo the other code's monkey patch as well, which is probably not what is intended.  Now, I don't know of any case of that happening in practice, so it might be the case that your concerns override this change.

tim-schilling:
The concern I have is that we would be running code in someone's application when we and the user don't expect the toolbar to be running code. I reviewed both your changes and it looks like this is prevented so my concern is already being addressed.

living180:
> @living180 Can you make this work via a `ContextVar`. My justification is that it's easier to reason about in that context than manipulating a private attribute across various functions. `self.db._djdt_in_record` isn't obvious that `self.db` our wrapped cursor. If you disagree, please let me know why you feel an attribute on the function is the better approach.

I suppose it is mainly a matter of preference.  A `ContextVar` is a semi-global variable ("semi-" because it _is_ scoped to the current thread / task), while the attribute is clearly associated with the object it is affecting.  I agree that it isn't immediately obvious that the the cursor's `.db` attribute is the same as the `connection` variable passed to `wrap_cursor()`.  If we go with the attribute approach, I'm happy to expand the comment to help clarify that.  (As an aside, part of the lack of clarity comes from Django's own inconsistent naming scheme, which we (rightfully, IMO) follow in the SQL tracking code.  The object with the `.cursor()` and `.chunked_cursor()` methods is an instance of Django's `DatabaseWrapper`, which the Django codebase usually calls `connection`, but also sometimes `db`.  To further muddy the waters a `DatabaseWrapper` instance has a `connection` attribute itself.  It was fun sorting through all of that when fixing the Debug Toolbar PostgreSQL transaction tracking code.)

> I'd also be curious about your thoughts on moving the setting of that context var as close as possible to `last_executed_query()` since we know that's what triggers it.

That sounds fine.

tim-schilling:
> I suppose it is mainly a matter of preference. A ContextVar is a semi-global variable ("semi-" because it is scoped to the current thread / task), while the attribute is clearly associated with the object it is affecting.

You've convinced me that my approach may not be the better one. I'm happy with whatever you decide to go with.

living180:
So I've updated this PR to be based on approach I mentioned above:  https://github.com/jazzband/django-debug-toolbar/pull/1773#issuecomment-1542123342.  @tim-schilling I borrowed your approach of adding a `_last_executed_query()` method to encapsulate the process of disabling and then re-enabling the cursor wrapping.  I also kept the portion of your commit that changed the Debug Toolbar cursor wrappers to inherit from Django's `CursorWrapper`.

If there are any further concerns or feedback please do let me know.  I will be AFK for the next couple of days but can respond to any comments on Sunday.

living180:
Made one more minor change to avoid wrapping a `CursorWrapper` in a `CursorWrapper`:  https://github.com/jazzband/django-debug-toolbar/compare/ee46d62e8d6ac34e4884d00a8a768091172b6449..e7575e87dc9e2d2560b87d6fd5a123b9398cbd34

debug_toolbar/panels/sql/tracking.py
```diff
@@ -31,10 +33,15 @@ class SQLQueryTriggered(Exception):
     """Thrown when template panel triggers a query"""
 
 
-def wrap_cursor(connection, panel):
+def wrap_cursor(connection):
+    # If running a Django SimpleTestCase, which isn't allowed to access the database,
+    # don't perform any monkey patching.
```
tim-schilling, line 38: We should include the "Why" part of the context from the other comment. It helps me understand the justification and purpose of the code.


        # Sometimes the cursor()/chunked_cursor() methods of the DatabaseWrapper
        # instance are already monkey patched before wrap_cursor() is called.  (In
        # particular, Django's SimpleTestCase monkey patches those methods for any
        # disallowed databases to raise an exception if they are accessed.)

tim-schilling (approved):
This looks great, thank you @living180! I'm going to make the comment change because I think there's value in it.

matthiask:
Thank you both! Looks good. Let's continue with it and see if we can roll a new release soon-ish.

debug_toolbar/panels/sql/tracking.py
living180: @tim-schilling I think you are right that expanding the comment a bit more could be helpful.  I've made an initial attempt [here](https://github.com/jazzband/django-debug-toolbar/commit/d51611be46ae6a09152881381810f559ebad8d2a).  If you think that is an improvement, let me know and I can make it a proper PR.

PR #1777, opened 2023-05-15
Replace time.time() with time.perf_counter()

living180:
# Description

`time.perf_counter()` is guaranteed to use the highest-resolution clock available, and is monotonically increasing, neither of which are true for `time.time()`.  Thanks @matthiask for the suggestion!

# Checklist:


matthiask:
Thank you!

PR #1781, opened 2023-05-16
Use ruff for linting

WhyNotHugo:
Use ruff instead of flake8+isort+pyupgrade+pygrep-hooks. This change configures ruff to apply all the same rules that these tools apply.

I've omitted E501 for now since there are a few violations.


WhyNotHugo:
My intent is to enable a few extra rules, but I've purposefully limited this initial patch to apply the same rules as before.

matthiask (changes requested):
Two minor points:

* My experience is better when running ruff before black.
* Is there a reason why you added ruff to `requirements_dev.txt`? I don't think it's necessary? Pruning flake8 and isort looks like the right thing to do anway, that has probably been forgotten when we switched to pre-commit.

Apart from that I'm very happy with ruff, and I agree with this

.pre-commit-config.yaml
```diff
@@ -7,37 +7,15 @@ repos:
...
-    -   id: isort
--   repo: https://github.com/pre-commit/pygrep-hooks
-    rev: v1.10.0
-    hooks:
-    -   id: python-check-blanket-noqa
-    -   id: python-check-mock-methods
-    -   id: python-no-eval
-    -   id: python-no-log-warn
-    -   id: rst-backticks
-    -   id: rst-directive-colons
```
matthiask, line 40: ruff doesn't handle these two hooks. Is there a good reason for removing them?

WhyNotHugo: My bad, I had the impression that ruff covered all of them. Fixed.

WhyNotHugo:
> My experience is better when running ruff before black.

I think that makes sense, I've inverted the order.

> Is there a reason why you added ruff to requirements_dev.txt? I don't think it's necessary? Pruning flake8 and isort looks like the right thing to do anway, that has probably been forgotten when we switched to pre-commit.

It was mostly to stay in line with the existing requirements. I'll go ahead and drop it; I think it's sane to expect that developers have this installed already (or know how to install it).

tim-schilling:
It's probably worth dropping a line in the changes.rst file about this too.

matthiask (approved):

pauloxnet (approved):

WhyNotHugo:
> It's probably worth dropping a line in the changes.rst file about this too.

This really doesn't have any user-facing changes, I'm not sure what to mention for this.

PR #1782, opened 2023-05-16
Minor testing cleanups

living180:
Make some minor tweaks to the `tox.ini` file (the only functional change is enabling selenium tests for Python 3.11/Django 4.2 instead of for Python 3.10/Django 4.1), and fix a few warnings generated by the tests.

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -21,7 +21,7 @@ except ImportError:
     psycopg = None
 
 from ..base import BaseMultiDBTestCase, BaseTestCase
-from ..models import PostgresJSON
+from ..models import Binary, PostgresJSON
 
 
 def sql_call(use_iterator=False):
@@ -149,7 +149,7 @@ class SQLPanelTestCase(BaseTestCase):
         self.assertEqual(len(self.panel._queries), 2)
 
         # non-ASCII bytes parameters
-        list(User.objects.filter(username="café".encode()))
+        list(Binary.objects.filter(field__in=["café".encode()]))
         self.assertEqual(len(self.panel._queries), 3)
 
         response = self.panel.process_request(self.request)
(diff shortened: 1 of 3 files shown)


matthiask (approved):
apt-get doesn't like us today it seems.

tox.ini
```diff
@@ -3,28 +3,27 @@ isolated_build = true
...
     packaging
-    py{38,39,310}-dj{32}-{sqlite,postgresql,postgis,mysql}
-    py{310}-dj{40}-{sqlite}
-    py{310,311}-dj{41}-{sqlite,postgresql,postgis,mysql}
-    py{310,311}-dj{42,main}-{sqlite,postgresql,postgis,mysql}
-    py{310,311}-dj{42,main}-psycopg3
+    py{38,39,310}-dj32-{sqlite,postgresql,postgis,mysql}
+    py310-dj40-sqlite
+    py{310,311}-dj41-{sqlite,postgresql,postgis,mysql}
+    py{310,311}-dj{42,main}-{sqlite,postgresql,psycopg3,postgis,mysql}
```
tim-schilling, line 9: Thank you for cleaning this up

tox.ini
```diff
@@ -40,7 +39,7 @@ setenv =
     PYTHONPATH = {toxinidir}
     PYTHONWARNINGS = d
     py39-dj32-postgresql: DJANGO_SELENIUM_TESTS = true
-    py310-dj41-postgresql: DJANGO_SELENIUM_TESTS = true
+    py311-dj42-postgresql: DJANGO_SELENIUM_TESTS = true
```
tim-schilling, line 42: Good catch

tests/panels/test_sql.py
```diff
@@ -149,7 +149,7 @@ def test_non_ascii_query(self):
         self.assertEqual(len(self.panel._queries), 2)
 
         # non-ASCII bytes parameters
-        list(User.objects.filter(username="café".encode()))
+        list(Binary.objects.filter(field__in=["café".encode()]))
```
tim-schilling, line 152: What's the purpose behind changing this from a `bytes` to a list of `bytes`?

living180: I found that even with `BinaryField`, using `.filter(field="café")` would generate a `BytesWarning` internally within Django.  Not wanting to shave that particular yak, I found that switching to the form above allowed the test to proceed without any `BytesWarning`.

matthiask:
Thanks!

PR #1784, opened 2023-05-17
Add the Store API and initial documentation.

tim-schilling:
# Description

Initial commit for https://github.com/orgs/jazzband/projects/9/views/1 (serializable toolbar).

This starts the process of creating a serializable branch in the repo for that work. I'm not sure we want to merge this work alongside the working toolbar. 

This is based on the work from #1432

# Checklist:

### DIFF
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -1,0 +1,127 @@
+import json
+from collections import defaultdict, deque
+from typing import Any, Dict, Iterable
+
+from django.core.serializers.json import DjangoJSONEncoder
+from django.utils.encoding import force_str
+from django.utils.module_loading import import_string
+
+from debug_toolbar import settings as dt_settings
+
+
+class DebugToolbarJSONEncoder(DjangoJSONEncoder):
+    def default(self, o: Any) -> Any:
+        try:
+            return super().default(o)
+        except TypeError:
+            return force_str(o)
+
+
+def serialize(data: Any) -> str:
+    return json.dumps(data, cls=DebugToolbarJSONEncoder)
+
+
+def deserialize(data: str) -> Any:
+    return json.loads(data)
+
+
+class BaseStore:
+    _config = dt_settings.get_config().copy()
+
+    @classmethod
+    def ids(cls) -> Iterable:
+        """The stored ids"""
+        raise NotImplementedError
+
+    @classmethod
+    def exists(cls, request_id: str) -> bool:
+        """Does the given request_id exist in the store"""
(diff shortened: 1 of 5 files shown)


debug_toolbar/store.py
```diff
@@ -0,0 +1,127 @@
...
+
+from debug_toolbar import settings as dt_settings
+
+
+class DebugToolbarJSONEncoder(DjangoJSONEncoder):
+    def default(self, o: Any) -> Any:
+        try:
+            return super().default(o)
+        except TypeError:
+            return force_str(o)
```
matthiask, line 17: I'm wondering if this is a good default for now or if we should rather let an eventual exception bubble up so that panels are forced to only save primitive data types?

debug_toolbar/store.py
```diff
@@ -0,0 +1,127 @@
...
+
+def deserialize(data: str) -> Any:
+    return json.loads(data)
+
+
+class BaseStore:
+    _config = dt_settings.get_config().copy()
+
+    @classmethod
+    def ids(cls) -> Iterable:
```
matthiask, line 32: Maybe request_ids to make the name a little bit less generic?

docs/changes.rst
```diff
@@ -4,6 +4,9 @@ Change log
 Pending
 -------
 
+* Defines the ``BaseStore`` interface for request storage mechanisms.
```
matthiask, line 7:
```suggestion
* Defined the ``BaseStore`` interface for request storage mechanisms.
```

matthiask (approved):
Thanks for starting this!

debug_toolbar/store.py
tim-schilling: I think we're going to need this because we can't let an exception bubble up. We're going to need to store a snapshot of a template's context which can include any type.

debug_toolbar/store.py
tim-schilling: That's a good idea.

debug_toolbar/store.py
matthiask: I'm not 100% convinced that silently stringifying is a good idea, but your point makes sense.

matthiask:
Sorry, I didn't mean to close this.

debug_toolbar/store.py
tim-schilling: I removed this with a note in the code for future reference. I decided it's better to have less code and make changes than to plan for any possibility.

matthiask:
Thanks!

PR #1785, opened 2023-05-17
Include all files in sdist archives

mgorny:
# Description

Remove the restriction on files inclduded in sdist archives in order to include documentation sources and tests there.  This is the default hatchling behavior and it is helpful to packagers (such as Linux distributions or Conda) as it permits using sdist archives to do packaging (instead of GitHub archives that are not guaranteed to be reproducible).  Most of the ordinary users will not be affected since starlette is a pure Python package and therefore pip will prefer wheels to install it everywhere.

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,12 +43,6 @@ dependencies = [
 Download = "https://pypi.org/project/django-debug-toolbar/"
 Homepage = "https://github.com/jazzband/django-debug-toolbar"
 
-[tool.hatch.build.targets.sdist]
-include = [
-    "/debug_toolbar",
-    "/CONTRIBUTING.md",
-]
-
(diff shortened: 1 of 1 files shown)


mgorny:
I haven't added it to changelog since it is primarily packager-facing change rather than developer/user-facing. If you'd like me to add it there nevertheless, please let me know.

matthiask:
Doesn't this mean that a `tests` package is installed into users virtualenvs? I'm not sure this is what we want?

mgorny:
No, it's not. The `tool.hatch.build.targets.sdist` affects only sdist archives, `tool.hatch.build.targets.wheel` control what goes into wheels and therefore what is installed.

mgorny:
Thanks!

PR #1786, opened 2023-05-17
SQL panel work

living180:
# Description

Various minor improvements and cleanups to the SQL panel code.  Individual commits should be self-explanitory.


tim-schilling (approved):
This is excellent, thank you @living180

PR #1801, opened 2023-07-01
Switch StaticFilesPanel to use ContextVar.

### LINKED ISSUE #1799: django-debug-toolbar is crashing postgresql by not closing connections
I am using Django 4.2.2, django-debug-toolbar 4.1.0, Python 3.11 and Postgresql 15.2
While developing with DEBUG=True and django-debug-toolbar activated in settings.py, after about 100 pages displayed, Postgresql is crashing with the messages.

```
FATAL: remaining connection slots are reserved for non-replication superuser connections
```

My max connexion is set at 100 on my Postgresql server. Nevertheless despite I am alone to request pages to Django ``runserver`` (there is no concurrency), the opened postgresql connections count is always rising up to the max, then the above message appears.

To check open connections I do : 

```
SELECT count(*) AS total_connections FROM pg_stat_activity;
```

If I use chrome to consult pages slowly or just a bench with ``ab -n 200`` : I have the problem anyway.
If I restart ``runserver``, I have again the right to read about 100 pages, but after it crashes.
The only solution I found is to remove django-debug-toolbar from settings.py : then ``total_connections`` is stable.

I guess django-debug-toolbar is not closing DB connections in some way : could you give me a workaround because your package is an absolute must.

tim-schilling:
The StaticFilesPanel thread collector was not closing connections to the database. This approach allows those connections to be closed while still collecting context across threads.

The test case is to hit an admin login screen with

    ab -n 200 http://localhost:8000/admin/login/

As far as I can tell, all the static files are collected properly and connections don't stay open.

Fixes #1799

# Checklist:

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -1,3 +1,5 @@
+import contextlib
+from contextvars import ContextVar
 from os.path import join, normpath
 
 from django.conf import settings
@@ -7,12 +9,6 @@ from django.utils.functional import LazyObject
 from django.utils.translation import gettext_lazy as _, ngettext
 
 from debug_toolbar import panels
-from debug_toolbar.utils import ThreadCollector
-
-try:
-    import threading
-except ImportError:
-    threading = None
 
 
 class StaticFile:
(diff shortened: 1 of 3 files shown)


matthiask (approved):
I didn't try it out, I only read the code. Using `contextvars` is a great idea.

PR #1802, opened 2023-07-01
Check JavaScript files content type.

### LINKED ISSUE #1764: Create a system check that determines if mimetypes issue exists
Our docs call out the troublsome mimetype issue, but it's pretty difficult to parse out whether it needs to be addressed for a beginner. Ideally we'd tell people when this is a problem and what to do exactly rather than rely on them to search our docs for "mimetype" when they hit a hurdle.

I suspect we should be able to load the `mimetypes` library and inspect what's returned for `.js`.

We should also look to improve the docs by indicating what the steps are or better highlight what to search for them. https://stackoverflow.com/a/64055514/1637351 shows a good example of what to do.

tim-schilling:
A common problem with windows set ups is that the registry is misconfigured when resolving the content type for a JavaScript file. This check captures this sooner to explain why the toolbar doesn't show and attempts to tell the user what to change to make it work.

Fixes #1764

# Checklist:

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -182,12 +182,8 @@ def js_mimetype_check(app_configs, **kwargs):
     """
     Check that JavaScript files are resolving to the correct content type.
     """
-    javascript_types = {
-        # Macs
-        "application/javascript",
-        # Windows
-        "text/javascript",
-    }
+    # Those types are fine. Different OSes return different types.
+    javascript_types = {"application/javascript", "text/javascript"}
     check_failed = not set(mimetypes.guess_type("toolbar.js")).intersection(
         javascript_types
     )
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -182,7 +182,8 @@ def js_mimetype_check(app_configs, **kwargs):
     """
     Check that JavaScript files are resolving to the correct content type.
     """
-    # Those types are fine. Different OSes return different types.
+    # Ideally application/javascript is returned, but text/javascript is
+    # acceptable.
     javascript_types = {"application/javascript", "text/javascript"}
     check_failed = not set(mimetypes.guess_type("toolbar.js")).intersection(
         javascript_types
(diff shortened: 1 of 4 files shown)


debug_toolbar/apps.py
```diff
@@ -174,3 +175,37 @@ def check_panels(app_configs, **kwargs):
...
     return errors
+
+
+@register()
+def js_mimetype_check(app_configs, **kwargs):
+    """
+    Check that JavaScript files are resolving to the correct content type.
+    """
+    javascript_types = {
+        # Macs
```
matthiask, line 186: Maybe `Linux, macOS` ? Or `Unices` ?

matthiask (approved):

debug_toolbar/apps.py
tim-schilling: This section confuses me. I put those comments there because on my mac, I found that mimetypes was returning application/javascript, but the windows solution was to use text/javascript.

I couldn't reproduce what windows users would actually experience for this. Or at least not easily.

tim-schilling: I installed python on my windows machine, which was easier than expected. We should check for text/javascript, but we should recommend that users use a value of application/javascript instead. I'll clean this up to reflect that.

debug_toolbar/apps.py
```diff
@@ -174,3 +175,37 @@ def check_panels(app_configs, **kwargs):
...
+        "application/javascript",
+        # Windows
+        "text/javascript",
+    }
```
matthiask, line 190: Maybe this could be made less specific, something like:

```suggestion
    # Those types are fine. Different OSes return different types.
    javascript_types = {"application/javascript", "text/javascript"}
```

I'm not sure if that would address your concerns @tim-schilling ?

tim-schilling:
@matthiask alright, if you're okay with the latest changes let's merge them.

matthiask:
Yes, thank you!

PR #1809, opened 2023-07-04
Document serialization API for panels.

### LINKED ISSUE #1806: Add and document serialize and deserialize methods Panel API.

tim-schilling:
Fixes #1806 

# Checklist:

### DIFF
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -175,6 +175,28 @@ class Panel:
         """
         return self.toolbar.stats.get(self.panel_id, {})
 
+    def deserialize_stats(self, data):
+        """
+        Deserialize stats coming from the store.
+
+        This is responsible for converting stored stats back into python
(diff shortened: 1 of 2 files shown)


matthiask:
The lint step fails with `../debug_toolbar/panels/__init__.py:docstring of debug_toolbar.panels.Panel.deserialize_stats:1: : Spell check: Deserialize: Deserialize stats coming from the store..`

I don't know what's going on with MariaDB/MySQL.

tim-schilling:
I don't think we're going to keep these API functions in. I'm discovering that I'd prefer to make the panel simply prepare the data in a serializable format rather than patch it in afterwards.

tim-schilling:
Regarding mysql, I probably need to rebase on main.

matthiask:
> I'm discovering that I'd prefer to make the panel simply prepare the data in a serializable format rather than patch it in afterwards.

Good point, makes sense to me.

tim-schilling:
Closing

PR #1812, opened 2023-07-07
Handle logging queries encoded as bytes under PostgreSQL

### LINKED ISSUE #987: Error when recording SQL queries that are byte-strings
When a byte-string SQL query is run, e.g. those produced by `psycopg2.execute_values`, I get the error `TypeError: startswith first arg must be bytes or a tuple of bytes, not str`. 

This seems to be a result of line 142 in `django-debug-toolbar/debug_toolbar/panels/sql/tracking.py`: `'is_select': sql.lower().strip().startswith('select')`
since 'select' is unicode. 

Flask-debugtoolbar [addressed this](https://github.com/mgood/flask-debugtoolbar/commit/70488fc14a387cfcae6c48811ea4b6ce68564fef) with `prefix = b'select' if isinstance(statement, bytes) else 'select'` - would that work in this case?

Lucidiot:
# Description

Unlike SQLite, psycopg supports queries encoded as bytes, and uses this in its own methods such as in [`execute_values`](https://www.psycopg.org/docs/extras.html#psycopg2.extras.execute_values). But the SQL logging only expects strings or [Composables](https://www.psycopg.org/docs/sql.html#psycopg2.sql.Composable). This was causing an `AttributeError` such as this:

```
[snip]
  File "/home/lucidiot/.virtualenvs/tmp/lib/python3.10/site-packages/psycopg2/extras.py", line 1270, in execute_values
    cur.execute(b''.join(parts))
  File "/home/lucidiot/.virtualenvs/tmp/lib/python3.10/site-packages/debug_toolbar/panels/sql/tracking.py", line 246, in execute
    return self._record(super().execute, sql, params)
  File "/home/lucidiot/.virtualenvs/tmp/lib/python3.10/site-packages/debug_toolbar/panels/sql/tracking.py", line 181, in _record
    sql = sql.as_string(conn)
AttributeError: 'bytes' object has no attribute 'as_string'
```

This adds an extra condition to decode a `bytes` instance with `.decode()` instead of `.as_string()`.

Fixes #987 

# Checklist:

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -180,7 +180,10 @@ class NormalCursorWrapper(DjDTCursorWrapper):
             # Sql might be an object (such as psycopg Composed).
             # For logging purposes, make sure it's str.
             if vendor == "postgresql" and not isinstance(sql, str):
-                sql = sql.as_string(conn)
+                if isinstance(sql, bytes):
+                    sql = sql.decode("utf-8")
+                else:
+                    sql = sql.as_string(conn)
             else:
                 sql = str(sql)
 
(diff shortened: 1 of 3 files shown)


matthiask:
Thank you!

PR #1818, opened 2023-07-31
Hacky way of getting the toolbar to work under ASGI

matthiask:
@tim-schilling I got the idea to try what happens when converting all of the toolbar to run in async mode only and let Django handle the synchronous case. Surprisingly enough almost everything just works. I don't know if we should pursuse this path but if we can let Django to the sync/async coordination we would have much much less to do.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,7 +31,7 @@ repos:
         args:
         -  --trailing-comma=es5
 -   repo: https://github.com/pre-commit/mirrors-eslint
-    rev: v8.41.0
+    rev: v8.42.0
     hooks:
     -   id: eslint
         files: \.js?$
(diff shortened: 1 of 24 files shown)


tim-schilling:
@matthiask what is all included when you say "almost everything just works"?

matthiask:
@tim-schilling I'm sorry, that was premature. I was happy that it worked at all and didn't check thoroughly.  Templates, caches, headers etc. work fine, and requests go through (responses are actually returned to the client) so that's much bette than the status quo already but not that much either.

Still, I think making the toolbar middleware async-only may be the way to go.

matthiask:
Closing this for now since I didn't have any bandwidth to continue working on this during the last few months. This probably will not change soon so I'm setting my priorities.

PR #1820, opened 2023-08-02
Patch CursorWrapper dynamically to allow multiple base classes.

### LINKED ISSUE #1814: Django logging not working when djdt is enabled
I recently noticed that django default logging is not working when I enable djdt.
django-debug-toolbar==4.1.0
Django==4.2.3

Is it possible to have SQL pannel and logging for sql queries work together?

Without djdt
Screenshot 2023-07-10 at 01 00 07

With djdt
Screenshot 2023-07-10 at 01 02 08

tim-schilling:
When the debug SQL logs are enabled, the wrapper class is CursorDebugWrapper not CursorWrapper. Since we have inspections based on that specific class they are removing the CursorDebugWrapper causing the SQL logs to not appear.

This attempts to dynamically patch the CursorWrapper or CursorDebugWrapper with the functionality we need.

This doesn't do a full regression test, but it may be possible to get it to work with:

    TEST_ARGS='--debug-sql' make test

Which causes the current set of tests to fail since they are keyed to CursorWrapper.

Fixes #1814

# Checklist:

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -59,7 +59,7 @@ def wrap_cursor(connection):
             cursor = connection._djdt_cursor(*args, **kwargs)
             if logger is None:
                 return cursor
-            mixin = NormalCursorWrapper if allow_sql.get() else ExceptionCursorWrapper
+            mixin = NormalCursorMixin if allow_sql.get() else ExceptionCursorMixin
             return patch_cursor_wrapper_with_mixin(cursor.__class__, mixin)(
                 cursor.cursor, connection, logger
             )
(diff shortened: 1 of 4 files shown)


tim-schilling:
@living180 I can't add you to the requested reviewers, but I'd appreciate you taking a look at this PR.

tim-schilling:
@matthiask sorry, can you review that last commit too. It's a bit of a hack, but it's either that or inspecting `self._outcome.results` to determine if the test is running with the debug sql option.

matthiask:
> @matthiask sorry, can you review that last commit too. It's a bit of a hack, but it's either that or inspecting `self._outcome.results` to determine if the test is running with the debug sql option.

Thanks, I think the way it's implemented makes sense!

living180:
Didn't have time to review this until just now, but the implementation looks good and a basic smoke test passes for me.  The only improvement I'd suggest would be a bit of bike-shedding by renaming `NormalCursorWrapper` and `ExceptionCursorWrapper` and `NormalCursorMixin` and `ExceptionCursorMixin`, respectively, but that's just a matter of taste.

tim-schilling:
Good enough for me.

Issue #1808, opened 2023-07-04
Extend and document Panel.get_stats to utilize store API and panel deserialization if it exists. Requires fallback.


tim-schilling:

tim-schilling:
Closed in #1826

PR #1826, opened 2023-08-21
Serialize the panels

### LINKED ISSUE #1807: Extend and document Panel.record_stats to utilize store API and panel serialization if it exists. Requires fallback.

tim-schilling:
This is a draft PR. I'm keeping my on-going work in this branch. I found that my efforts to close #1807 and #1808 quickly spiraled into a major refactor of core logic.

This currently handles the SQLPanel which required the following changes:

- Drop `raw_params` entirely. We'll need to serialize the parameters every time now so we can't rely on the original params
- Refactor the SQL queries to use a `djdt_query_id` which uniquely identifies each query. The SQL forms now reference this id and avoid passing SQL to be executed.
- Move the formatting logic of SQL queries back to just before rendering. Otherwise we're rendering and storing the html in the store (I could be convinced I'm over-engineering for the future)

In addition this also changed our design for `Panel`

- `load_stats_from_store` - Generates a `Panel` instance with the passed in data.
- `Panel.panel_id` is now a class member to avoid needing to instantiate a panel to programatically get the panel id. With the new focus on fetching data for a specific panel given its ID, this became more of a need


tim-schilling:
I think this is at a point where we can add some async tests to confirm that this approach will work or not.

tests/panels/test_sql.py
```diff
@@ -383,8 +384,6 @@ def test_insert_content(self):
         """
         list(User.objects.filter(username="café"))
         response = self.panel.process_request(self.request)
-        # ensure the panel does not have content yet.
-        self.assertNotIn("café", self.panel.content)
```
matthiask, line 387: This test didn't just check for the presence of the string but also our strings vs bytes handling, see https://github.com/jazzband/django-debug-toolbar/commit/476b5ce2a9d8261886b6f4a60fcbe7c62693d7de

I'm not sure if this is still being tested?

matthiask (commented):
Thanks!

- I like the change from store IDs to request IDs.
- It's nice that we no longer accept SQL strings from the client and execute them (even with signing)

I'm not totally sure, is this (somewhat) backwards compatible for third party panels? This wouldn't be an argument against it, just something to clarify.

tim-schilling:
@matthiask I'm back to forcing everything that can't be serialized to JSON to a str. Otherwise we're trying to do it manually per panel which will involve a function that's likely going to do similar object and collection iteration as `json.dumps()`.

tim-schilling:
This assumes we'll have some shared memory store for the cache hence the change to `should_render_panels`

salty-ivy:
from the above changes in changes in consideration I was able able to run example with async calls
image

image

tim-schilling:
> however I believe this async example would also work without serialised panels.

Why is that?

tim-schilling:
I missed this PR up. I meant to do the merge into serializable then rebase this. If the tests pass here, I'll push it to serializable.

PR #1832, opened 2023-09-16
do not quote SQL params before passing them to mogrify

### LINKED ISSUE #1831: list values incorrectly adapted in SQL panel
I have a sql query containing `... WHERE id = ANY(%s)` and the param being a list, e.g. `[1, 2]`.

Adapted, the query looks like `id = ANY('{1,2}'::int2[]) `, but in the SQL panel, it shows up as `id = ANY('[1,2]')`. This is due to the parameters being quoted already before being passed to mogrify. Dropping the extra quoting step, the query shows up correctly in the SQL panel.

```
diff --git a/debug_toolbar/panels/sql/tracking.py b/debug_toolbar/panels/sql/tracking.py
index 0c53dc2c..bcff5905 100644
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -115,15 +115,6 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
         else:
             return repr(element)
 
-    def _quote_params(self, params):
-        if not params:
-            return params
-        if isinstance(params, dict):
-            return {key: self._quote_expr(value) for key, value in params.items()}
-        if isinstance(params, tuple):
-            return tuple(self._quote_expr(p) for p in params)
-        return [self._quote_expr(p) for p in params]
-
     def _decode(self, param):
         if PostgresJson and isinstance(param, PostgresJson):
             # psycopg3
@@ -157,9 +148,7 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
         # process during the .last_executed_query() call.
         self.db._djdt_logger = None
         try:
-            return self.db.ops.last_executed_query(
-                self.cursor, sql, self._quote_params(params)
-            )
+            return self.db.ops.last_executed_query(self.cursor, sql, params)
         finally:
             self.db._djdt_logger = self.logger
 
```
Note that I've tested only with psycopg2/3, I don't know if other backends require the extra quoting.

[tracking.txt](https://github.com/jazzband/django-debug-toolbar/files/12611437/tracking.txt)

tkoschnick:
# Description

SQL params should be adapted to the correct database representation by the corresponding driver's mogrify() for display in the SQL panel. Using the default Python representation works only for simple data types.

Fixes #1831

# Checklist:

- [x] I have added no tests for this change - no new functionality, current tests pass.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,6 +5,8 @@ Pending
 -------
 
 * Removed outdated third-party panels from the list.
+* Do not quote SQL params before passing them to mogrify() for display in SQL
+  panel.
 
 4.2.0 (2023-08-10)
 ------------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,8 +5,7 @@ Pending
 -------
 
 * Removed outdated third-party panels from the list.
-* Do not quote SQL params before passing them to mogrify() for display in SQL
-  panel.
+* Avoided the unnecessary work of recursively quoting SQL parameters.
 
 4.2.0 (2023-08-10)
 ------------------
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -109,21 +109,6 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
     Wraps a cursor and logs queries.
     """
 
-    def _quote_expr(self, element):
-        if isinstance(element, str):
-            return "'%s'" % element.replace("'", "''")
-        else:
-            return repr(element)
-
-    def _quote_params(self, params):
-        if not params:
-            return params
-        if isinstance(params, dict):
-            return {key: self._quote_expr(value) for key, value in params.items()}
-        if isinstance(params, tuple):
-            return tuple(self._quote_expr(p) for p in params)
-        return [self._quote_expr(p) for p in params]
-
     def _decode(self, param):
         if PostgresJson and isinstance(param, PostgresJson):
             # psycopg3
@@ -157,9 +142,7 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
         # process during the .last_executed_query() call.
         self.db._djdt_logger = None
         try:
-            return self.db.ops.last_executed_query(
-                self.cursor, sql, self._quote_params(params)
-            )
+            return self.db.ops.last_executed_query(self.cursor, sql, params)
         finally:
             self.db._djdt_logger = self.logger
 


docs/changes.rst
```diff
@@ -5,6 +5,8 @@ Pending
 -------
 
 * Removed outdated third-party panels from the list.
+* Do not quote SQL params before passing them to mogrify() for display in SQL
+  panel.
```
matthiask, line 9: The spell check failed; maybe something like the following would be a better description of the intent behind the change?

```suggestion
* Avoided the unnecessary work of recursively quoting SQL parameters.
```

matthiask (commented):
I have to say I'm surprised that the tests all passed. it seems that the quoting really isn't necessary after all...?

tkoschnick:
> I have to say I'm surprised that the tests all passed. it seems that the quoting really isn't necessary after all...?

I would even argue that it's not only not necessary, but counterproductive. The job of mogrify() is to adapt/quote the query and params exactly as they would be by execute() before being passed to the databases' parser. If all parameters are quoted beforehand, they are presented as strings to mogrify(), which can't possibly adapt them correctly any more.

tkoschnick:
Regarding the tests passing - this is essentially only a presentational change, concerning how the query is displayed to the user in the SQL panel. So far, it has been represented differently from what is actually sent to the database; now it matches, since execute() and mogrify() receive the same parameters.

tim-schilling:
I don't think @matthiask is surprised that this works for postgres, but rather for mysql and sqlite. The process of getting a executable sql query is different between postgres, mysql and sqlite. I'm very skeptical of this change simply because our test suite doesn't cover every possibility of SQL access that the panel instruments in the wild.

tim-schilling (approved):
I think I'm onboard with this. As far as I can tell the logic checks out and the original reason for the quoting is covered.

living180:
This change doesn't affect Django's MySQL backend, as it doesn't use the `params` argument for its `last_executed_query()` implementation.  The sqlite backend on the otherhand does, but I haven't tried that backend with this change.

tim-schilling:
I ran this locally with sqlite and tested it with an int, string and datetime. It went fine which is why I approved it.

living180:
It looks like the code in question was originally introduced in #224 (merged January 2012), and appears to have been for MySQL, but Django 1.4 (May 2012) added a `last_executed_query()` implementation for the MySQL backend (https://github.com/django/django/commit/5b0e4e49d4ab5e976fbfdde70c525a13220f3259) which no longer used the params argument.

PR #1833, opened 2023-09-19
panels(templates): avoid evaluating LazyObject

nijel:
# Description

LazyObject is typically used for something expensive to evaluate, so avoid evaluating it just for showing it in the debug toolbar.

Fixes # (issue)

# Checklist:

I will look at writing tests once I get an approval, that this is a viable approach.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -8,7 +8,8 @@ Pending
 * Avoided the unnecessary work of recursively quoting SQL parameters.
 * Postponed context process in templates panel to include lazy evaluated
   content.
-* Fixed template panel to avoid evaluating ``LazyObject`` when not already evaluated.
+* Fixed template panel to avoid evaluating ``LazyObject`` when not already 
+  evaluated.
 
 4.2.0 (2023-08-10)
 ------------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -8,7 +8,7 @@ Pending
 * Avoided the unnecessary work of recursively quoting SQL parameters.
 * Postponed context process in templates panel to include lazy evaluated
   content.
-* Fixed template panel to avoid evaluating ``LazyObject`` when not already 
+* Fixed template panel to avoid evaluating ``LazyObject`` when not already
   evaluated.
 
 4.2.0 (2023-08-10)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -7,6 +7,7 @@ Pending
 * Removed outdated third-party panels from the list.
 * Postponed context process in templates panel to include lazy evaluated
   content.
+* Fixed template panel to avoid evaluating ``LazyObject`` when not already evaluated.
 
 4.2.0 (2023-08-10)
 ------------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -5,6 +5,8 @@ Pending
 -------
 
 * Removed outdated third-party panels from the list.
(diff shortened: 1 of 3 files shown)


matthiask:
Hey, 

Thanks for your contribution! 

I have questions:

- SQL queries should be impossible anyway; which other expensive context variables could be wrapped by `LazyObject` in your experience? I myself am using `lazy()` for all sorts of things, but mostly to wrap objects which generate SQL queries or translations, so until now evaluating such an object once too much didn't bother me.
- I had to read the code and I think that we still show the object if it has been evaluated already, is that correct?

nijel:
For me, the reason was an expensive SQL query. I turned something possibly expensive (and rarely used) to a `LazyObject` (in https://github.com/WeblateOrg/weblate/commit/d49c46e2f77777f274430d769406454f1a48b702) and the query was still executed when running with Django debug toolbar. That took me some time to figure out that it's not my broken code, but the Django debug toolbar evaluates all template variables to display them.

SimpleLazyObject has `__repr__` that shows the value if it has already been evaluated (see https://github.com/django/django/blob/8af3ae4ad9ca475f2428fac950de8df56b575e6a/django/utils/functional.py#L406-L413), that's why I used `repr()`. This should have been mentioned in the comment, I'll include that.

PS: `lazy()` seems to do a different thing than `LazyObject`, so this change covers `LazyObject` only.

matthiask:
Thanks for the explanation! Makes sense. 

> I will look at writing tests once I get an approval, that this is a viable approach.

I approve :)

docs/changes.rst
```diff
@@ -5,6 +5,7 @@ Pending
 -------
 
 * Removed outdated third-party panels from the list.
+* Avoid evaluating ``LazyObject`` in the templates panel.
```
matthiask, line 8: We mostly use the past tense for the changelog.

tests/panels/test_template.py
```diff
@@ -109,6 +110,14 @@ def test_empty_context(self):
...
         )
 
+    def test_lazyobject(self):
+        response = self.panel.process_request(self.request)
+        t = Template("")
+        c = Context({"lazy": SimpleLazyObject(lambda: "lazy_value")})
+        t.render(c)
+        self.panel.generate_stats(self.request, response)
+        self.assertNotIn("lazy_value", self.panel.content)
+
```
matthiask, line 120: Please add a "positive" test to demonstrate that the value is shown when it has been evaluated before.

matthiask (approved):
Looks good! Two minor comments.

tests/panels/test_template.py
nijel: Good point, it actually doesn't work as intended because the `template_rendered` signal is sent *before* the template is actually rendered. It looked like it was working for me because I forced evaluation before rendering the template… So the patch needs to be a bit more complex.

docs/changes.rst
```diff
@@ -6,6 +6,9 @@ Pending
 
 * Removed outdated third-party panels from the list.
 * Avoided the unnecessary work of recursively quoting SQL parameters.
+* Postponed context process in templates panel to include lazy evaluated
+  content.
+* Fixed template panel to avoid evaluating ``LazyObject`` when not already evaluated.
```
matthiask, line 11:
```suggestion
* Fixed template panel to avoid evaluating ``LazyObject`` when not already 
  evaluated.
```

matthiask:
Thanks!

Issue #1828, opened 2023-08-30
Deadlock when using concurrency in async views


salomvary:
Hi folks,

First of all thanks for maintaining this amazing project! I've recently started migrating an existing project to asynchronous views and noticed that the view gets deadlocked when tasks are executed concurrently **and** the Debug Toolbar is installed.

If no tasks run concurrently (ie. using a sequence of `await`s) or the Debug Toolbar is removed, everything seems fine.

Below is an example that hangs forever on `await asyncio.gather`:

```python
async def get(request):
    (one_thing, another_thing) = await asyncio.gather(
        Thing.objects.filter(pk=1).afirst(),
        Thing.objects.filter(pk=2).afirst()
    )

    return HttpResponse(content=f"Hello {one_thing.pk} {another_thing.pk}")
```

I'm relatively new to Python `asyncio` and have therefore no idea what's going on. I see the ongoing async issues in this issue tracker but I am not sure what's the state of Django async support (or lack of it) is here. (Something that might be worth documenting on the Readme.)

These are the dependencies in use:
```
$ poetry show -T   
daphne               4.0.0   Django ASGI (HTTP/WebSocket) server
django               4.2.4   A high-level Python web framework that encourages rapid development and clean, pragmatic design.
django-debug-toolbar 4.2.0   A configurable set of panels that display various debug information about the current request/response.
psycopg2-binary      2.9.7   psycopg2 - Python-PostgreSQL Database Adapter
```

I can share a minimal test project if someone wants to have a look.

tim-schilling:
The toolbar is sync only at this time. If you'd like to open a PR to call that out explicitly in the Readme and installation docs, that would be helpful. There was another idea to log a message or not fail too. If you can help, it'd be appreciated.

salomvary:
> The toolbar is sync only at this time. If you'd like to open a PR to call that out explicitly in the Readme and installation docs, that would be helpful. There was another idea to log a message or not fail too. If you can help, it'd be appreciated.

@tim-schilling Something like this? https://github.com/jazzband/django-debug-toolbar/pull/1829

How big of an undertaking would be to add async support? (Full or partial). Looking at these issues, I have no idea: https://github.com/jazzband/django-debug-toolbar/issues?q=is%3Aissue+project%3Ajazzband%2F9+is%3Aopen

[...]

Issue #1819, opened 2023-08-01
Add test for async views and ORM usage.


matthiask:

matthiask:
I have started digging into tracking SQL statements for a bit and I'm unsure how to proceed.

ORM queries are started inside a thread pool executor when using an ASGI server. I have no idea how we could go from the DB cursor to the view which originally caused or started the request.


Operating systems don't generally store data about where threads come from, async tasks do not generally have an accessible parent or task group. Instrumentation is currently enabled per thread, but that also doesn't work when the activation happens in a different thread (or does it? maybe the `sync_to_async(..., thread_sensitive=True)` thing helps us and we can just assume we're in the correct thread? But then we still don't know to which view we belong?)

(Things are complicated even more because views can kick off tasks which will produce ORM queries after the response has been returned to the client...)

tim-schilling:
@matthiask @salty-ivy can this be closed due to the GSoC-2024 project?

salty-ivy:
I believe yes, as far as I understand this issues description.

We have added tests for both async and orm usage.

PR #1835, opened 2023-09-19
Add async tests

salomvary:
As discussed with @tim-schilling in #1828 a good start towards adding async support is adding tests involving async views and async ORM usage.

Unfortunately, none of the automated tests showcase the problem described in #1828 as there is no ASGI equivalent of `LiveServerTestCase` and the issue seems to be specific to running an actual ASGI server. 

Therefore, despite all tests are green, async support is still not there. Running the example app with an ASGI server using `ASYNC_SERVER=true python example/manage.py runserver` and visiting http://127.0.0.1:8000/async/db-concurrent/ does exhibit the deadlock though.

See also [#1819](https://github.com/jazzband/django-debug-toolbar/issues/1819).

### DIFF
--- a/example/README.rst
+++ b/example/README.rst
@@ -46,3 +46,13 @@ environment variable::
 
     $ DB_BACKEND=postgresql python example/manage.py migrate
     $ DB_BACKEND=postgresql python example/manage.py runserver
+
+Using an asynchronous (ASGI) server:
+
+Install [Daphne](https://pypi.org/project/daphne/) first:
+
+    $ python -m pip install daphne
+
+Then run the Django development server:
+
+    $ ASYNC_SERVER=true python example/manage.py runserver
--- a/example/settings.py
+++ b/example/settings.py
@@ -18,6 +18,7 @@ INTERNAL_IPS = ["127.0.0.1", "::1"]
 # Application definition
 
 INSTALLED_APPS = [
+    *(["daphne"] if os.getenv("ASYNC_SERVER", False) else []),  # noqa: FBT003
     "django.contrib.admin",
     "django.contrib.auth",
     "django.contrib.contenttypes",
@@ -66,6 +67,7 @@ TEMPLATES = [
 USE_TZ = True
 
 WSGI_APPLICATION = "example.wsgi.application"
+ASGI_APPLICATION = "example.asgi.application"
 
 
 # Cache and database
@@ -103,7 +105,6 @@ if os.environ.get("DB_BACKEND", "").lower() == "mysql":
 
 STATICFILES_DIRS = [os.path.join(BASE_DIR, "example", "static")]
 
-
 # Only enable the toolbar when we're in debug mode and we're
 # not running tests. Django will change DEBUG to be False for
 # tests, so we can't rely on DEBUG alone.
@@ -117,3 +118,26 @@ if ENABLE_DEBUG_TOOLBAR:
     ]
     # Customize the config to support turbo and htmx boosting.
     DEBUG_TOOLBAR_CONFIG = {"ROOT_TAG_EXTRA_ATTRS": "data-turbo-permanent hx-preserve"}
+
+LOGGING = {
+    "version": 1,
+    "disable_existing_loggers": False,
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+        },
+    },
+    "root": {
+        "handlers": ["console"],
+        "level": "WARNING",
+    },
+    "loggers": {
+        # Log when an asynchronous handler is adapted for middleware.
+        # See warning here: https://docs.djangoproject.com/en/4.2/topics/async/#async-views
+        "django.request": {
+            "handlers": ["console"],
+            "level": os.getenv("DJANGO_REQUEST_LOG_LEVEL", "INFO"),
+            "propagate": False,
+        },
+    },
+}
--- a/example/settings.py
+++ b/example/settings.py
@@ -118,26 +118,3 @@ if ENABLE_DEBUG_TOOLBAR:
     ]
     # Customize the config to support turbo and htmx boosting.
     DEBUG_TOOLBAR_CONFIG = {"ROOT_TAG_EXTRA_ATTRS": "data-turbo-permanent hx-preserve"}
-
-LOGGING = {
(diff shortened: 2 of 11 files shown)


example/README.rst
```diff
@@ -46,3 +46,13 @@ environment variable::
 
     $ DB_BACKEND=postgresql python example/manage.py migrate
     $ DB_BACKEND=postgresql python example/manage.py runserver
+
+Using an asynchronous (ASGI) server:
+
+Install [Daphne](https://pypi.org/project/daphne/) first:
+
+    $ python -m pip install daphne
```
tim-schilling, line 54: Let's just add it to `requirements-dev.txt` instead.

example/settings.py
```diff
@@ -16,6 +16,7 @@
 # Application definition
 
 INSTALLED_APPS = [
+    *(["daphne"] if os.getenv("ASYNC_SERVER", False) else []),
```
tim-schilling, line 19: What do you think about defining an `async_settings.py` file that would contain the async settings such as this one and `ASGI_APPLICATION`? It would change the run command to `python example/manage.py runserver --settings=example.async_settings`

example/settings.py
```diff
@@ -97,3 +99,26 @@
...
+            "class": "logging.StreamHandler",
+        },
+    },
+    "root": {
+        "handlers": ["console"],
+        "level": "WARNING",
+    },
+    "loggers": {
+        # Log when an asynchronous handler is adapted for middleware.
+        # See warning here: https://docs.djangoproject.com/en/4.2/topics/async/#async-views
```
tim-schilling, line 117: Sorry, I don't fully understand the purpose here. Can you help me understand this?

tim-schilling (commented):
Sorry for the late reply on this. Thank you for creating this PR!

I wonder if we could test this functionality by starting up `runserver` with the asgi application, then using a separate selenium test against that rather than using the Django test runner.

example/settings.py
matthiask: Warning the user when having to switch from async to sync in the middleware/view stack makes a lot of sense.

I'm not 100% sure if that has to be a part of django-debug-toolbar's example settings though. Is it necessary to define `LOGGING` at all? Isn't it the default behavior already to emit those warnings?

example/settings.py
matthiask: FWIW I prefer the single settings module pattern (I'd rather have a few `if` statements than star imports) but I don't feel strongly about it in this project. If different environments have to have different settings at all I find it clearer to see at a glance what the differences are. Maybe that's just me though.

salomvary: I don't have a strong opinion here (use single settings on real-life projects myself but that does not have to apply everywhere else).

tim-schilling: > FWIW I prefer the single settings module pattern (I'd rather have a few if statements than star imports) but I don't feel strongly about it in this project. If different environments have to have different settings at all I find it clearer to see at a glance what the differences are.

Wouldn't it be easier to see the differences are between environments with the multiple settings modules?

example/settings.py
salomvary: This allows surfacing the fact that django-debug-toolbar is not an async middleware, therefore Django needs to some extra stuff when using async views. 

Probably only relevant until django-debug-toolbar becomes an async middleware.

Shall we remove from the example then?

example/settings.py
matthiask: @tim-schilling As I wrote above, it's my preference and I don't want to impose it on anyone. I find it easier because I do not have to check several files for the settings, only one file. I don't have to do the "merging" of settings in my head, it's right there in the code in an if statement. But again, just my preference. If you feel strongly about that (for django-debug-toolbar) feel free to request splitting up the settings modules. I'm `-0` on the split, not `-1`.

tim-schilling: Gotcha. I was confused by your earlier statement. We can keep it as is for now.

salomvary: ..and for the record, `example/settings.py` already has a few `if`s :)

example/settings.py
tim-schilling: I think my vote is remove it or keep it but default it to DEBUG so that we get that logging you mention.

example/README.rst
tim-schilling: Since this doc also includes the instruction to install django, I think you were right initially. We should keep this instruction. I did add it to the requirements_dev.txt file though since contributors are likely to run this and it makes sense to simply that flow.

tim-schilling:
Closing in favor of #1938. Thank you for pushing this forward @salomvary! Your work here was helpful to the GSoC project to push the toolbar forward with async compatibility.

tim-schilling:
Nope, I was wrong. This still has value.

example/settings.py
tim-schilling: I am going to remove this from the PR since #1938 makes the middleware async.

tim-schilling:
Rebased on main

tim-schilling:
@salomvary could you confirm that the deadlock issue is not a problem in this version? @salty-ivy removed the sql panel from async installations which seems to have resolved that issue. Granted, now the issue is that the toolbar doesn't have a SQL panel for async apps, but at least it doesn't lock up.

salty-ivy:
> @salomvary could you confirm that the deadlock issue is not a problem in this version? @salty-ivy removed the sql panel from async installations which seems to have resolved that issue. Granted, now the issue is that the toolbar doesn't have a SQL panel for async apps, but at least it doesn't lock up.

I think it was more a django ORM issue rather than the toolbar itself. @salomvary  mentioned it here 
https://forum.djangoproject.com/t/are-concurrent-database-queries-in-asgi-a-thing/24136/2

tim-schilling:
Weird, on my machine it was executing them concurrently (I think). Or at least they didn't result in a deadlock.

salomvary:
> Thank you @salomvary for this! I'd still like your confirmation the deadlock has been handled when/if you have time.

Yep, the deadlock was very likely not cause by the Debug Toolbar, it is a Django issue. As far as I can remember I was later able to reproduce the deadlock with Debug Toolbar not even installed.

Issue #1836, opened 2023-09-22
Profiler not correctly filtering files on Windows


Oisota:
The profiler is not filtering out non-project code when running on Windows. The culprit seems to be [these lines](https://github.com/jazzband/django-debug-toolbar/blob/199c2b3c6bc916d371bcfb11b0fcbbb690ee5e36/debug_toolbar/panels/profiling.py#L47-L48) in the `is_project_func` method. The forward slashes in the paths will never be in the path when on windows and all non project code gets included in the profile.

I was able to fix the issue locally by simply removing the forward slashes from the two paths. If this is an acceptable fix, I'd be happy to submit a PR.

tim-schilling:
Hmm, there's probably a better way to inspect the filepath. I think we'd want to check that these are directories in the filepath and not simply check for the directory name in the filepath. A PR is definitely welcome!

DraKen0009:
This is my first pull request , any advises would be greatly appreciated.

PR #1857, opened 2023-11-17
Refactor is_project_func method for Windows compatibility

DraKen0009:
The previous implementation of the is_project_func method had issues related to path handling on Windows systems, potentially leading to incorrect identification of project code. 

Changes made:
- Utilized os.path.normpath() for path normalization.
- Modified file path comparison to correctly identify project code.
- Implemented exclusion of 'site-packages' and 'dist-packages' directories.

Fixes: #1836

### DIFF
--- a/debug_toolbar/panels/profiling.py
+++ b/debug_toolbar/panels/profiling.py
@@ -42,10 +42,17 @@ class FunctionCall:
         """
         if hasattr(settings, "BASE_DIR"):
             file_name, _, _ = self.func
+            base_dir = str(settings.BASE_DIR)
+
+            file_name = os.path.normpath(file_name)
+            base_dir = os.path.normpath(base_dir)
+
             return (
-                str(settings.BASE_DIR) in file_name
-                and "/site-packages/" not in file_name
-                and "/dist-packages/" not in file_name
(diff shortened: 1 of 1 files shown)


tim-schilling:
@DraKen0009 do you have a test that confirms that this fixes the problem on windows? I doubt we'll run it in CI, but it'll make me feel better if there were a test along with this.

DraKen0009:
@tim-schilling  ,  I haven't created tests yet , but I can create a simple unit test for it, can you guide me where in project directory do I need to create it ?

tim-schilling:
It probably belongs in the test file for the profile panel. If you look at the tests directory, there should be a panels directory with files for each of the panels.

matthiask (commented):
I'm a bit saddened by the move from `Path` objects to `os.path`. Isn't there a way to achieve the same thing using `pathlib`?

tim-schilling:
@DraKen0009 is there any option to continue using `pathlib`?

matthiask:
Tests pass and the implementation is strictly superior to the current one. I have removed the pathlib flake8 rules from my own projects since it's annoying and using pathlib isn't always better.

Thanks!

Issue #1842, opened 2023-10-18
Improve accessibility of the toolbar


tim-schilling:
We should start identifying what UI changes we need to support WCAG 2.2.

A tool that can help identify issues is: https://accessibilityinsights.io/

Thibaud has a good talk on accessibility from a few years ago: https://www.youtube.com/watch?v=PO0w2BJk33s

sabderemane:
Hi there, I'll make an accessibility audit and see with the Django Accessibility Team to break it into tickets and provide help on that :)

thibaudcolas:
FYI @sabderemane @knyghty discussed this – we’re keen to try to do this with a Djangonaut Space participant. @sabderemane will do the audit and let them do the fixes if no one wants to audit. Or they can do it all.

tim-schilling:
Sounds good, I'll bring it up with the group next week and let you know. We'll probably take some of the logistical discussion offline, but put the long term stuff here.

sabderemane:
Thank you @thibaudcolas ❤️

VeldaKiara:
Based on the incoming design, I ran automated accessibility tests with axe-core 4.10.2 (WCAG 2.0/2.1 A + AA rulesets) against the rendered toolbar in both themes:</p>
  | Light | Dark
-- | -- | --
Violations | 0 | 0
Passes | 11 | 11
Incomplete | 0 ✅ | 0 ✅
Inapplicable | 49 | 49

A second scan against populated panels also reports 0 violations in both themes.

PR #1865, opened 2023-12-23
Improving accessibility with ARIA attributes in templates

ACK1D:
# Description

This PR aims to enhance accessibility across templates tool. The changes involve the addition of appropriate attributes, such as role attributes and those related to content narration, to improve the user experience for individuals with limited capabilities.

The motivation behind these changes is to create a more inclusive and user-friendly environment within the tool.

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/base.html
+++ b/debug_toolbar/templates/debug_toolbar/base.html
@@ -1,38 +1,51 @@
 {% load i18n static %}
+
 {% block css %}
-<link rel="stylesheet" href="{% static 'debug_toolbar/css/print.css' %}" media="print">
-<link rel="stylesheet" href="{% static 'debug_toolbar/css/toolbar.css' %}">
+    <link rel="stylesheet" href="{% static 'debug_toolbar/css/print.css' %}" media="print">
+    <link rel="stylesheet" href="{% static 'debug_toolbar/css/toolbar.css' %}">
 {% endblock %}
+
 {% block js %}
-<script type="module" src="{% static 'debug_toolbar/js/toolbar.js' %}" async></script>
+    <script type="module" src="{% static 'debug_toolbar/js/toolbar.js' %}" async></script>
 {% endblock %}
+
 <div id="djDebug" class="djdt-hidden" dir="ltr"
      {% if not toolbar.should_render_panels %}
-     data-store-id="{{ toolbar.store_id }}"
-     data-render-panel-url="{% url 'djdt:render_panel' %}"
+         data-store-id="{{ toolbar.store_id }}"
+         data-render-panel-url="{% url 'djdt:render_panel' %}"
      {% endif %}
      {% url 'djdt:history_sidebar' as history_url %}
      {% if history_url %}
-     data-sidebar-url="{{ history_url }}"
+         data-sidebar-url="{{ history_url }}"
      {% endif %}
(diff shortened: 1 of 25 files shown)


matthiask (commented):
I don't really know enough about a11y to judge this pull request, but

- `role="status"` for a stacktrace seems strange?
- `role="row"` for a tr element seems very redundant, as does `role="columnheader"` for a cell inside `thead` (this also concerns other changes)
- tagging both `dt` and `dt` using `listitem` doesn't seem correct to me, one is a title and one isn't

If someone can confirm that this change fixes real compatibility issues I'm for it; if not I'm definitely against introducing all those `role` attributes. I fear it's more noise than signal at this point.

ACK1D:
@matthiask I'm not sure if attributes would be useful for use in Django. But when debugging site availability in [Wagtail](https://github.com/wagtail/wagtail), the lack of attributes gets in the way. 
ss1

tim-schilling:
@ACK1D it looks like the changes are also causing problems with the test suite. Can you look into getting that to pass? I suspect getting it to pass one environment will resolve it for all of them.

ACK1D:
@tim-schilling done.

matthiask:
@sabderemane Feel free to ignore this, but I saw in #1842 that you're (probably/surely) much more knowledgeable in a11y than I am and I would appreciate a review or a few opinions on this.

sabderemane:
> @sabderemane Feel free to ignore this, but I saw in #1842 that you're (probably/surely) much more knowledgeable in a11y than I am and I would appreciate a review or a few opinions on this. 

I will have a look !

thibaudcolas:
I wouldn’t recommend proceeding with those changes as-is. There’s lots of attributes added that wouldn’t do anything, lots where using the more correct HTML elements to start with would be a better fix, and lots where ARIA is used incorrectly.

I think it’d be simpler to assess those fixes if the issues were defined in isolation from one-another, and solutions devised separately, rather than a blanket addition of ARIA attributes.

For example,

- Adding `role="link"` on a `a` element with a href does nothing. It’s already a `link`.
- Panel button list items can’t be `role="checkbox"` and have interactive elements inside, nor can a `ul` have descendants that are ` role="checkbox"`
- `<h4 role="heading" aria-level="2">` should just be an `h2`. 

Please have a look at [the first rule of ARIA](https://www.w3.org/TR/using-aria/). In particular, use of the `role` attribute should be a last resort and I can’t imagine it would ever be the right approach except for [ARIA roles with no corresponding HTML element](https://www.w3.org/TR/using-aria/#html5na) (and if so, follow [ARIA authoring practices](https://www.w3.org/WAI/ARIA/apg/)). See [ARIA in HTML](https://www.w3.org/TR/html-aria/) for a list of HTML elements’ built-in roles.

@ACK1D if you want to spend more time on this I would recommend to start by sharing the results of accessibility testing (#1842), possibly from automated tests. Then report those findings, and fix the issues one by one, making sure to test appropriately. If you want to add ARIA attributes in particular, I would recommend testing with at least one screen reader. See https://github.com/django/django/pull/17338 for recommended accessibility testing steps for Django contributors.

matthiask:
@thibaudcolas Thank you, that's very helpful!

tim-schilling:
I'm closing this in favor of Thibaud's direction here: https://github.com/jazzband/django-debug-toolbar/pull/1865#issuecomment-1891286544

Issue #1853, opened 2023-11-01
The djdt handle shouldn't be stuck at the top of the browser window initially


matthiask:
The handle can always be dragged around. But I'd prefer it if the handle wouldn't be at the top, but rather somewhere in the upper half of the browser window by default. (Hardcoding a better value is fine I think.)

The area at the top right is often used for other interactive elements, if I wouldn't have to move the toolbar handle in all user agents again that would be great.

VeldaKiara:
Hi @tim-schilling and @matthiask  I would be happy to work on this issue.

matthiask:
@VeldaKiara Yes, please!

VeldaKiara:
On it.

VeldaKiara:
Is this okay, or should I lower it?
<img width="1680" alt="Screenshot 2024-01-23 at 13 40 49" src="https://github.com/jazzband/django-debug-toolbar/assets/32552296/ede6facc-84ab-4542-ac6c-cf37c7651446">

matthiask:
I think somewhere in the upper half is fine (as you've done it)

VeldaKiara:
Done

matthiask:
Closed by #1871

PR #1871, opened 2024-01-23
The djdt handle shouldn't be stuck at the top of the browser window initially #1853

VeldaKiara:
# Description
The issue was as much as the toolbar can be moved, it would be preferred if its position was on the upper top instead of the top.

Fixes # (issue)
The issue is mentioned here #1853 
# Checklist:

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -220,7 +220,7 @@
     background-color: #fff;
     border: 1px solid #111;
     border-bottom: 0;
-    top: 0;
+    top: 268px;
     right: 0;
     z-index: 100000000;
     opacity: 0.75;
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -220,7 +220,7 @@
     background-color: #fff;
     border: 1px solid #111;
     border-bottom: 0;
-    top: 268px;
+    top: 0;
     right: 0;
     z-index: 100000000;
     opacity: 0.75;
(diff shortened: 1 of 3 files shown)


debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -220,7 +220,7 @@
     background-color: #fff;
     border: 1px solid #111;
     border-bottom: 0;
-    top: 0;
+    top: 268px;
```
tim-schilling, line 223: What do you think about using a relative position rather than an fixed value?

VeldaKiara: Yes, that would work too. Let me try it out and see the changes

matthiask (commented):
I tried it out and the handle was still stuck at the top. Maybe I did something wrong, but it seems to me that a better way would be to change the fallback in `ensureHandleVisibility` to not be `0` but something like `200`:

https://github.com/jazzband/django-debug-toolbar/blob/77aa47a761da203c52e2328990123abb252d8dd5/debug_toolbar/static/debug_toolbar/js/toolbar.js#L225-L233

I like Tim's idea of using a relative value for the default but maybe it will be harder to implement the handle dragging since you'd have to convert pixels to relative values there as well.

debug_toolbar/static/debug_toolbar/css/toolbar.css
VeldaKiara: or what if instead of changing the position, I adjust the y translation?
What do you think?

VeldaKiara:
> I tried it out and the handle was still stuck at the top. Maybe I did something wrong, but it seems to me that a better way would be to change the fallback in `ensureHandleVisibility` to not be `0` but something like `200`:
> 
> https://github.com/jazzband/django-debug-toolbar/blob/77aa47a761da203c52e2328990123abb252d8dd5/debug_toolbar/static/debug_toolbar/js/toolbar.js#L225-L233
> 
> I like Tim's idea of using a relative value for the default but maybe it will be harder to implement the handle dragging since you'd have to convert pixels to relative values there as well.

I noticed that if I use position relative, I will have to make numerous changes however I decided to leave the button as is because it was always getting overridden by the fallback function. You and @tim-schilling can test it out now to see if it works now.

matthiask:
Thank you!

VeldaKiara:
Happy to help.

PR #1872, opened 2024-01-24
1843 new ajax request resets whole view if historypanel is enabled

### LINKED ISSUE #1843: New AJAX request resets whole view if HistoryPanel is enabled.
Maybe I am doing something wrong, but I find working with DDT with HistoryPanel enabled quite annoying.

I have notifications on my website which makes request every ~5 seconds if there is anything new.

If I have HistoryPanel enabled in DDT this means that if I am exploring some request (from history or just the last one) and this AJAX notification request is made I loose everything I am seeing and the whole DDT resets to the newest (notification) request.

Would it be possible to set DDT so that it switches the request only if I explicitly select it from history?

elineda:
# Description

[+] Create a setting which enable or disable the automatic refresh when an ajax request occurs.

Fixes #1843

# Checklist:

### DIFF
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -42,6 +42,7 @@ CONFIG_DEFAULTS = {
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
     "TOOLBAR_LANGUAGE": None,
+    "UDPATE_ON_AJAX": False,
 }
 
 
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -42,7 +42,7 @@ CONFIG_DEFAULTS = {
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
     "TOOLBAR_LANGUAGE": None,
-    "UPDATE_ON_FETCH": True,
+    "UPDATE_ON_FETCH": False,
 }
 
 
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -42,7 +42,7 @@ CONFIG_DEFAULTS = {
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
     "TOOLBAR_LANGUAGE": None,
-    "UDPATE_ON_AJAX": False,
+    "UPDATE_ON_FETCH": True,
 }
 
 
--- a/debug_toolbar/static/debug_toolbar/js/history.js
+++ b/debug_toolbar/static/debug_toolbar/js/history.js
@@ -104,5 +104,6 @@ $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     refreshHistory();
 });
-// As we don't refresh the whole toolbar each fetch or ajax request we need to refresh the history when we open the panel
+// We don't refresh the whole toolbar each fetch or ajax request,
+// so we need to refresh the history when we open the panel
 $$.onPanelRender(djDebug, "HistoryPanel", refreshHistory);
--- a/debug_toolbar/static/debug_toolbar/js/history.js
+++ b/debug_toolbar/static/debug_toolbar/js/history.js
@@ -104,4 +104,5 @@ $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     refreshHistory();
 });
+// As we don't refresh the whole toolbar each fetch or ajax request we need to refresh the history when we open the panel
 $$.onPanelRender(djDebug, "HistoryPanel", refreshHistory);
--- a/debug_toolbar/static/debug_toolbar/js/history.js
+++ b/debug_toolbar/static/debug_toolbar/js/history.js
@@ -76,6 +76,7 @@ function switchHistory(newStoreId) {
     const formTarget = djDebug.querySelector(
         ".switchHistory[data-store-id='" + newStoreId + "']"
     );
+    djdt.history_storeId = newStoreId
     const tbody = formTarget.closest("tbody");
 
     const highlighted = tbody.querySelector(".djdt-highlighted");
@@ -104,3 +105,5 @@ $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     refreshHistory();
 });
+
+export { refreshHistory };
--- a/debug_toolbar/static/debug_toolbar/js/history.js
+++ b/debug_toolbar/static/debug_toolbar/js/history.js
@@ -76,7 +76,6 @@ function switchHistory(newStoreId) {
     const formTarget = djDebug.querySelector(
         ".switchHistory[data-store-id='" + newStoreId + "']"
     );
-    djdt.history_storeId = newStoreId
     const tbody = formTarget.closest("tbody");
 
     const highlighted = tbody.querySelector(".djdt-highlighted");
@@ -105,5 +104,4 @@ $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     refreshHistory();
 });
-
-export { refreshHistory };
+$$.onPanelRender(djDebug, "HistoryPanel", refreshHistory);
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -19,7 +19,7 @@ const djdt = {
     handleDragged: false,
     init() {
         const djDebug = getDebugElement();
-        window.djdt.update_on_ajax = djDebug.getAttribute("update-on-ajax") === "True"
+        window.djdt.update_on_ajax = djDebug.getAttribute("update-on-ajax") === "True";
         $$.on(djDebug, "click", "#djDebugPanelList li a", function (event) {
             event.preventDefault();
             if (!this.className) {
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -1,4 +1,5 @@
 import { $$, ajax, replaceToolbarState, debounce } from "./utils.js";
+import { refreshHistory } from "./history.js"
 
 function onKeyDown(event) {
     if (event.keyCode === 27) {
@@ -19,6 +20,7 @@ const djdt = {
     handleDragged: false,
     init() {
         const djDebug = getDebugElement();
+        console.log(djDebug)
         $$.on(djDebug, "click", "#djDebugPanelList li a", function (event) {
             event.preventDefault();
             if (!this.className) {
@@ -49,12 +51,16 @@ const djdt = {
                         inner.previousElementSibling.remove(); // Remove AJAX loader
                         inner.innerHTML = data.content;
                         $$.executeScripts(data.scripts);
+                        if (panelId == "HistoryPanel"){
+                            refreshHistory();
+                        }
                         $$.applyStyles(inner);
                         djDebug.dispatchEvent(
                             new CustomEvent("djdt.panel.render", {
                                 detail: { panelId: panelId },
                             })
                         );
+
                     });
                 } else {
                     djDebug.dispatchEvent(
@@ -62,6 +68,9 @@ const djdt = {
                             detail: { panelId: panelId },
                         })
                     );
+                    if (panelId == "HistoryPanel"){
+                            refreshHistory();
+                        }
                 }
             }
         });
@@ -274,7 +283,9 @@ const djdt = {
             storeId = encodeURIComponent(storeId);
             const dest = `${sidebarUrl}?store_id=${storeId}`;
             slowjax(dest).then(function (data) {
-                replaceToolbarState(storeId, data);
+                if (window.djdt.history_storeId == null){
+                    replaceToolbarState(storeId, data);
+                }
             });
         }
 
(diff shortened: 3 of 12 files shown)


debug_toolbar/settings.py
```diff
@@ -42,6 +42,7 @@
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
     "TOOLBAR_LANGUAGE": None,
+    "UDPATE_ON_AJAX": False,
```
matthiask, line 45: Typo (`UPDATE` not `UDPATE`)

Also, I think the setting should use `FETCH`, not `AJAX`, to reference the newer and preferred API.

debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -19,6 +19,7 @@ const djdt = {
     handleDragged: false,
     init() {
         const djDebug = getDebugElement();
+        window.djdt.update_on_ajax = djDebug.getAttribute("update-on-ajax") === "True";
```
matthiask, line 22: It's not completely consistent but we generally prefer using `camelCase` in JavaScript code.

Also, other parts of the code use the dataset attribute, so this should be `djDebug.dataset.updateOnAjax` and `data-update-on-ajax` (or whatever the name of the attribute is). A `data-` prefix for attributes is a good idea anyway because such attributes are always valid HTML.

I wonder if there is a better place for this property though. Tim suggested using localStorage to allow switching between the auto update and manual update behavior on the fly, and this probably wouldn't work with the way it's proposed here. Also, I don't think we want to expose the attribute to external modification (just yet). Putting it in the same place as the `handleDragged` property may be a good idea?

debug_toolbar/static/debug_toolbar/js/history.js
```diff
@@ -104,3 +104,4 @@ $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     refreshHistory();
 });
+$$.onPanelRender(djDebug, "HistoryPanel", refreshHistory);
```
matthiask, line 107: Can you add a comment for this change? I'm not sure why we need this.

matthiask (commented):
Thank you for working on this!

debug_toolbar/static/debug_toolbar/js/toolbar.js
tim-schilling: > Tim suggested using localStorage to allow switching between the auto update and manual update behavior on the fly, and this probably wouldn't work with the way it's proposed here.

For some context, when Elineda and I spoke yesterday, we decided we could start with a settings control. If we find that it doesn't meet developers needs, we can add the JS to make it more stateful.

matthiask: Makes sense!

elineda: I did change asked, but I don't put on a localstorage, I just with handleDragged.

debug_toolbar/static/debug_toolbar/js/history.js
elineda: Done. We need this because the en ajax request occurs we don't necessary refresh the toolbar. In the first load on the history panel we haven't problem, but after the list don't auto udpate. We can click on refresh button but I think it's better to update the list each render of history.

debug_toolbar/static/debug_toolbar/js/history.js
```diff
@@ -104,3 +104,5 @@ $$.on(djDebug, "click", ".refreshHistory", function (event) {
     event.preventDefault();
     refreshHistory();
 });
+// As we don't refresh the whole toolbar each fetch or ajax request we need to refresh the history when we open the panel
```
tim-schilling, line 107:
```suggestion
// We don't refresh the whole toolbar each fetch or ajax request,
// so we need to refresh the history when we open the panel
```
Small grammar change and updating the line breaks.

debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -55,6 +57,7 @@ const djdt = {
                                 detail: { panelId: panelId },
                             })
                         );
+
```
tim-schilling, line 60:
```suggestion
```
Removing unnecessary whitespace change. (you can ignore these, I'll batch the changes together)

docs/changes.rst
```diff
@@ -19,6 +19,7 @@ Pending
   <https://astral.sh/blog/the-ruff-formatter>`__.
 * Changed the default position of the toolbar from top to the upper top
   position.
+* Add a setting for refresh or not the toolbar if an ajax request occurs.
```
tim-schilling, line 22:
```suggestion
* Added the setting, ``UPDATE_ON_FETCH`` to control whether the 
  toolbar automatically updates to the latest AJAX request or not.
  It defaults to ``False``.
```
Change the tense, added the setting name for folks and mentioned the default value.

docs/configuration.rst
```diff
@@ -163,6 +163,15 @@ Toolbar options
...
   ``"en-us"`` and :setting:`LANGUAGE_CODE` to ``"fr"``.
 
+.. _UPDATE_ON_FETCH:
+
+* ``UPDATE_ON_FETCH``
+
+  Default: ``False``
+
+  This setting specifies if  the whole toolbar need to be refresh if an ajax
+  request is done.
```
tim-schilling, line 173:
```suggestion
  This controls whether the toolbar should update to the latest AJAX request when it occurs.
  This is useful to enable if when "boosting" the web application with a JavaScript library such
  as htmx.
```

tim-schilling:
@matthiask can you review the documentation suggestions?

docs/configuration.rst
```diff
@@ -163,6 +163,16 @@ Toolbar options
...
 
+.. _UPDATE_ON_FETCH:
+
+* ``UPDATE_ON_FETCH``
+
+  Default: ``False``
+
+  This controls whether the toolbar should update to the latest AJAX request when it occurs.
+  This is useful to enable if when "boosting" the web application with a JavaScript library such
+  as htmx.
```
tim-schilling, line 174:
```suggestion
  This controls whether the toolbar should update to the latest AJAX
  request when it occurs. This is useful to enable if when "boosting"
  the web application with a JavaScript library such as htmx.
```

debug_toolbar/templates/debug_toolbar/base.html
```diff
@@ -16,7 +16,7 @@
      data-sidebar-url="{{ history_url }}"
      {% endif %}
      data-default-show="{% if toolbar.config.SHOW_COLLAPSED %}false{% else %}true{% endif %}"
-     {{ toolbar.config.ROOT_TAG_EXTRA_ATTRS|safe }}>
+     {{ toolbar.config.ROOT_TAG_EXTRA_ATTRS|safe }}  data-update-on-fetch="{{ toolbar.config.UPDATE_ON_FETCH|safe }}">
```
matthiask, line 19:
```suggestion
     {{ toolbar.config.ROOT_TAG_EXTRA_ATTRS|safe }}  data-update-on-fetch="{{ toolbar.config.UPDATE_ON_FETCH }}">
```

I'm not 100% sure but I don't think `|safe` is necessary.

Also, you could use `|yesno:'true,false'` or do it as `data-default-show` so that the value is lowercased already, but I'm just suggesting that, not asking for anything.

docs/configuration.rst
```diff
@@ -163,6 +163,16 @@ Toolbar options
...
 
+.. _UPDATE_ON_FETCH:
+
+* ``UPDATE_ON_FETCH``
+
+  Default: ``False``
+
+  This controls whether the toolbar should update to the latest AJAX
+  request when it occurs. This is useful to enable if when "boosting"
+  the web application with a JavaScript library such as htmx.
```
matthiask, line 174: I like the changelog entry a lot. "to control whether the toolbar automatically updates to the latest AJAX request or not" reads beautifully. 

The second sentence is a bit strange because of the "if when". Maybe something like this would be better? "This is especially useful when using htmx boosting or similar JavaScript techniques."

I'm not a native speaker/writer either though, so I'm not sure.

matthiask (approved):
I like it! Only small suggestions this time.

docs/configuration.rst
tim-schilling: "if when" is definitely a mistake :sweat_smile: Your English is perfect.

debug_toolbar/templates/debug_toolbar/base.html
elineda: I removed safe. But I didn't user yesno because the result of this will be a string and we will need do to the same thing in js for checking it lowercased or not.

tim-schilling:
@matthiask I'm realizing I broke from our typical merge process. @elineda asked in private if she could merge, I said yes.

matthiask:
@tim-schilling No worries. I was a little bit surprised but also relieved. I think merging faster is fine, really, especially in Jazzband.

living180:
This is a great change, thanks @elineda!  I did notice that this seems to duplicate the functionality of the `OBSERVE_REQUEST_CALLBACK` setting, though in a much more discoverable and usable way.  I did some looking at the code, and now that the `UPDATE_ON_FETCH` setting exists, I cannot come up with a reason for `OBSERVE_REQUEST_CALLBACK` to exist.  If an AJAX request should not be instrumented, `SHOW_REQUEST_CALLBACK` can be used to block that rather than `OBSERVE_REQUEST_CALLBACK` (which doesn't actually block the instrumentation, but instead simply prevents an AJAX request from showing up in the history panel until the next full page load).  Should `OBSERVE_REQUEST_CALLBACK` be deprecated in favor of `UPDATE_ON_FETCH`?

matthiask:
I went back and read the discussion in https://github.com/jazzband/django-debug-toolbar/pull/1577

I think I agree with your assessment. I remember that I had some problems understanding what `OBSERVE_REQUEST_CALLBACK` actually meant, and having only `SHOW_REQUEST_CALLBACK` and `UPDATE_ON_FETCH` at some point in the future clarifies things a lot for me.

living180:
OK, so I just played around with this feature some more, and either I'm doing something wrong, or it is not working the way I expect.  I confirmed that I have DjDT 4.3.0 using the Versions panel, and I have an explicit `UPDATE_ON_FETCH: False,` in my `DEBUG_TOOLBAR_CONFIG` (which should be the same as the default).  However, with the Debug Toolbar sidebar expanded, if I trigger an AJAX request, the sidebar still updates to show the data of that AJAX request.  The behavior that I want, and was hoping for from `UPDATE_ON_FETCH` is that the Debug Toolbar would show the data for the request of the top-level page until I specifically select a different request from the History panel.

I am currently able to achieve that desired behavior by including `'OBSERVE_REQUEST_CALLBACK': lambda request: False,` in my `DEBUG_TOOLBAR_CONFIG`.  Is there something important I am missing?

tim-schilling:
`OBSERVE_REQUEST_CALLBACK` is meant to identify which AJAX requests need to be instrumented. That setting may not be needed, but it is different from `UPDATE_ON_FETCH`. `UPDATE_ON_FETCH` is intended to determine whether or not the toolbar will automatically update with the latest AJAX request or not. As you said @living180, it should work the way you expect it to. I'll take a closer look.

tim-schilling:
@living180 I can't reproduce the issue you're describing. When I run `make example`, then go to http://127.0.0.1:8000/ and click on either of the increment buttons, the toolbar doesn't update with the latest request. When I set `UPDATE_ON_FETCH` to False, it still doesn't update. When I set `UPDATE_ON_FETCH` to True, it does update. When I browse to http://127.0.0.1:8000/htmx/boost/ it works as expected as well. Would you be able to produce a reproducible example for me?

living180:
Sure, it's late where I am, but I'll try to reproduce with a simpler example hopefully tomorrow.

living180:
Unless I'm really reading something wrong, `DebugToolbarMiddleware` never instruments its own requests:

https://github.com/jazzband/django-debug-toolbar/blob/c688ce4ad7d18c5ecb800869298ca8cf6c08be1d/debug_toolbar/middleware.py#L48-L49

In that case the `OBSERVE_REQUEST_CALLBACK` callback never gets called, because it only gets called via the `HistoryPanel.get_headers()` method:

https://github.com/jazzband/django-debug-toolbar/blob/c688ce4ad7d18c5ecb800869298ca8cf6c08be1d/debug_toolbar/panels/history/panel.py#L23-L29

which is only called by `DebugToolbarMiddleware` after the early return show above:

https://github.com/jazzband/django-debug-toolbar/blob/c688ce4ad7d18c5ecb800869298ca8cf6c08be1d/debug_toolbar/middleware.py#L76-L77

So no matter how `SHOW_TOOLBAR_CALLBACK` or `OBSERVE_REQUEST_CALLBACK` are defined, the toolbar will never instrument its own requests, AJAX or otherwise.

This also means that the default definition of `OBSERVE_REQUEST_CALLBACK` should simply be `return True` because it will never be called if `DebugToolbar.is_toolbar_request(request)` returns `False`.

tim-schilling:
Ack, good point. I wasn't digging deep enough. I agree that with the new setting, we could direct people to using `SHOW_TOOLBAR_CALLBACK` instead of using `OBSERVE_REQUEST_CALLBACK`.

dangelsaurus:
I appreciate this enhancement, but very curious why the default is set to False when the historical behavior has been to update with an ajax request?  If someone (me) misses this in the changelog, it can make for an annoying hour of troubleshooting. This is not a developer friendly attitude.

tim-schilling:
There wasn't a great choice for us here. Your experience is the worse case for changing the default. However, that's less worse than the original bug request where the default cause hijacks the toolbar without consent.

I'm sorry you had to deal with the difficult consequences of that choice.

tim-schilling:
The release notes could have better called it out though.

elineda:
If I'm remember well, this pr was the reason to not put a minor version on this release.

Issue #1813, opened 2023-07-09
SQL Panel DeprecationWarning - The "default.html" templates for forms and formsets will be removed


bluetech:
When I open the SQL Panel on Django 4.2, django-debug-toolbar 4.1, I get the following warning:

> django.utils.deprecation.RemovedInDjango50Warning: The "default.html" templates for forms and formsets will be removed. These were proxies to the equivalent "table.html" templates, but the new "div.html" templates will be the default from Django 5.0. Transitional renderers are provided to allow you to opt-in to the new output style now. See https://docs.djangoproject.com/en/4.2/releases/4.1/ for more details

I tried to narrow down a bit what's causing it using pdb, but I only managed to see that the `self` is `SignedDataForm`, not much more.

living180:
I looked at this awhile back, and AFAICT, the warning is actually generated by an internal compatibility shim in Django.  It should go away once the compatibility shim is removed.  Although the Debug Toolbar could be updated to avoid the warning, any changes would be unnecessary once targeting Django 5.0.  Thus I decided that it wasn't worth code churn on the Debug Toolbar end.  If anyone would like more details, let me know, and I can try to gather them together in the next few days.

bluetech:
Thank you, makes sense. I silenced the warning for now.

matthiask:
Fixed #1878

PR #1878, opened 2024-02-08
Fix DeprecationWarnings about form default templates

pfouque:
# Description

Improve [Django5 compatibility](https://github.com/jazzband/django-debug-toolbar/pull/1839) by [solving warnings related to form default templates that will change with Django5 from `table` to `div`](https://github.com/jazzband/django-debug-toolbar/issues/1813)

Fixes # (issue)

https://github.com/jazzband/django-debug-toolbar/issues/1813

# Checklist:


matthiask (commented):
Are we rendering table rows right now also?

`as_table` doesn't render the surrounding `<table>` element.

```django/forms/utils.py:89:    def as_table(self):
django/forms/utils.py-90-        """Render as <tr> elements excluding the surrounding <table> tag."""
django/forms/utils.py-91-        return self.render(self.template_name_table)
```

The HTML generated here isn't correct it seems to me.

pfouque:
Weirdly, during my tests, both `as_div` and `as_table` were only rendering the `<input>`.
But I agree it would probably make more sense to use `as_div`

tim-schilling:
It's likely because there's no need to render any additional html around a hidden input so both `.as_div` and `.as_table` are resulting in the same output. If we were to add another field input, then the problem would be realized.

matthiask:
Maybe the time has come for us to move to `Django>=4.2`?

Django 3.2 LTS support ends this April https://www.djangoproject.com/download/

tim-schilling:
Let's confirm there isn't anything wrong with update on fetch with @living180 first. If there isn't, I'm good with dropping 3.2 early.

matthiask:
Yes, that's a good idea.

I have submitted https://github.com/jazzband/django-debug-toolbar/pull/1880 anyway.

Issue #1881, opened 2024-02-12
Jinja2 Support Broken


ae99:
Hey! Spent some time trying to get Jinja2 templates to appear in the debug toolbar correctly. Templates tab was simply saying "No Templates". This had me stumped as there's a test case for this in the codebase.

Upon some further digging, I think the Jinja2 test in `tests/templates/jinja2/basic.jinja` is actually being rendered as a Django template, not through Jinja 2. Modifying that template to contain:
```
{% extends 'base.html' %}
{% block content %}Test for {{ title }} (Jinja)
{% for i in range(10) %}{{ i }}{% endfor %}
{% endblock %}
```
Causes the test to fail with Django throwing `django.template.exceptions.TemplateSyntaxError` - even though this is valid Jinja and Django's template shouldn't be involved, so this test isn't actually using Jinja2.

---
It appears if you *are* using Jinja2 to render templates, DDT isn't able to pick up what goes on inside them, at least with default configurations. Any pointers on how I could get this working for myself would be great!

PR #1882, opened 2024-02-12
Actually use the Jinja2 template backend

matthiask:
Refs #1881.

# Description

this information as well.

Fixes # (issue)

# Checklist:

### DIFF
--- a/debug_toolbar/panels/templates/jinja2.py
+++ b/debug_toolbar/panels/templates/jinja2.py
@@ -1,0 +1,23 @@
+import functools
+
+from django.template.backends.jinja2 import Template as JinjaTemplate
+from django.template.context import make_context
+from django.test.signals import template_rendered
+
+
+def patch_jinja_render():
+    orig_render = JinjaTemplate.render
+
+    @functools.wraps(orig_render)
+    def wrapped_render(self, context=None, request=None):
+        # This patching of render only instruments the rendering
+        # of the immediate template. It won't include the parent template(s).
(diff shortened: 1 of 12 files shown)


tim-schilling:
I pushed this forward for the next person. Here's my message from the last commit:

> It instruments the single template render, but not the inherited
templates and I'm guessing not the included templates either.
I suspect we're going to have to patch jinja templates more robustly
than relying on the django jinja backend template class.

matthiask:
@tim-schilling It would be great if we were able to inspect all of Jinja2's rendering but maybe that's harder because of Jinja2's template compilation, or what do you think? I didn't do much research here, so I'm basically just speculating.

I think what we have with your changes is already much better than the status quo. We could merge this, document the limitations and move on?

AlexCLeduc:
For what it's worth, this patch on its own would be very useful for my jinja projects. I very rarely use the toolbar to inspect parents/included templates.

tim-schilling:
No, it is better. I'd prefer to document the caveat better for sure, but if it helps others that's still a win.

tim-schilling:
@matthiask this is ready for your review. I don't think you'll be able to give it a green checkmark since you opened the PR.

tim-schilling:
I'm going to assume your previous comments were tacit approval. I'm about to head out for the day and I'd like to get this in for 4.4.4 with #1946

matthiask:
Thank you! I'm late to the party today.

PR #1887, opened 2024-02-17
Show toolbar for docker's internal IP address

### LINKED ISSUE #1854: The Debug Toolbar has stopped showing under Docker
It seems like the [snippet of code](https://django-debug-toolbar.readthedocs.io/en/stable/installation.html#configure-internal-ips) in the documentation for using Docker has stopped working.

image

That snippet of code will set `INTERNAL_IPS` to `['172.27.0.1', '127.0.0.1', '10.0.2.2']`, however, inside `debug_toolbar.middleware.show_toolbar()`, `request.META.get("REMOTE_ADDR")` returns `'192.168.65.1'`, which is why it ends up not showing the Debug Toolbar.

This snippet of code used to work. I'm guessing something must of changed in a recent version of Docker to cause it to not get the correct IP address. I'm not sure at this point how to get the IP dynamically because even `ifconfig` shows the `172.27.01` IP address instead of the `192.168.65.1` IP address.

Versions:
* Docker version 24.0.6
* Docker Compose version v2.22.0-desktop.2
* macOS Sonoma version 14.0
* django-debug-toolba v3.8.1
* django v4.2.5

tim-schilling:
# Description
This attempts to automatically look up the docker internal IP address for the developer.

Fixes #1854

# Checklist:

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -3,6 +3,7 @@ Debug Toolbar middleware
 """
 
 import re
+import socket
 from functools import lru_cache
 
 from django.conf import settings
@@ -19,7 +20,22 @@ def show_toolbar(request):
     """
     Default function to determine whether to show the toolbar on a given page.
     """
-    return settings.DEBUG and request.META.get("REMOTE_ADDR") in settings.INTERNAL_IPS
+    internal_ips = settings.INTERNAL_IPS.copy()
+
+    try:
+        # This is a hack for docker installations. It attempts to look
+        # up the IP address of the docker host.
+        # This is not guaranteed to work.
+        docker_ip = (
+            # Convert the last segment of the IP address to be .1
+            ".".join(socket.gethostbyname("host.docker.internal").rsplit(".")[:-1])
+            + ".1"
+        )
+        internal_ips.append(docker_ip)
+    except socket.gaierror:
+        # It's fine if the lookup errored since they may not be using docker
+        pass
+    return settings.DEBUG and request.META.get("REMOTE_ADDR") in internal_ips
 
 
 @lru_cache(maxsize=None)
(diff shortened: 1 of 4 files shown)


tim-schilling:
@epicserve could you give this a test run for me?

epicserve:
@tim-schilling,

I tested it out and it worked. You can see my code [here](https://github.com/epicserve/django-base-site/compare/experiment/djdt-docker-fix).

nitairoy:
If you're developing with a Django server in a Docker container with docker, the instructions for enabling the toolbar don't work. The reason is related to the fact that the actual address that you would need to add to INTERNAL_IPS is going to be something dynamic, like 172.24.0.1. Rather than trying to dynamically set the value of INTERNAL_IPS, the straightforward solution is to replace the function that enables the toolbar, in your settings.py, for example:

DEBUG_TOOLBAR_CONFIG = {
    'SHOW_TOOLBAR_CALLBACK': lambda _request: DEBUG
}

This should also work for other dynamic routing situations, like Vagrant or Heroku.

follow the link: https://stackoverflow.com/questions/10517765/django-debug-toolbar-not-showing-up

tim-schilling:
@nitairoy did you attempt to use the code in this PR? Also, please see the discussion in #1854.

epicserve:
@tim-schilling, mocking is how I would handle testing it. I can't think of a better alternative.

nitairoy:
Yes, I did and it work.

Best Regards,
Nitai Roy

matthiask:
> Excellent thank you. @matthiask I know this is a hack, but I personally don't see a difference between asking a developer to copy/paste the hack themselves versus us writing it in for them and letting them know where it exists. At least for a development tool like this.

Yes, I agree. It would certainly be nice if the toolbar worked out of the box for everyone.

johnthagen:
For historical purposes, this caused a regression

- https://github.com/jazzband/django-debug-toolbar/issues/1927
- https://github.com/moby/moby/issues/48266

Fix in

- #1929
- Released in 4.4.3

Issue #1405, opened 2020-11-21
'djdt' is not a registered namespace


convers39:
I got this error when I **run django unittest**, I have tried all the answers I could find, seems not working.
Here are my **settings:**

```#settings.py
DEBUG_TOOLBAR_PATCH_SETTINGS = False
if DEBUG:
    INSTALLED_APPS += ['debug_toolbar']
    MIDDLEWARE += ['debug_toolbar.middleware.DebugToolbarMiddleware', ]
    DEBUG_TOOLBAR_CONFIG = {
        'SHOW_TOOLBAR_CALLBACK': lambda _request: DEBUG
    }
```
Project **url file**:
```# urls.py
if settings.DEBUG:
    import debug_toolbar
    urlpatterns += [path('__debug__/', include(debug_toolbar.urls)), ]\
        + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```

**If I turn off the debug to False manually, then the error is gone**, if I just run manage.py test with all those settings above or either same as the official doc , then I get the this error.
Is there any method to avoid this error? **Doesn't the debug suppose to be False when running a test?**
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/template/base.py", line 829, in _resolve_lookup
    current = current[bit]
TypeError: 'HistoryPanel' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/urls/base.py", line 72, in reverse
    extra, resolver = resolver.namespace_dict[ns]
KeyError: 'djdt'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/src/app/apps/shop/tests/test_views.py", line 63, in test_shop_list_view_sort_GET
    res = self.client.get(reverse('shop:product-list')+'?sorting=sales')
  File "/usr/local/lib/python3.8/site-packages/django/test/client.py", line 733, in get
    response = super().get(path, data=data, secure=secure, **extra)
  File "/usr/local/lib/python3.8/site-packages/django/test/client.py", line 393, in get
    return self.generic('GET', path, secure=secure, **{
  File "/usr/local/lib/python3.8/site-packages/django/test/client.py", line 470, in generic
    return self.request(**r)
  File "/usr/local/lib/python3.8/site-packages/django/test/client.py", line 710, in request
    self.check_exception(response)
[...]

PR #1889, opened 2024-02-27
'djdt' is not a registered namespace #1405

VeldaKiara:
# Description

The error was `'djdt' is not a registered namespace`. It occurred due to misconfiguration of the toolbar settings as well as the debug settings.

Fixes # ([issue 1405](https://github.com/jazzband/django-debug-toolbar/issues/1405))
- Added a check to result in an official error 
- Added tests for the check
- Updated the docs with the right way to configure the toolbar to work with debug without an issue

# Checklist:

### DIFF
--- a/debug_toolbar/apps.py
+++ b/debug_toolbar/apps.py
@@ -1,9 +1,10 @@
 import inspect
 import mimetypes
+import sys
 
 from django.apps import AppConfig
 from django.conf import settings
-from django.core.checks import Warning, register
+from django.core.checks import Error, Warning, register
 from django.middleware.gzip import GZipMiddleware
 from django.utils.module_loading import import_string
 from django.utils.translation import gettext_lazy as _
@@ -206,3 +207,23 @@ def js_mimetype_check(app_configs, **kwargs):
             )
         ]
     return []
+
+
+@register()
+def debug_toolbar_namespace_check(app_configs, **kwargs):
+    """
+    Check that the Django Debug Toolbar is registered as a namespace.
+    """
+    if (
+        settings.DEBUG is False
+        and dt_settings.get_config()["DEBUG_TOOLBAR_CONFIG"] == "test" in sys.argv
+    ):
+        return [
+            Error(
+                "django debug toolbar  is not a registered namespace ",
+                hint="The Django Debug Toolbar is misconfigured, check the documentation for proper configuration and run the tests.",
+                id="debug_toolbar.W008",
+            )
+        ]
+    else:
+        return []
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -1,3 +1,4 @@
+import sys
 import warnings
 from functools import lru_cache
 
@@ -42,6 +43,7 @@ CONFIG_DEFAULTS = {
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
     "TOOLBAR_LANGUAGE": None,
+    "DEBUG_TOOLBAR_CONFIG": "test" in sys.argv,
 }
 
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -19,6 +19,8 @@ Pending
   <https://astral.sh/blog/the-ruff-formatter>`__.
 * Changed the default position of the toolbar from top to the upper top
   position.
+* Fixed the bug causing ``'djdt' is not a registered namespace`` and updated
+  docs to help in initial configuration.
 
 4.2.0 (2023-08-10)
 ------------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -17,6 +17,8 @@ Pending
   ``django.contrib.admindocs.utils.get_view_name``.
 * Switched from black to the `ruff formatter
   <https://astral.sh/blog/the-ruff-formatter>`__.
+* Changed the default position of the toolbar from top to the upper top
+  position.
 
 4.2.0 (2023-08-10)
 ------------------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -99,7 +99,34 @@ Add django-debug-toolbar's URLs to your project's URLconf:
 This example uses the ``__debug__`` prefix, but you can use any prefix that
 doesn't clash with your application's URLs.
 
-5. Add the Middleware
+5. Configure django-debug-toolbar
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Add the following configuration for django-debug-toolbar in your setting:
+
+.. code-block:: python
+
+    import sys
+
+    DEBUG = bool({"runserver"}.intersection(sys.argv))
+
+    DEBUG_TOOLBAR_CONFIG = "test" in sys.argv
+
+    DEBUG_TOOLBAR = DEBUG and not DEBUG_TOOLBAR_CONFIG
+
+
+To explicitly check the ``settings.DEBUG_TOOLBAR_CONFIG``variable before
+including the URLs, add the following code in your ``urls`` file.
+
+.. code-block:: python
+
+    if settings.DEBUG_TOOLBAR_CONFIG:
+
+        import debug_toolbar
+
+        urlpatterns += [path('__debug__/', include(debug_toolbar.urls))]
+
+6. Add the Middleware
 ^^^^^^^^^^^^^^^^^^^^^
 
 The Debug Toolbar is mostly implemented in a middleware. Add it to your
@@ -122,7 +149,7 @@ The Debug Toolbar is mostly implemented in a middleware. Add it to your
 
 .. _internal-ips:
 
-6. Configure Internal IPs
+7. Configure Internal IPs
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The Debug Toolbar is shown only if your IP address is listed in Django’s
--- a/tests/test_checks.py
+++ b/tests/test_checks.py
@@ -1,11 +1,15 @@
 import os
+import sys
 import unittest
 from unittest.mock import patch
 
 import django
 from django.conf import settings
-from django.core.checks import Warning, run_checks
+from django.core.checks import Error, Warning, run_checks
 from django.test import SimpleTestCase, override_settings
+
+from debug_toolbar import settings as dt_settings
+from debug_toolbar.apps import debug_toolbar_namespace_check
 
 PATH_DOES_NOT_EXIST = os.path.join(settings.BASE_DIR, "tests", "invalid_static")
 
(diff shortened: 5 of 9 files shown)


debug_toolbar/settings.py
```diff
@@ -42,6 +43,7 @@
     "SQL_WARNING_THRESHOLD": 500,  # milliseconds
     "OBSERVE_REQUEST_CALLBACK": "debug_toolbar.toolbar.observe_request",
     "TOOLBAR_LANGUAGE": None,
+    "DEBUG_TOOLBAR_CONFIG": "test" in sys.argv,
```
tim-schilling, line 46: I don't think we should use `DEBUG_TOOLBAR_CONFIG as the name for this setting. It should be related to the fact that the application is running the test suite. Some other options are:

- TESTING
- RUNNING_TESTS
- IS_TESTING
- TEST

debug_toolbar/apps.py
```diff
@@ -206,3 +207,23 @@ def js_mimetype_check(app_configs, **kwargs):
...
+
+
+@register()
+def debug_toolbar_namespace_check(app_configs, **kwargs):
+    """
+    Check that the Django Debug Toolbar is registered as a namespace.
+    """
+    if (
+        settings.DEBUG is False
+        and dt_settings.get_config()["DEBUG_TOOLBAR_CONFIG"] == "test" in sys.argv
```
tim-schilling, line 219:
```suggestion
        and dt_settings.get_config()["DEBUG_TOOLBAR_CONFIG"]
```
The check on `sys.argv` is handled in the settings and is unnecessary to do here.

debug_toolbar/apps.py
```diff
@@ -206,3 +207,23 @@ def js_mimetype_check(app_configs, **kwargs):
             )
         ]
     return []
+
+
+@register()
+def debug_toolbar_namespace_check(app_configs, **kwargs):
```
tim-schilling, line 213: While the namespace issue is what the problem presents as, what we're really checking is if the toolbar is installed when running tests. This function name, doc string and changelog entry should reflect that.

debug_toolbar/settings.py
VeldaKiara: I think I will change it to `RUNNING_TESTS`

debug_toolbar/apps.py
VeldaKiara: so instead of calling it namespace check I should call it debug_tolbar_installed_when_running_Tests

debug_toolbar/apps.py
VeldaKiara: Since its a condition I think it still needs to be there. I tried to remove it and got the djdt error.

debug_toolbar/apps.py
```diff
@@ -206,3 +207,23 @@ def js_mimetype_check(app_configs, **kwargs):
...
+
+
+@register()
+def debug_toolbar_installed_when_running_tests_check(app_configs, **kwargs):
+    """
+    Check that the Django Debug Toolbar is installed and works well with the current tests
+    """
+    if (
+        settings.DEBUG is False
+        and dt_settings.get_config()["RUNNING_TESTS"] == "test" in sys.argv
```
elineda, line 219: you don't need == "test" in sys.argv because you set already in debug_toolbar/settings.py:46 (if you print it you can check the value) also you can just put "not settings.DEBUG" it's already a bool too.

docs/installation.rst
```diff
@@ -99,7 +99,34 @@ Add django-debug-toolbar's URLs to your project's URLconf:
...
+5. Configure django-debug-toolbar
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Add the configuration for django-debug-toolbar in your setting to run tests:
+
+.. code-block:: python
+
+    import sys
+
+    DEBUG = bool({"runserver"}.intersection(sys.argv))
```
elineda, line 111: I don't like it. On my app the DEBUG is a bool fixed by the user. I can do bool even if I don't use the runserver.

elineda (commented):
The test with make test doesn't work, you need to change the setting on test to change accept test of the debug toolbar 
```
DEBUG_TOOLBAR_CONFIG = {
    # Django's test client sets wsgi.multiprocess to True inappropriately
    "RENDER_PANELS": False,
    "RUNNING_TESTS": False
}
```

docs/installation.rst
VeldaKiara: What suggestion do you have?

debug_toolbar/apps.py
VeldaKiara: Let me test this out?

docs/installation.rst
VeldaKiara: @matthiask I borrowed this from a solution you provided [here](https://github.com/matthiask/workbench/blob/5e3663cb7c28cec707cfe746e9072e4fa6778f30/workbench/settings.py#L16-L17)

I am open to suggestions of how to make it better.

VeldaKiara: > DEBUG = server_settings.get('DEBUG', False) if you extract the settings on a file (I do with a yaml) or DEBUG = os.environ.get("DEBUG", False) if you want to use an env variable

thanks let me test this out

VeldaKiara: The previous solution was more like this 
`DEBUG = env("DEBUG", default=bool({"runserver"}.intersection(sys.argv)))`

I remove the env

tim-schilling: I think we can leave the DEBUG setting as-is in the instructions. The user can define how they want to configure that on their own.

matthiask: @VeldaKiara Yes, I have used that pattern for some time. I am now requiring that `DEBUG` is set in the environment in all new projects, it works better for me. It's maybe more annoying to start, but I like the fact that it's explicit instead of magic.

VeldaKiara:
> @elineda @matthiask @tim-schilling could you test it out and let me know if everything is good and any changes or suggestions you have at this point?

Is it okay to merge?

tim-schilling:
@VeldaKiara no, we have some miscommunications yet. I started on a change that will better communicate them. I'll try to push that up tonight.

VeldaKiara:
> @VeldaKiara no, we have some miscommunications yet. I started on a change that will better communicate them. I'll try to push that up tonight.

Awesome, I will be waiting on the changes.

tim-schilling:
@VeldaKiara I pushed some changes to the check and the settings in the example app. Let me know what you think.

I suspect we should ignore our new check in the test settings until we test it explicitly. In that case we should use the `@override_settings` decorator to enable it. That should get the tests to pass.

Regarding the documentation, maybe we keep the docs as they are, then link them to the example app to see how it's setup in an advanced configuration?

VeldaKiara:
Let me check out the changes.

I think that would work to pointing to the example app

VeldaKiara:
I have incorporated the changes requested. 
Let me know if you have any more suggestions.

tests/test_checks.py
```diff
@@ -258,3 +261,57 @@ def test_check_w007_invalid(self, mocked_guess_type):
...
+                    hint="Django changes the DEBUG setting to False when running "
+                    "tests. By default the Django Debug Toolbar is installed because "
+                    "DEBUG is set to True. For most cases, you need to avoid installing "
+                    "the toolbar when running tests. If you feel this check is in error, "
+                    "you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to "
+                    "bypass this check.",
+                    id="debug_toolbar.W008",
+                )
+            ],
+        )
```
tim-schilling, line 284:
```suggestion
```
This can be removed since it's covered below.

tests/test_checks.py
```diff
@@ -258,3 +261,57 @@ def test_check_w007_invalid(self, mocked_guess_type):
...
+
+    @patch("debug_toolbar.apps.debug_toolbar_installed_when_running_tests_check", [])
```
tim-schilling, line 286:
```suggestion
```
We don't want to patch the function we're trying to test.

docs/installation.rst
```diff
@@ -99,6 +99,12 @@ Add django-debug-toolbar's URLs to your project's URLconf:
 This example uses the ``__debug__`` prefix, but you can use any prefix that
 doesn't clash with your application's URLs.
 
+> [!NOTE]
+> Check  out the configuration example in the
+[example app](https://github.com/jazzband/django-debug-toolbar/tree/main/example)
+to learn how to set up the toolbar to function smoothly while running your
+tests.
```
tim-schilling, line 106: We're using restructured text which has a different syntax than markdown. You'll want to use a note admonition which you can read about here: https://docutils.sourceforge.io/docs/ref/rst/directives.html#note

docs/installation.rst
```diff
@@ -99,6 +99,12 @@ Add django-debug-toolbar's URLs to your project's URLconf:
 This example uses the ``__debug__`` prefix, but you can use any prefix that
 doesn't clash with your application's URLs.
 
+> [!NOTE]
```
tim-schilling, line 102: Looking at everything altogether, it's probably better to call out this difference in step 3 "Install the app". This is because we want to let them know earlier on, there's a different way to configure the toolbar.

docs/changes.rst
```diff
@@ -29,6 +29,8 @@ Pending
   <https://astral.sh/blog/the-ruff-formatter>`__.
 * Changed the default position of the toolbar from top to the upper top
   position.
+* Fixed the bug causing ``'djdt' is not a registered namespace`` and updated
+  docs to help in initial configuration while running tests.
```
tim-schilling, line 33:
```suggestion
* Added a link in the installation docs to a more complete installation example in the example app.
* Added check to prevent the toolbar from being installed when tests are running.
* Added test to example app and command to run the example app's tests.  
```

tim-schilling (commented):
This is coming along nicely Velda. I've got a few suggested changes. Hopefully it doesn't break any of the tests.

VeldaKiara:
I have removed the patch decorator, updated the docs and made some changes to the test_panel_empty check for the test to pass

tests/test_checks.py
```diff
@@ -120,7 +123,17 @@ def test_panels_is_empty(self):
...
+                Error(
+                    "The Django Debug Toolbar can't be used with tests",
+                    hint="Django changes the DEBUG setting to False when running "
+                    "tests. By default the Django Debug Toolbar is installed because "
+                    "DEBUG is set to True. For most cases, you need to avoid installing "
+                    "the toolbar when running tests. If you feel this check is in error, "
+                    "you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to "
+                    "bypass this check.",
+                    id="debug_toolbar.W008",
+                ),
```
tim-schilling, line 136: Why is `run_checks` returning this error? I wouldn't have expected this test to require a change.

tests/test_checks.py
```diff
@@ -258,3 +271,68 @@ def test_check_w007_invalid(self, mocked_guess_type):
...
+    #                     "tests. By default the Django Debug Toolbar is installed because "
+    #                     "DEBUG is set to True. For most cases, you need to avoid installing "
+    #                     "the toolbar when running tests. If you feel this check is in error, "
+    #                     "you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to "
+    #                     "bypass this check.",
+    #                     id="debug_toolbar.W008",
+    #                 )
+    #             ],
+
+    # )
```
tim-schilling, line 306: In most cases, we don't want to commit commented out code.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "esbonio.sphinx.confDir": ""
+}
```
tim-schilling, line 3: If we're going to commit changes for vscode, let's do it in a separate PR. If you didn't intend to commit this file, [you can add it to your `.git/info/exclude` file which is the same as .gitignore, but just for your machine](https://stackoverflow.com/a/1753078/1637351).

VeldaKiara: I did not mean to commit this, and I forgot to remove the comments, I will rectify in the next update.

tests/test_checks.py
VeldaKiara: this is what I was getting 

```
FAIL: test_panels_is_empty (tests.test_checks.ChecksTestCase)
  File "django-debug-toolbar/tests/test_checks.py", line 98, in test_panels_is_empty
    self.assertEqual(
AssertionError: Lists differ: [<War[165 chars]005'>, <Error: level=40, msg="The Django Debug[413 chars]08'>] != [<War[165 chars]005'>]

First list contains 1 additional elements.
First extra element 1:
<Error: level=40, msg="The Django Debug Toolbar can't be used with tests", hint="Django changes the DEBUG setting to False when running tests. By default the Django Debug Toolbar is installed because DEBUG is set to True. For most cases, you need to avoid installing the toolbar when running tests. If you feel this check is in error, you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to bypass this check.", obj=None, id='debug_toolbar.W008'>

Diff is 1179 characters long. Set self.maxDiff to None to see it.
```

tim-schilling: Why is the test settings `IS_RUNNING_TESTS` value of `False` not being respected?

VeldaKiara: I am working on this to see what the issue is.

tim-schilling (approved):
Thank you for sticking with this one @VeldaKiara!

@matthiask @elineda @living180 could one of you give this a look over to confirm it makes sense to you? Specifically the naming and documentation.

docs/changes.rst
```diff
@@ -22,6 +22,16 @@ Pending
   ``DEBUG_TOOLBAR_SETTINGS``.
 * Add a note on the profiling panel about using Python 3.12 and later
   about needing ``--nothreading``
+* Added ``RUNNING_TESTS`` setting to allow overriding the
```
matthiask, line 25:
```suggestion
* Added ``IS_RUNNING_TESTS`` setting to allow overriding the
```

debug_toolbar/apps.py
```diff
@@ -208,7 +208,29 @@ def js_mimetype_check(app_configs, **kwargs):
...
+        return [
+            Error(
+                "The Django Debug Toolbar can't be used with tests",
+                hint="Django changes the DEBUG setting to False when running "
+                "tests. By default the Django Debug Toolbar is installed because "
+                "DEBUG is set to True. For most cases, you need to avoid installing "
+                "the toolbar when running tests. If you feel this check is in error, "
+                "you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to "
+                "bypass this check.",
+                id="debug_toolbar.W009",
```
matthiask, line 226: I notice that we're using a different naming convention to Django here. Django uses `W` for warnings and `E` for errors. Just saying, it doesn't have to be changed.

tests/settings.py
```diff
@@ -126,5 +126,6 @@
 
 DEBUG_TOOLBAR_CONFIG = {
     # Django's test client sets wsgi.multiprocess to True inappropriately
-    "RENDER_PANELS": False
+    "RENDER_PANELS": False,
+    "IS_RUNNING_TESTS": False,
```
matthiask, line 130: Maybe add a comment here?

```suggestion
    # IS_RUNNING_TESTS must be False even though we're running tests because we're running the toolbar's own tests.
    "IS_RUNNING_TESTS": False,
```

tests/test_checks.py
```diff
@@ -236,8 +239,43 @@ def test_check_w007_invalid(self, mocked_guess_type):
             ],
         )
 
+    def test_debug_toolbar_installed_when_running_tests(self):
```
matthiask, line 242: I'm not sure why you're modifying the `dt_settings.get_config()` return value instead of using e.g. `with override_settings(...)`? I suspect that there's a good reason for it; if there is, please add a short docstring.

tests/test_checks.py
```diff
@@ -236,8 +239,43 @@ def test_check_w007_invalid(self, mocked_guess_type):
...
+            errors = debug_toolbar_installed_when_running_tests_check(None)
+            self.assertEqual(len(errors), 0)
+        with self.settings(DEBUG=True):
+            dt_settings.get_config()["IS_RUNNING_TESTS"] = False
+            errors = debug_toolbar_installed_when_running_tests_check(None)
+            self.assertEqual(len(errors), 0)
+        with self.settings(DEBUG=False):
+            dt_settings.get_config()["IS_RUNNING_TESTS"] = True
+            errors = debug_toolbar_installed_when_running_tests_check(None)
+            self.assertEqual(
```
matthiask, line 258: If I had written this I would have written it as `self.assertEqual([error.id for error in errors], ["debug_toolbat.W009"])` so that the test doesn't have to change if we reword the error itself for some reason. But I'm fine with your more strict version.

matthiask (approved):
Thanks for working on this! 

I mostly have suggestions and questions for clarifications, so I'm approving already.

tests/test_checks.py
tim-schilling: If we were to use `override_settings`, we'd have to define the entire `DEBUG_TOOLBAR_CONFIG` dictionary. And I don't think there's a merge settings method.

debug_toolbar/apps.py
tim-schilling: That's a good catch. We should follow that standard.

tim-schilling: I updated it to E001

tests/test_checks.py
tim-schilling: Let me know what you think of the new comment.

matthiask: Thanks! No further comments :-)

matthiask:
I think this is ready to merge, or is there anything left to do?

PR #1892, opened 2024-03-11
Make tox pass selenium environment variables

living180:
# Description

When running tox, pass through the user's `DISPLAY` and `DJANGO_SELENIUM_TESTS` environment variables, so that

    DJANGO_SELENIUM_TESTS=true tox

will actually run the Selenuim integration tests.  Without this change, the test suite never sees the `DJANGO_SELENIUM_TESTS` variable and thus skips the integration tests.  Without `DISPLAY`, the integration tests will error out (unless `CI` is present in the environment to instruct the test suite to run the Selenium webdriver in headless mode).

# Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -30,6 +30,8 @@ passenv=
     DB_PASSWORD
     DB_HOST
     DB_PORT
+    DISPLAY
+    DJANGO_SELENIUM_TESTS
     GITHUB_*
 setenv =
     PYTHONPATH = {toxinidir}


tim-schilling:
I think we previously left the selenium tests off because they're a bit burdensome to run locally.

matthiask (approved):
Makes sense. This should make it easier to run the selenium tests locally.

living180:
Yes, this PR isn't changing that default, it is just making it possible to enable them through tox if desired.  Right now the [Contributing](https://django-debug-toolbar.readthedocs.io/en/latest/contributing.html) docs includes the following snippet:

image

The first two lines using `make` work, but the third using `tox` doesn't without this change, since tox won't pass through the `DJANGO_SELENIUM_TESTS` variable without it.

Issue #1905, opened 2024-04-12
Exception while resolving variable 'toolbar' in template


foundkey:
When sending a request using ajax, toolbar.js will send a request at the same time:
`GET /__debug__/history_sidebar/?store_id=<id> HTTP/1.1`

This request will throw the following exception when rendering the template：
> Traceback (most recent call last):
  File "/Users/.venv/lib/python3.10/site-packages/django/template/base.py", line 898, in _resolve_lookup
    raise VariableDoesNotExist(
django.template.base.VariableDoesNotExist: Failed lookup for key [toolbar] in [{'True': True, 'False': False, 'None': None}, {'panel': <debug_toolbar.panels.cache.CachePanel object at 0x11f936590>}]
[DEBUG] 12/Apr/2024 10:31:22 - Exception while resolving variable 'toolbar' in template 'debug_toolbar/includes/panel_content.html'.
Traceback (most recent call last):
  File "/Users/.venv/lib/python3.10/site-packages/django/template/base.py", line 875, in _resolve_lookup
    current = current[bit]
  File "/Users/.venv/lib/python3.10/site-packages/django/template/context.py", line 83, in __getitem__
    raise KeyError(key)
KeyError: 'toolbar'

This request is handled by this method: [history_sidebar(request)](https://github.com/jazzband/django-debug-toolbar/blob/main/debug_toolbar/panels/history/views.py#L11)
It seems because `toolbar` is not passed in `panel_context`.

foundkey:
After turning on Debug level logging, VariableDoesNotExist errors appear frequently during use. But it does not affect the use.

tim-schilling:
Hmm. I'm softening on my earlier stance on the PR.

Riskcomplexx:
I'm currently seeing the same thing but oddly only in firefox and not chrome.

JasonSanDiego:
I see the issue on Chrome also

tim-schilling:
@Riskcomplexx @JasonSanDiego I appreciate your seconding of the issue, but what will help get this resolve a reproducible example. Would you be able to provide one?

PR #1906, opened 2024-04-12
fix Exception while resolving variable 'toolbar' in template

foundkey:
# Description

When sending a request using ajax, toolbar.js will send a request at the same time:
`GET /__debug__/history_sidebar/?store_id=<id> HTTP/1.1`

This request will throw the following exception when rendering the template.
It seems because toolbar is not passed in panel_context.

Fixes # (#1905 )

1. add `toolbar` into `panel_context`.

# Checklist:

### DIFF
--- a/debug_toolbar/panels/history/views.py
+++ b/debug_toolbar/panels/history/views.py
@@ -24,7 +24,7 @@ def history_sidebar(request):
         for panel in toolbar.panels:
             if exclude_history and not panel.is_historical:
                 continue
-            panel_context = {"panel": panel}
+            panel_context = {"panel": panel, "toolbar": toolbar}
             context[panel.panel_id] = {
                 "button": render_to_string(
                     "debug_toolbar/includes/panel_button.html", panel_context


matthiask:
I think this makes sense, but I'm not 100% sure if you're just fixing an issue with a missing variable or if the functionality of loading panel contents with `RENDER_PANELS = True` and older requests was broken previously.

If it's the former I'm good with the PR, but if it's the later we'd really want a test to make sure we're not regressing here.

foundkey:
> I think this makes sense, but I'm not 100% sure if you're just fixing an issue with a missing variable or if the functionality of loading panel contents with `RENDER_PANELS = True` and older requests was broken previously.
> 
> If it's the former I'm good with the PR, but if it's the later we'd really want a test to make sure we're not regressing here.

My local test was fine. After modification, no exception logs can be seen in the same scenario.
However, I only tested the scenarios I encountered, and I did not conduct a comprehensive test.

tim-schilling (changes requested):
Thanks for identifying a solution! Would you be able to include a test that confirms that this is fixed? It's hard to accept a change without being able to see what it's fixing.

foundkey:
> Thanks for identifying a solution! Would you be able to include a test that confirms that this is fixed? It's hard to accept a change without being able to see what it's fixing.

Sorry, this test doesn't look easy to add.

According to the Django Template rendering process, the exception `django.template.base.VariableDoesNotExist` is handled gracefully. This exception is not thrown and is replaced with a default value, which is an empty string if not configured.
template_resolve

This results in an empty string being obtained at the exception and going to the false branch, which is in line with the expected logic.
debug-toolbar-exception

According to `debug-toolbar`'s code:
if `panel_content.html` is rendered under a non-ajax request, there will be a toolbar variable in the context, which is normal. 
If `panel_content.html` is rendered under an ajax request, the toolbar variable is missing from the context, eventually leading to the false branch. In ajax request, `toolbar.should_render_panels` value is always false.

It seems that this exception does not trigger any problems. Only in debug level logs, you will see relevant error logs.
To sum up, it won't hurt if this anomaly is not fixed.

foundkey:
After turning on Debug level logging, VariableDoesNotExist errors appear frequently during use. But it does not affect the use. I think my PR is redundant. It can be closed.

Riskcomplexx:
Just curious why you closed this? I don't think your PR was redundant at all. Error should not be thrown unless there is an error. 

I just spent half my day trying to figure out this issue, luckily I stumbled on your PR.

tim-schilling:
@Riskcomplexx would you be able to open a PR with the fix and a test to prevent a regression?

foundkey:
> Just curious why you closed this? I don't think your PR was redundant at all. Error should not be thrown unless there is an error.
> 
> I just spent half my day trying to figure out this issue, luckily I stumbled on your PR.

@Riskcomplexx 

During the subsequent debugging process, I found that this exception did not affect the use. And there is already code to handle this exception. You can refer to the analysis in my comment above.

In addition, the test for this PR is not easy to add. I don’t have much time to complete this test, so I can only close this PR.

PR #1911, opened 2024-05-01
Draft: Add warning async is not supported

### LINKED ISSUE #1845: Document that the toolbar does not support async functionality
We should include in our readme that the application doesn't not support async functionality and that effort is under development. If possible, it'd be great to detect that async functionality is being used and warn the user that the toolbar won't be effective. @michaelurban suggested this in https://github.com/jazzband/django-debug-toolbar/issues/1430#issuecomment-1697910837

denisSurkov:
# Description

Add warning async is not supported.

Fixes #1845 

# Checklist:

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -13,7 +13,7 @@ from django.core.cache import cache
 from django.db import connection
 from django.http import HttpResponse
 from django.template.loader import get_template
-from django.test import RequestFactory
+from django.test import RequestFactory, AsyncRequestFactory
 from django.test.utils import override_settings
 
 from debug_toolbar.forms import SignedDataForm
@@ -33,12 +33,6 @@ try:
     from selenium.webdriver.support.wait import WebDriverWait
 except ImportError:
     webdriver = None
-
-
-try:
-    from django.test import AsyncRequestFactory
-except ImportError:
-    AsyncRequestFactory = None
 
 
 rf = RequestFactory()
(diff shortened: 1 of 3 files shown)


tests/test_integration.py
```diff
@@ -792,3 +794,46 @@ def test_ajax_refresh(self):
         )
         self.assertNotIn("/ajax/", history_panel.text)
         self.assertIn("/json_view/", history_panel.text)
+
+
```
cclauss, line 798:
```suggestion

from django.test import AsyncRequestFactory

```

denisSurkov: Thank you for the tip

cclauss: You must move that line up near the top of the file with the other imports.

denisSurkov:
As far as I see the problem would become irrelevant soon, after #1933 

If so, feel free to close this PR

tests/test_integration.py
```diff
@@ -33,6 +35,12 @@
     webdriver = None
 
 
+try:
+    from django.test import AsyncRequestFactory
+except ImportError:
+    AsyncRequestFactory = None
```
cclauss, line 41: `AsyncRequestFactory` is defined in Django >= v3.1 so the `ImportError` should not happen on all supported versions of Django.

Just modify line 16 to read `from django.test import AsyncRequestFactory, RequestFactory`

tim-schilling:
Sorry @denisSurkov as you mentioned #1938 should make this work unnecessary. It should be merged in the next few weeks so I think we can put this to rest for the time being. I appreciate your help and desire to make the toolbar a better library!

If you're looking to contribute, there are several things that could be improved. If you're async focused, #1938 has identified several panels that are async incompatible (`Panel.is_async = False`) that need further investigation.

PR #1913, opened 2024-05-12
Dark mode support

TheRealVizard:
# Description

This PR aims to add native support for dark mode.
It adds a button to switch between themes.
image
image

# Checklist:


debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -213,6 +213,33 @@ const djdt = {
...
+                    break;
+                case "light":
+                    djDebug.setAttribute("data-theme", "dark");
+                    localStorage.setItem("djdt.user-theme", "dark");
+                    break;
+                case "dark":
+                    djDebug.setAttribute("data-theme", "auto");
+                    localStorage.setItem("djdt.user-theme", "auto");
+                    break;
+                default:
```
matthiask, line 237:
```suggestion
                default: /* dark is the default */
```

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -654,3 +720,17 @@
...
+    align-items: center;
+    cursor: pointer;
+}
+#djToggleThemeButton > svg {
+    margin-left: auto;
+}
+#djDebug[data-theme="light"] #djToggleThemeButton svg.theme-light,
+#djDebug[data-theme="dark"] #djToggleThemeButton svg.theme-dark,
+#djDebug[data-theme="auto"] #djToggleThemeButton svg.theme-auto {
+    display: block !important;
```
matthiask, line 735: Can you explain why the `!Important` is necessary here and above? We have tried avoiding `!important` in our CSS until now.

matthiask (commented):
Thanks for your contribution!

I think I like it. I wasn't convinced it's worth it, but the diff looks really clean and the changes are well localized. Thanks.

docs/changes.rst
```diff
@@ -32,6 +32,8 @@ Pending
 * Added check to prevent the toolbar from being installed when tests
   are running.
 * Added test to example app and command to run the example app's tests.
+* Implemented dark mode theme and button to toggle the theme,
+  introduced config var ``DEFAULT_THEME`` to set default theme to use.
```
matthiask, line 36:
```suggestion
  introduced the ``DEFAULT_THEME`` setting which sets the default theme to use.
```

The spell checker doesn't like `config var`.

debug_toolbar/static/debug_toolbar/css/toolbar.css
TheRealVizard: Nice catch, I missed these on the final cleanup, in [e8c6820](https://github.com/jazzband/django-debug-toolbar/pull/1913/commits/e8c68200b107f0a564bdaaf29ade5028f900a84a) they are removed

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -122,9 +190,7 @@
 #djDebug button:active {
     border: 1px solid #aaa;
     border-bottom: 1px solid #888;
-    box-shadow:
-        inset 0 0 5px 2px #aaa,
-        0 1px 0 0 #eee;
+    box-shadow: inset 0 0 5px 2px #aaa, 0 1px 0 0 #eee;
```
matthiask, line 193: This doesn't look as if it has been formatted by prettier v4 (using `pre-commit`)?

matthiask (commented):
Thank you! One more thing. The Django admin panel only uses `prefers-color-scheme` in the JavaScript code to determine the initial value. This simplifies the CSS insofar as the variables for the dark mode do not have to be repeated. Do you think something like that would be a good idea here as well? See https://github.com/django/django/blob/a09082a9bec18f8e3ee8c10d473013ec67ffe93b/django/contrib/admin/static/admin/js/theme.js#L14

debug_toolbar/static/debug_toolbar/css/toolbar.css
TheRealVizard: probably was my VS formatter, odd since it's set to use prettier (maybe was not using the same rules)
image
My last commit reverted that change

TheRealVizard:
> > Thank you! One more thing. The Django admin panel only uses `prefers-color-scheme` in the JavaScript code to determine the initial value. This simplifies the CSS insofar as the variables for the dark mode do not have to be repeated. Do you think something like that would be a good idea here as well? See
> 
> Yes! This can be used to avoid duplication.
> 
> They still have it on their CSS, maybe in case JS is disabled by default... https://github.com/django/django/blob/ceaf1e2848583ba832cc74715da38c802b6b0671/django/contrib/admin/static/admin/css/dark_mode.css#L1C1-L38C26

Ill push a change to this once i get home

tim-schilling (approved):
This looks really nice! My only gripe is that we're taking up an entire row of the toolbar for the toggle button. I'm not sure there's a better option though. I'm bringing it up to see if anyone else has an opinion.

matthiask:
For the record, I agree with this, but I think that this can be addressed later, in a different pull request. Thanks!

PR #1921, opened 2024-05-27
Add a section to the installation docs about running tests

### LINKED ISSUE #1920: Error after update to 4.4.1 (debug_toolbar.E001) The Django Debug Toolbar can't be used with tests
After upgrading to 4.4.1 we got an error when trying to run our tests, despite `DEBUG` being `False`:
```
(debug_toolbar.E001) The Django Debug Toolbar can't be used with tests
	HINT: Django changes the DEBUG setting to False when running tests. By default the Django Debug Toolbar is installed because DEBUG is set to True. For most cases, you need to avoid installing the toolbar when running tests. If you feel this check is in error, you can set `DEBUG_TOOLBAR_CONFIG['IS_RUNNING_TESTS'] = False` to bypass this check.
```
It took us a while to figure out what went wrong. It turns out the default installation method, recommended in the docs, doesn't work with the the changes from 4.4.0 (even if `DEBUG` is `False`, which should deactivate Debug Toolbar) 

After some back and forth we figured out that the example directory configuration was the only way to run our tests:
```
ENABLE_DEBUG_TOOLBAR = DEBUG and "test" not in sys.argv
if ENABLE_DEBUG_TOOLBAR:
    INSTALLED_APPS += [
        "debug_toolbar",
    ]
    MIDDLEWARE += [
        "debug_toolbar.middleware.DebugToolbarMiddleware",
    ]
```
The default installation docs are haven't changed for the 4.4.0 release and the example app is different than the recommended installation in the docs. Is this a missed docs update or an unintended breaking change?

matthiask:
Fixes #1920.

I thought about including the relevant documentation in the earlier steps, but I'd have to explain DEBUG, INTERNAL_IPS and TESTING all at once instead of introducing everything step by step. So even though it may be annoying to go back and modify code the user just added it still reads better to me, especially since it only applies to users running tests in their project. (I would hope a lot of them do, but still.)

# Checklist:

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -156,6 +156,30 @@ option.
     able to get the toolbar to work with your docker installation, review
     the code in ``debug_toolbar.middleware.show_toolbar``.
 
+7. Disable the toolbar when running tests (optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you're running tests in your project you shouldn't activate the toolbar. You can do this by adding another setting:
+
+.. code-block:: python
+
+    TESTING = "argv" in sys.argv
+
+    if not TESTING:
+        INSTALLED_APPS = [*INSTALLED_APPS, "debug_toolbar"]
+        MIDDLEWARE = ["debug_toolbar.middleware.DebugToolbarMiddleware", *MIDDLEWARE]
+
+You should also modify your URLconf file:
+
+.. code-block:: python
+
+    if not settings.TESTING:
+        urlpatterns = [
+            *urlpatterns,
+            path("__debug__/", include("debug_toolbar.urls")),
+        ]
+
+
 Troubleshooting
 ---------------
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -164,7 +164,7 @@ can do this by adding another setting:
 
 .. code-block:: python
 
-    TESTING = "argv" in sys.argv
+    TESTING = "test" in sys.argv
 
     if not TESTING:
         INSTALLED_APPS = [
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -159,15 +159,22 @@ option.
 7. Disable the toolbar when running tests (optional)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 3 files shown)


docs/installation.rst
```diff
@@ -156,6 +156,39 @@ option.
...
 
+7. Disable the toolbar when running tests (optional)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If you're running tests in your project you shouldn't activate the toolbar. You
+can do this by adding another setting:
+
+.. code-block:: python
+
+    TESTING = "argv" in sys.argv
```
gegoune, line 167: Shouldn't that be 
```suggestion
    TESTING = "test" in sys.argv
```
?

matthiask: Yes, thanks!

tim-schilling:
I don't think this was necessary. The docs point the user to the example app in the repo.

tim-schilling:
> Check out the configuration example in the [example app](https://github.com/jazzband/django-debug-toolbar/tree/main/example) to learn how to set up the toolbar to function smoothly while running your tests.

matthiask:
I'm not so sure. The example contains many aspects and the test suite thing is just one of them. The question arrived really quickly and I thought I'd lean towards more docs. Maybe the section could be shortened somewhat or people could be made aware of the problem in ways which are shorter than the current version but maybe still more visible than what we had before?

tim-schilling:
I feel like there's a better installation process with `AppConfig.ready` and the middleware where we could check `DEBUG` after the Django test runner does its magic.

matthiask:
Definitely sounds interesting. One of my first big(ger) decisions in this project was to remove the magical installation, so I'm a bit wary, but `AppConfig.ready` sounds like a good way to opt into some behaviors.

tim-schilling:
Ah we may have a difference of opinion then. I feel like asking a person to add any if statements to their apps configuration is a shortcoming of the project.

matthiask:
Yes, maybe :-)

I lean towards the "worse is better" school of software design where it's slightly better to be simple than to be correct. I think that everything breaks down at a point, and if the implementation is simple it will be easier to understand and fix (or work around) than if the implementation is complex.

That being said, I like Django a lot and its implementation certainly isn't simple. So it really depends. As long as we can put behaviors into parts which can be easily replaced and which have a small API I'm fine with it. The `show_toolbar(request) -> bool` callback is a great example of this.

PR #1923, opened 2024-05-27
Avoid setting color-scheme on :root, we're only a guest on pages

### LINKED ISSUE #1922: Version 4.4.1 changes the look of DRF browsable API
After upgrading to version 4.4.1 the browsable API suddenly becomes partly dark:
image

Whereas on version 4.3 it looks like this:
image

matthiask:
Fixes #1922.

# Checklist:

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -11,7 +11,6 @@
         monospace, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol",
         "Noto Color Emoji";
 
-    color-scheme: light;
     --djdt-font-color: black;
     --djdt-background-color: white;
     --djdt-panel-content-background-color: #eee;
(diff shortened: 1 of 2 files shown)


debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -58,7 +56,6 @@
 }
 
 #djDebug[data-theme="dark"] {
-    color-scheme: dark;
```
tim-schilling, line 61: Couldn't this one have stayed?

matthiask: I thought about it and I'm not sure what it even does since we're basically specifying all colors and backgrounds. Yes, it could stay, but then we'd also have to set the opposite on `#djDebug[data-theme="light"]` (I think) 

The bigger point here is that while I do have much experience with CSS generally I don't really know how the `color-scheme` property behaves, and it felt safer to just remove it at this time.

tim-schilling: Alright that's fair.

matthiask: Btw, sorry for pressing ahead, I wanted to release a better version quickly instead of waiting for more reports.

PR #1925, opened 2024-05-28
Limit the cases for E001 to likely scenarios

### LINKED ISSUE #1920: Error after update to 4.4.1 (debug_toolbar.E001) The Django Debug Toolbar can't be used with tests

tim-schilling:
# Description 

Only when the user is customizing both the show toolbar callback setting and the URLs aren't installed will the underlying NoReverseMatch error occur.

Fixes #1920

# Checklist:

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -152,6 +152,16 @@ Toolbar options
      implication is that it is possible to execute arbitrary SQL through the
      SQL panel when the ``SECRET_KEY`` value is leaked somehow.
 
+  .. warning::
+
+     Do not use
+     ``DEBUG_TOOLBAR_CONFIG = {"SHOW_TOOLBAR_CALLBACK": lambda request: DEBUG}``
+     in your project's settings.py file. The toolbar expects to use
+     ``django.conf.settings.DEBUG``. Using your project's setting's ``DEBUG``
+     is likely to cause unexpected when running your tests. This is because
+     Django automatically sets ``settings.DEBUG = False``, but your project's
+     setting's ``DEBUG`` will still be set to ``True``.
+
 .. _OBSERVE_REQUEST_CALLBACK:
 
 * ``OBSERVE_REQUEST_CALLBACK``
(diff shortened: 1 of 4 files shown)


tim-schilling:
@matthiask I'm pretty confident the advanced installation is only necessary if the show toolbar callback isn't making use of `settings.DEBUG`.

matthiask (approved):
Interesting. The explanation of `DEBUG` vs `settings.DEBUG` makes a lot of sense to me. Code looks good!

docs/configuration.rst
```diff
@@ -152,6 +152,16 @@ Toolbar options
      implication is that it is possible to execute arbitrary SQL through the
      SQL panel when the ``SECRET_KEY`` value is leaked somehow.
 
+  .. warning::
+
+     Do not use
+     ``DEBUG_TOOLBAR_CONFIG = {"SHOW_TOOLBAR_CALLBACK": lambda request: DEBUG}``
+     in your project's settings.py file. The toolbar expects to use
+     ``django.conf.settings.DEBUG``. Using your project's setting's ``DEBUG``
+     is likely to cause unexpected when running your tests. This is because
```
tim-schilling, line 161:
```suggestion
     is likely to cause unexpected results when running your tests. This is because
```

PR #1926, opened 2024-05-28
Introduce debug_toolbar_urls to simplify installation

tim-schilling:
# Description

While this isn't a huge improvement, it provides a hook for us
to more easily introduce library level controls for when urls
are installed. It also eliminates the user from having to
write an if statement for the most basic of installations.

# Checklist:

This is a follow-up to #1925


matthiask (approved):
Seems to work well; I'm not sure if I have one of the odd projects where running tests with the debug toolbar was problematic or not (I don't remember), but everything works fine now.

PR #1930, opened 2024-06-01
Fix overriding font-family for both light and dark themes

federicobond:
The [docs](https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#theming-support) mention that the toolbar can be customized by overriding the --djdt-* variables defined in :root, like this:
```css
:root {
    --djdt-font-family-primary: 'Roboto', sans-serif;
}
```

But with the implementation of dark theme support these same variables [are also defined](https://github.com/jazzband/django-debug-toolbar/blob/d4811824f971eb79a72091b26119439915647949/debug_toolbar/static/debug_toolbar/css/toolbar.css#L2-L3) in #djDebug[data-theme="light"] with a higher specificity, so in reality you will need to match the specificity of that rule to override the variables.

This moves the font-family declarations to root so changes to the font-family on :root apply to both light and dark themes.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -1,6 +1,5 @@
 /* Variable definitions */
-:root,
-#djDebug[data-theme="light"] {
+:root {
     /* Font families are the same as in Django admin/css/base.css */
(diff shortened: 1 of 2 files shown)


matthiask:
Makes sense, thanks!

PR #1933, opened 2024-06-07
Fixed #1682 -- alert user when using file field without proper encoding

### LINKED ISSUE #1682: Log when form is using file fields without proper enctype attribute on the form.
One of the issues that can trip up new developers is the fact that forms require `enctype="multipart/form-data"` when there's a file being uploaded. If a developer doesn't know this, it can be hard to track down. I think it'd be reasonable to have the toolbar inspect the loaded HTML content and inspect it for invalid configurations. The first of which being a `<input type="file" ..> />` in a form that isn't defined as `<form enctype="multipart/form-data">`.

bkdekoning:
# Description

Added functionality where the toolbar inspects the loaded HTML content for a form that includes a file input but does not have the encoding type set to `multipart/form-data`, and warns the user if so. This is implemented in a separate alerts panel, as discussed. The alerts panel is placed at the top of the toolbar, to ensure that it receives appropriate attention when issues arise. My one concern with this implementation is that panel may cause people to expect that the toolbar checks for a wide range of issues, whereas it currently just checks for a single one.

What I did specifically is subclass Python's HTML parser and added a function called `check_invalid_file_form_configuration` in `panels/templates/panel.py`. The function checks all forms for a file input type and, if it exists, checks if the form has encoding type multipart/form-data. If it does not, the error message is shown in the alerts panel by passing it to `record_stats`. It also amends the `nav_subtitle` property of the panel to notify the user that there is an issue. Added four tests and checked that the test suite passes for all compatible versions of Django and Python through Tox.

Fixes #1682

# Checklist:

### DIFF
--- a/debug_toolbar/panels/alerts.py
+++ b/debug_toolbar/panels/alerts.py
@@ -61,6 +61,18 @@ class AlertsPanel(Panel):
     A panel to alert users to issues.
     """
 
+    messages = {
+        "form_id_missing_enctype": _(
+            'Form with id "{form_id}" contains file input, but does not have the attribute enctype="multipart/form-data".'
+        ),
+        "form_missing_enctype": _(
+            'Form contains file input, but does not have the attribute enctype="multipart/form-data".'
+        ),
+        "input_refs_form_missing_enctype": _(
+            'Input element references form with id "{form_id}", but the form does not have the attribute enctype="multipart/form-data".'
+        ),
+    }
+
     title = _("Alerts")
 
     template = "debug_toolbar/panels/alerts.html"
@@ -101,11 +113,12 @@ class AlertsPanel(Panel):
                     for elem in form["submit_element_attrs"]
                 )
             ):
-                form_id = form["form_attrs"].get("id", "no form id")
-                alert = (
-                    f'Form with id "{form_id}" contains file input but '
-                    "does not have multipart/form-data encoding."
-                )
+                if form_id := form["form_attrs"].get("id"):
+                    alert = self.messages["form_id_missing_enctype"].format(
+                        form_id=form_id
+                    )
+                else:
+                    alert = self.messages["form_missing_enctype"]
                 self.add_alert({"alert": alert})
 
         # Check for file inputs that reference a form
@@ -118,9 +131,8 @@ class AlertsPanel(Panel):
             if form_id and attrs.get("type") == "file":
                 form_attrs = form_attrs_by_id.get(form_id)
                 if form_attrs and form_attrs.get("enctype") != "multipart/form-data":
-                    alert = (
-                        f'Input element references form with id "{form_id}" '
-                        "but the form does not have multipart/form-data encoding."
+                    alert = self.messages["input_refs_form_missing_enctype"].format(
+                        form_id=form_id
                     )
                     self.add_alert({"alert": alert})
 
--- a/debug_toolbar/panels/alerts.py
+++ b/debug_toolbar/panels/alerts.py
@@ -1,0 +1,95 @@
+from html.parser import HTMLParser
+
+from django.utils.translation import gettext_lazy as _
+
+from debug_toolbar.panels import Panel
+
+
+class FormParser(HTMLParser):
+    """
+    HTML form parser, used to check for invalid configurations of forms that
+    take file inputs.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.in_form = False
+        self.current_form = {}
+        self.forms = []
+
+    def handle_starttag(self, tag, attrs):
+        attrs = dict(attrs)
+        if tag == "form":
+            self.in_form = True
+            self.current_form = {
+                "file_form": False,
+                "form_attrs": attrs,
+                "submit_element_attrs": [],
+            }
+        elif self.in_form and tag == "input" and attrs.get("type") == "file":
+            self.current_form["file_form"] = True
+        elif self.in_form and (
+            (tag == "input" and attrs.get("type") in {"submit", "image"})
+            or tag == "button"
+        ):
+            self.current_form["submit_element_attrs"].append(attrs)
+
+    def handle_endtag(self, tag):
+        if tag == "form" and self.in_form:
+            self.forms.append(self.current_form)
+            self.in_form = False
+
+
+class AlertsPanel(Panel):
+    """
+    A panel to alert users to issues.
+    """
+
+    title = _("Alerts")
+
+    template = "debug_toolbar/panels/alerts.html"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.issues = []
+
+    @property
+    def nav_subtitle(self):
+        if self.issues:
+            issue_text = "issue" if len(self.issues) == 1 else "issues"
+            return f"{len(self.issues)} {issue_text} found"
+        else:
+            return ""
+
+    def add_issue(self, issue):
+        self.issues.append(issue)
+
+    def check_invalid_file_form_configuration(self, html_content):
+        parser = FormParser()
+        parser.feed(html_content)
+
+        for form in parser.forms:
+            if (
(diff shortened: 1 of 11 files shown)


tim-schilling:
By the way, you don't have to recreate the PRs. I think for 95% of cases, we're happy with re-using the same PR for a change.

debug_toolbar/panels/alerts.py
```diff
@@ -0,0 +1,95 @@
...
+        parser.feed(html_content)
+
+        for form in parser.forms:
+            if (
+                form["file_form"]
+                and form["form_attrs"].get("enctype") != "multipart/form-data"
+                and not any(
+                    elem.get("formenctype") == "multipart/form-data"
+                    for elem in form["submit_element_attrs"]
+                )
```
tim-schilling, line 78: We should also check that the [`form` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#form) is not set or set to an empty string. We could potentially handle this, but it's pretty un-django-ific in my opinion. Basically if someone is using that attribute, then let's disable this check for that input.

debug_toolbar/panels/alerts.py
```diff
@@ -0,0 +1,95 @@
...
+    A panel to alert users to issues.
+    """
+
+    title = _("Alerts")
+
+    template = "debug_toolbar/panels/alerts.html"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.issues = []
```
tim-schilling, line 54: I wonder if it'd be better to set this in [`enable_instrumentation`](https://django-debug-toolbar.readthedocs.io/en/latest/panels.html#debug_toolbar.panels.Panel.enable_instrumentation) instead

debug_toolbar/panels/alerts.py
```diff
@@ -0,0 +1,95 @@
...
+    @property
+    def nav_subtitle(self):
+        if self.issues:
+            issue_text = "issue" if len(self.issues) == 1 else "issues"
+            return f"{len(self.issues)} {issue_text} found"
+        else:
+            return ""
+
+    def add_issue(self, issue):
+        self.issues.append(issue)
```
tim-schilling, line 65: Since the panel is called `AlertsPanel`, let's use `alerts` rather than `issues`.

debug_toolbar/panels/alerts.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def check_invalid_file_form_configuration(self, html_content):
```
tim-schilling, line 67: Let's include your great description from your commit in the doc string here:

> Inspects HTML content for a form that includes a file input but does not have the encoding type set to multipart/form-data, and warns the user if so.

docs/panels.rst
```diff
@@ -9,6 +9,15 @@ Default built-in panels
...
 The following panels are enabled by default.
 
+Alerts
+~~~~~~~
+
+.. class:: debug_toolbar.panels.alerts.AlertsPanel
+
+This panel shows alerts for a set of pre-defined issues. Currently, the only
+issue it checks for is the encoding of a form that takes a file input not
+being set to ``multipart/form-data``.
```
tim-schilling, line 19:
```suggestion
This panel shows alerts for a set of pre-defined cases:

- Alerts when the response has a form without the
  ``enctype="multipart/form-data"`` attribute and the form contains
  a file input
```

I think it's prudent to start listing them off from the start.

debug_toolbar/panels/alerts.py
```diff
@@ -0,0 +1,95 @@
...
+
+    template = "debug_toolbar/panels/alerts.html"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.issues = []
+
+    @property
+    def nav_subtitle(self):
+        if self.issues:
```
tim-schilling, line 58: Please use `get_stats` to find the alerts rather than relying on the instance's member. It'll remove a step for us with the async project.

tim-schilling (changes requested):
I've got a few opinionated takes on the approach. Let me know if I'm overreaching with any of them. I really like what you have here. I think changing the `nav_subtitle` is a great first version for notifying the user of a problem! Nice solution there.

Well done with the tests, documentation and arrangement of the code too!

debug_toolbar/panels/alerts.py
bkdekoning: I trust your call on this, but have to admit that I do not fully understand the `enable_instrumentation` method. What would the advantage be of setting it there? Looking at the other panels, I commonly see lists being initialized in `__init__`.

debug_toolbar/panels/alerts.py
bkdekoning: I'm happy to try and handle this. We can check for the existence of `tag == "input" and attrs.get("form")` and then check the encoding of the referenced forms by their IDs. Would that be an appropriate solution in your view?

bkdekoning:
While working on the above, I realized that I am implicitly assuming that the form is not nested. While forms [should not be nested](https://html.spec.whatwg.org/multipage/forms.html#the-form-element), I tested such a construction with Chrome, Firefox, and Edge, and they handle it by simply ignoring the "child" form. 

As such, we might end up in a situation where we are not alerting the developer to an issue that may exist for some users -- when the encoding is properly set in the "child" form but not the "parent" form. Is this something we want to issue an additional alert for or would that be overly cautious?

debug_toolbar/panels/alerts.py
bkdekoning: I have implemented this.

tim-schilling:
As a heads up, I probably won't get to this towards the end of June. Things are ramping up elsewhere and the toolbar is taking a back seat. Thank you for making the changes so far!

bkdekoning:
Alright, thanks for the heads-up. For future reference, I think the remaining items to settle on are: 
- Whether to use `enable_instrumentation` or `__init__` to initialize alerts
- Whether we should handle direct form references in file inputs
- Whether we should handle nested forms

Good luck with your other activities!

tim-schilling:
> Whether we should handle nested forms

Nested forms are against the HTML spec, so I think what we have is fine. Most people aren't relying on this to be perfect, so let's get something "good" out.

> Whether we should handle direct form references in file inputs

While I'd like to see us do that, I don't think we have to worry too much about it. It would take a fair amount of work to get a Django form renderer to handle that. Let's lean into what's the common flow for Django devs.

tim-schilling:
> Whether to use enable_instrumentation or __init__ to initialize alerts

Looking at this again, it's not a big deal. Let's go with what you have.

tim-schilling (approved):
Code looks great. I'm going to pull it down and check the UI to see what it looks like. If that's good, I'll merge it.

bkdekoning:
> > Whether we should handle direct form references in file inputs
> 
> While I'd like to see us do that, I don't think we have to worry too much about it. It would take a fair amount of work to get a Django form renderer to handle that. Let's lean into what's the common flow for Django devs.

Just to be sure: I did implement this. When the parser sees a file input field that directly references a form, it adds it to the `referenced_file_inputs` list. The `check_invalid_file_form_configuration` function then fetches the referenced form's encoding type and checks if it is set to `multipart/form-data`.

tim-schilling:
Oh, perfect!

Take a look at the latest commits and let me know if you have any issues with them. The biggest changes were moving the panel and rewording the messages a bit.

bkdekoning:
All looks good to me!

tim-schilling:
@matthiask I'd like to get your approval on the new panel before we merge this.

matthiask:
Yes sure, let's do it!

bkdekoning:
Thank you, @tim-schilling and @matthiask! Very excited to have made my first contribution. I really appreciated the thorough review and guidance.

matthiask:
I haven't dont much here, the thank you belongs to you and Tim!

PR #1938, opened 2024-06-24
Async compatible debug-toolbar middleware

salty-ivy:
# Description
GSoC 2024

This PR aims to add async compatibility in django-debug-toolbar middleware

Fixes # (issue)

# Checklist:

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -6,6 +6,7 @@ import re
 import socket
 from functools import lru_cache
 
+from asgiref.sync import iscoroutinefunction, markcoroutinefunction
 from django.conf import settings
 from django.utils.module_loading import import_string
 
@@ -62,17 +63,29 @@ class DebugToolbarMiddleware:
     on outgoing response.
     """
 
+    sync_capable = True
+    async_capable = True
+
     def __init__(self, get_response):
         self.get_response = get_response
+        # If get_response is a coroutine function, turns us into async mode so
+        # a thread is not consumed during a whole request.
+        self.async_mode = iscoroutinefunction(self.get_response)
+
+        if self.async_mode:
+            # Mark the class as async-capable, but do the actual switch inside
+            # __call__ to avoid swapping out dunder methods.
+            markcoroutinefunction(self)
 
     def __call__(self, request):
+        # Decide whether the toolbar is active for this request.
+        if self.async_mode:
+            return self.__acall__(request)
         # Decide whether the toolbar is active for this request.
         show_toolbar = get_show_toolbar()
         if not show_toolbar(request) or DebugToolbar.is_toolbar_request(request):
             return self.get_response(request)
-
         toolbar = DebugToolbar(request, self.get_response)
-
         # Activate instrumentation ie. monkey-patch.
         for panel in toolbar.enabled_panels:
             panel.enable_instrumentation()
@@ -86,6 +99,36 @@ class DebugToolbarMiddleware:
             for panel in reversed(toolbar.enabled_panels):
                 panel.disable_instrumentation()
 
+        return self._postprocess(request, response, toolbar)
+
+    async def __acall__(self, request):
+        # Decide whether the toolbar is active for this request.
+        show_toolbar = get_show_toolbar()
+        if not show_toolbar(request) or DebugToolbar.is_toolbar_request(request):
+            response = await self.get_response(request)
+            return response
+
+        toolbar = DebugToolbar(request, self.get_response)
+
+        # Activate instrumentation ie. monkey-patch.
+        for panel in toolbar.enabled_panels:
+            panel.enable_instrumentation()
+        try:
+            # Run panels like Django middleware.
+            response = await toolbar.process_request(request)
+        finally:
+            clear_stack_trace_caches()
+            # Deactivate instrumentation ie. monkey-unpatch. This must run
+            # regardless of the response. Keep 'return' clauses below.
+            for panel in reversed(toolbar.enabled_panels):
+                panel.disable_instrumentation()
+
+        return self._postprocess(request, response, toolbar)
+
+    def _postprocess(self, request, response, toolbar):
+        """
+        Post-process the response.
+        """
         # Generate the stats for all requests when the toolbar is being shown,
         # but not necessarily inserted.
         for panel in reversed(toolbar.enabled_panels):
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -1,3 +1,4 @@
+from django.core.handlers.asgi import ASGIRequest
 from django.template.loader import render_to_string
 
 from debug_toolbar import settings as dt_settings
@@ -8,6 +9,8 @@ class Panel:
     """
     Base class for panels.
     """
+
+    is_async = True
 
     def __init__(self, toolbar, get_response):
         self.toolbar = toolbar
@@ -21,6 +24,10 @@ class Panel:
 
     @property
     def enabled(self) -> bool:
+        # check if the panel is async compatible
+        if not self.is_async and isinstance(self.toolbar.request, ASGIRequest):
+            return False
+
         # The user's cookies should override the default value
         cookie_value = self.toolbar.request.COOKIES.get("djdt" + self.panel_id)
         if cookie_value is not None:
--- a/tests/panels/test_async_panel_compatibility.py
+++ b/tests/panels/test_async_panel_compatibility.py
@@ -1,0 +1,39 @@
+from django.http import HttpResponse
+from django.test import AsyncRequestFactory, RequestFactory, TestCase
+
+from debug_toolbar.panels import Panel
+from debug_toolbar.toolbar import DebugToolbar
+
+
+class MockAsyncPanel(Panel):
+    is_async = True
+
+
+class MockSyncPanel(Panel):
+    is_async = False
+
+
+class PanelAsyncCompatibilityTestCase(TestCase):
+    def setUp(self):
+        self.async_factory = AsyncRequestFactory()
+        self.wsgi_factory = RequestFactory()
+
+    def test_panels_with_asgi(self):
+        async_request = self.async_factory.get("/")
+        toolbar = DebugToolbar(async_request, lambda request: HttpResponse())
+
+        async_panel = MockAsyncPanel(toolbar, async_request)
+        sync_panel = MockSyncPanel(toolbar, async_request)
+
+        self.assertTrue(async_panel.enabled)
+        self.assertFalse(sync_panel.enabled)
+
+    def test_panels_with_wsgi(self):
+        wsgi_request = self.wsgi_factory.get("/")
+        toolbar = DebugToolbar(wsgi_request, lambda request: HttpResponse())
+
+        async_panel = MockAsyncPanel(toolbar, wsgi_request)
+        sync_panel = MockSyncPanel(toolbar, wsgi_request)
+
+        self.assertTrue(async_panel.enabled)
+        self.assertTrue(sync_panel.enabled)
--- a/tests/test_middleware_compatibility.py
+++ b/tests/test_middleware_compatibility.py
@@ -1,0 +1,44 @@
+import asyncio
+
+from django.http import HttpResponse
+from django.test import AsyncRequestFactory, RequestFactory, TestCase
+
+from debug_toolbar.middleware import DebugToolbarMiddleware
+
+
+class MiddlewareSyncAsyncCompatibilityTestCase(TestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+        self.async_factory = AsyncRequestFactory()
+
+    def test_sync_mode(self):
+        """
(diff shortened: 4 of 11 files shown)


debug_toolbar/middleware.py
```diff
@@ -64,10 +65,30 @@ class DebugToolbarMiddleware:
     on outgoing response.
     """
 
+    sync_capable = True
+    async_capable = True
+
     def __init__(self, get_response):
+        if get_response is None:
+            raise ValueError("get_response must be provided.")
```
matthiask, line 73: Does this ever happen?

debug_toolbar/middleware.py
```diff
@@ -123,6 +144,64 @@ def __call__(self, request):
...
+        insert_before = dt_settings.get_config()["INSERT_BEFORE"]
+        pattern = re.escape(insert_before)
+        bits = re.split(pattern, content, flags=re.IGNORECASE)
+        if len(bits) > 1:
+            bits[-2] += rendered
+            response.content = insert_before.join(bits)
+            if "Content-Length" in response:
+                response["Content-Length"] = len(response.content)
+        return response
+
```
matthiask, line 204: I think all of that is copy-pasted from `def __call__`? If yes, move all of it to a postprocess method, for example `def _postprocess(self, request, response)`, and call this from both `__call__` and `__acall__`.

matthiask (commented):
Thanks!

matthiask (commented):
LGTM. I wonder how the panels can differentiate in `process_request` between the sync and async version. Also, `await`ing `def process_request` (NOT `async def process_request`) looks a bit strange, but seems fine to me after thinking about it a bit.

debug_toolbar/panels/__init__.py
```diff
@@ -21,6 +24,10 @@ def panel_id(self):
 
     @property
     def enabled(self) -> bool:
+        # check if the panel is async compatible
+        if not self.is_async and isinstance(self.toolbar.request, ASGIRequest):
+            return False
```
tim-schilling, line 29: I think we're missing a test for this logic

salty-ivy:
> LGTM. I wonder how the panels can differentiate in `process_request` between the sync and async version. 

panels would receive either `WSGIRequest` or `ASGIRequest` ( not async ) instance which are subclasses of `HttpRequest`.

> Also, awaiting def process_request (NOT async def process_request) looks a bit strange, but seems fine to me after thinking about it a bit.

`process_request` call of panels would ultimately reach down to view via chain of calls of both panels and then middlewares, the response will always be adapted to async  by `adapt_method_mode` that has to be awaited.

https://github.com/django/django/blob/e56a32b89bb7fadffdfaa2cdf12b4863ccd5af9b/django/core/handlers/base.py#L104

it even adapts the view if its not `async def`  the response will be adapted to async coroutines if the receiving middleware is `async` and  we are running in async context. 
 
https://github.com/django/django/blob/e56a32b89bb7fadffdfaa2cdf12b4863ccd5af9b/django/core/handlers/base.py#L54

this is my overall understanding so far.

debug_toolbar/panels/__init__.py
salty-ivy: this should be covered by panels test seprately?

tests/panels/test_async_panel_compatibility.py
```diff
@@ -0,0 +1,39 @@
+from django.http import HttpResponse
+from django.test import AsyncRequestFactory, RequestFactory, TestCase
+from django.test.utils import override_settings
+
+from debug_toolbar import settings as dt_settings
+from debug_toolbar.toolbar import DebugToolbar
+
+
+def set_custom_toolbar_config():
```
tim-schilling, line 9: Can we add a comment to this method explaining it's significance?

tests/test_middleware_compatibility.py
```diff
@@ -0,0 +1,36 @@
...
+
+        # test middlware switches to sync (__call__) based on get_response type
+        self.assertFalse(asyncio.iscoroutinefunction(middleware))
+
+        response = middleware(request)
+        self.assertEqual(response.status_code, 200)
+
+    async def test_async_mode(self):
+        async def get_response(request):
+            return HttpResponse("<html><body>Django debug toolbar</body></html>")
```
tim-schilling, line 26: How come this response has content, but the `test_sync_mode` doesn't?

tests/test_middleware_compatibility.py
```diff
@@ -0,0 +1,36 @@
...
+class MiddlewareSyncAsyncCompatibilityTestCase(TestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+        self.async_factory = AsyncRequestFactory()
+
+    def test_sync_mode(self):
+        request = self.factory.get("/")
+        middleware = DebugToolbarMiddleware(lambda x: HttpResponse(""))
+
+        # test middlware switches to sync (__call__) based on get_response type
```
tim-schilling, line 18: This comment should be in the docstring for the method definition rather than inline a few lines into the method.

tests/test_middleware_compatibility.py
```diff
@@ -0,0 +1,36 @@
...
+        self.assertEqual(response.status_code, 200)
+
+    async def test_async_mode(self):
+        async def get_response(request):
+            return HttpResponse("<html><body>Django debug toolbar</body></html>")
+
+        middleware = DebugToolbarMiddleware(get_response)
+        request = self.async_factory.get("/")
+
+        # test middlware switches to async (__acall__) based on get_response type
```
tim-schilling, line 31: This comment should be in the docstring for the method definition rather than inline a few lines into the method.

tests/panels/test_async_panel_compatibility.py
```diff
@@ -0,0 +1,39 @@
...
+        self.toolbar = None
+
+    @override_settings(DEBUG_TOOLBAR_CONFIG=set_custom_toolbar_config())
+    def test_async_panel_enabling_with_asgi(self):
+        self.toolbar = DebugToolbar(self.request, lambda request: HttpResponse())
+        for panel in self.toolbar.panels:
+            panel.is_async = False
+            self.assertFalse(panel.enabled)
+            panel.is_async = True
+            self.assertTrue(panel.enabled)
```
tim-schilling, line 28: What's the purpose of this test? Are we testing `Panel.enabled` or each of the individual panels' `enabled` property.

If it's the former, it doesn't need to loop through the panels of the toolbar, it should be able to use an instance of `Panel`. If it's the latter, it shouldn't be setting `is_async` as it changes the panel.

tim-schilling (changes requested):
This is coming along nicely! Thank you for all your work. I have a few comments on particular aspects. We also need to document the new panel API before this can be merged in.

debug_toolbar/middleware.py
```diff
@@ -88,6 +103,36 @@ def __call__(self, request):
             for panel in reversed(toolbar.enabled_panels):
                 panel.disable_instrumentation()
 
+        return self._postprocess(request, response, toolbar)
+
+    async def __acall__(self, request):
```
tim-schilling, line 108: Instead of `__acall__` what do you think about making the async checks a bit more targeted:

```python

    async def aget_response(self, request):
        return await self.get_response(request)

    async def aprocess_request(self, toolbar, request):
        return await toolbar.process_request(request)

    def __call__(self, request):
        # Decide whether the toolbar is active for this request.
        if not show_toolbar(request) or DebugToolbar.is_toolbar_request(request):
            return (
                self.aget_response(request)
                if self.async_mode
                else self.get_response(request)
            )

        toolbar = DebugToolbar(request, self.get_response)

        # Activate instrumentation ie. monkey-patch.
        for panel in toolbar.enabled_panels:
            panel.enable_instrumentation()
        try:
            # Run panels like Django middleware.
            response = (
                self.aprocess_request(toolbar, request)
                if self.async_mode
                else toolbar.process_request(request)
            )
```

cc: @matthiask

tim-schilling: This would get rid of `_postprocess` method and the duplicated logic between `call` and `acall`.

tests/panels/test_async_panel_compatibility.py
salty-ivy: It's for the latter, testing each panel `enabled` method individually. 

The goal is to test each panel in 4 different situations, with `ASGI`  `is_async` true or false and with `WSGI` `is_async` true or false, given that other conditions for each of the panel holds true for enabling. That is why `set_custom_config` is used to make `ProfilingPanel` and `RedirectPanel` enable as well on top of the settings used for testing.

debug_toolbar/middleware.py
salty-ivy: But wouldn't calling async functions like `aget_response` and `aprocess_request` would just return the `coroutine` object rather than the actual value `HttpResponse` in this case, it needs to be awaited  in order to get the actual value that function returns and  awaiting can only be done in async function itself.

then we also need to change other panel methods like `generate_stats` and `generate_server_timing` etc and other parts which are working with response instance directly.

tests/panels/test_async_panel_compatibility.py
tim-schilling: If you're testing the individual panels, the test shouldn't modify the panel. Does that make sense?

debug_toolbar/middleware.py
tim-schilling: Ah, you're right. That was would be a problem for process_request case. The get_response matches the __acall__ approach.

tim-schilling: Yeah, I can't think of a way around that. I'm not super thrilled about the `_process` design, but I don't have anything better at the moment. And we can always change it in the future. Go ahead and revert that change.

tests/panels/test_async_panel_compatibility.py
salty-ivy: I think we can just test them by creating `MockPanel` with `is_async = False` and `is_async = True`
rather than testing each panels individually.

salty-ivy:
I have changed the approach to test async panel compatibility tests.

tests/panels/test_async_panel_compatibility.py
```diff
@@ -1,45 +1,39 @@
...
 
-    Returns a new configuration with DISABLE_PANELS set to an empty dictionary.
-    This allows all panels to be enabled by default.
-    """
-    new_settings = dt_settings.get_config().copy()
-    new_settings["DISABLE_PANELS"] = {}
-    return new_settings
+
+class MockSyncPanel(Panel):
+    is_async = False
```
tim-schilling: Nice, thank you!

tim-schilling (changes requested):
Code-wise, this is great! Thank you Aman. I think we need to update the docs in two areas:

1. Change `Panel.is_async` to a `@property` method with a doc string, then update https://github.com/jazzband/django-debug-toolbar/blob/main/docs/panels.rst?plain=1#L332 to include that attribute in the documentation.
2. Update the architecture document here: https://github.com/jazzband/django-debug-toolbar/blob/9bcd6cac17fd1721d60c2b1b8218884caf0ee45e/docs/architecture.rst?plain=1#L82-L84

salty-ivy:
> 1. Change `Panel.is_async` to a `@property` method with a doc string, then update https://github.com/jazzband/django-debug-toolbar/blob/main/docs/panels.rst?plain=1#L332 to 

Sorry for the nit pick question but should we use `@classproperty` or `@property` with class level private attribute `_is_async`?  my main idea is that is should be class level we might need to access it in future directly via class and makes subclassing easy

tim-schilling:
TIL, then let's go with what Matthias suggested. Thank you for pointing that out

matthiask:
If you search this page for `#:` you'll find examples:
https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoproperty

So, if the autodoc support is the only reason for converting the attribute to a property I think I'm -0 on it. No strong opposition certainly, I just don't see the need for it.

tim-schilling:
I wanted the autodoc and then thought it would bring consistency to the API. Though you and @salty-ivy both questioning it is enough for me to agree to having it as an attribute.

salty-ivy:
> 2. Update the architecture document here: https://github.com/jazzband/django-debug-toolbar/blob/9bcd6cac17fd1721d60c2b1b8218884caf0ee45e/docs/architecture.rst?plain=1#L82-L84

So we are left with this only while keeping `is_async` a class level attribute for now.

tim-schilling:
Sounds good.

salty-ivy:
Not sure, may be something related to sphinx missing  in the updated docs to throw the lint error?

salty-ivy:
> You need to change `enviroment` to `environment`

ahh, how can I miss this, thank you so much. Interestingly It even fails on `ASGI` keyword.

tim-schilling:
See! I didn't even catch that other one. I thought it was two lines because the word was used twice on that line :grin:

tim-schilling (approved):

matthiask (approved):

tim-schilling:
I squashed the commits down to one, cleaned up the commit message, and rebased on main. After tests pass, I'll merge it in

PR #1942, opened 2024-07-04
Remove rem units from svg

### LINKED ISSUE #1935: SVG definitions have width in rem which throws console errors
The problematic definitions are in this file: https://github.com/jazzband/django-debug-toolbar/blob/0a622dcab4c78ce0f1d339f0e3147a1023350efa/debug_toolbar/templates/debug_toolbar/includes/theme_selector.html#L5

As I understand it, SVGs cannot define height in em or rem, but require pixels.

The current status results in errors in the console on Firefox stating "Unexpected value 1rem parsing width attribute." - 6 errors in total, one for each SVG definition. 

Tested on Edge too and that does not seem to produce the errors, so this looks to be browser dependent. Looking at similar issues on other repos, I suspect Safari would also show them, but I am not using apple so cannot verify this. 

Happy to make the change and submit a PR if somebody can say which pixel size they would need.

michjnich:
# Description

Removes the "1rem" height and width from the svg attributes for the dark/light icons and puts it into css height and width attributes instead

Fixes #1935 

# Checklist:

- [x] I have added the relevant tests for this change. (*Note: No tests added*)


michjnich:
Quick note: I couldn't run full tests here without setting up a load of databases locally (side note, might be an idea to include a compose file that sets up the various dbs for the full tox run, to make life simpler), and since this was a pure display change I figured that was probably OK :)

tim-schilling (approved):

tim-schilling:
Looks good! I'll pull this down, verify it's right then merge. 

We don't expect people to run the tests locally against all DBs unless that's what the change is relevant for. Testing against sqlite is sufficient.

tim-schilling:
Thank you @michjnich!

michjnich:
And thank you for a pleasant contributor experience :)

PR #1970, opened 2024-07-22
Make toolbar compatible with `FORCE_SCRIPT_NAME`

### LINKED ISSUE #1961: History sidebar panel keeps loading infinitely?
Hey,

not sure what I'm mistaking here, but as I open the history panel (without using daphne so it can be turned on), I get endless requests upon trying to open thw history panel:

```
[09/Jul/2024 20:03:01] "GET /__debug__/history_sidebar/?store_id=665facf9fc8043268809c407f26c9239 HTTP/1.1" 200 10752
[09/Jul/2024 20:03:01] "GET /__debug__/history_sidebar/?store_id=ce761655224342c5a2ee91f81ba1d6a0 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:02] "GET /__debug__/history_sidebar/?store_id=7064e973b5434229a62c5e7e05f99c4f HTTP/1.1" 200 10691
[09/Jul/2024 20:03:02] "GET /__debug__/history_sidebar/?store_id=2dc12ca23eec4949bd50bada5c9aa21d HTTP/1.1" 200 10691
[09/Jul/2024 20:03:02] "GET /__debug__/history_sidebar/?store_id=95ea6042de90411a9a9475f860e24e60 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:03] "GET /__debug__/history_sidebar/?store_id=459c1020581f48bbb01cccd6ec83d53e HTTP/1.1" 200 10691
[09/Jul/2024 20:03:03] "GET /__debug__/history_sidebar/?store_id=243033191ce542dcb8ef39ea3cfd58b3 HTTP/1.1" 200 10693
[09/Jul/2024 20:03:03] "GET /__debug__/history_sidebar/?store_id=8729ff48831942adab47caa2d3d747cc HTTP/1.1" 200 10691
[09/Jul/2024 20:03:03] "GET /__debug__/history_sidebar/?store_id=43841f09a4cf4821824ea9f1e990b749 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:04] "GET /__debug__/history_sidebar/?store_id=52787c0bb5704a9ea86ae6612e92a53c HTTP/1.1" 200 10691
[09/Jul/2024 20:03:04] "GET /__debug__/history_sidebar/?store_id=090b9eab8c5a4e09b614b45eb81652f1 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:04] "GET /__debug__/history_sidebar/?store_id=c471d653bf054895af20625207e4adf9 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:04] "GET /__debug__/history_sidebar/?store_id=c4a770548a6347bdae27f0c880c2cb15 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:05] "GET /__debug__/history_sidebar/?store_id=ac146df811cb46a691f0ed646bc44df2 HTTP/1.1" 200 10691
[09/Jul/2024 20:03:05] "GET /__debug__/history_sidebar/?store_id=464f724930df458f87f83aad07a3f231 HTTP/1.1" 200 10691
```

It is about 3 requests/sec and I have no idea if that is alright or where to configure it to emit less requests. Normally I would prefer that it only gets the request history when I open the panel, but I couldn't find such an option. It is super hard to find information for the request I really need because what I see is this:

image

Any help here?

dmartin:
# Description

Previously, if a project used the [`FORCE_SCRIPT_NAME`](https://docs.djangoproject.com/en/5.0/ref/settings/#force-script-name) setting (such as when [hosting a Django application under a subdirectory path via a reverse proxy](https://stackoverflow.com/a/28407366/1508246)), `is_toolbar_request()` would always return False.

This is because `request.resolver_match` is not yet available in the middleware context, so `django.urls.resolve()` is [called instead](https://github.com/jazzband/django-debug-toolbar/blob/f2e389cdbef0a214f51f3c49a35b32e4295dd9c9/debug_toolbar/toolbar.py#L167-L169). `resolve()` then raised `Resolver404`, because it [does not take `FORCE_SCRIPT_NAME` into account](https://code.djangoproject.com/ticket/31724).

This caused internal toolbar URLs to be inspected, leading to a request loop when refreshing the history panel.

This PR makes calls to `django.urls.resolve()` take [`request.path_info`](https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpRequest.path_info) rather than `request.path`. This effectively strips the request's leading `SCRIPT_NAME` if one is present.

Fixes #1961 

# Checklist:

### DIFF
--- a/debug_toolbar/panels/request.py
+++ b/debug_toolbar/panels/request.py
@@ -41,7 +41,7 @@ class RequestPanel(Panel):
             "view_urlname": "None",
         }
         try:
-            match = resolve(request.path)
+            match = resolve(request.path_info)
             func, args, kwargs = match
             view_info["view_func"] = get_name_from_obj(func)
             view_info["view_args"] = args
--- a/debug_toolbar/toolbar.py
+++ b/debug_toolbar/toolbar.py
@@ -13,7 +13,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.dispatch import Signal
 from django.template import TemplateSyntaxError
 from django.template.loader import render_to_string
-from django.urls import get_script_prefix, include, path, re_path, resolve
+from django.urls import include, path, re_path, resolve
 from django.urls.exceptions import Resolver404
 from django.utils.module_loading import import_string
 from django.utils.translation import get_language, override as lang_override
@@ -165,8 +165,7 @@ class DebugToolbar:
         # not have resolver_match set.
         try:
             resolver_match = request.resolver_match or resolve(
-                request.path.replace(get_script_prefix(), "/", 1),
-                getattr(request, "urlconf", None),
+                request.path_info, getattr(request, "urlconf", None)
             )
         except Resolver404:
             return False
--- a/debug_toolbar/toolbar.py
+++ b/debug_toolbar/toolbar.py
@@ -13,7 +13,7 @@ from django.core.exceptions import ImproperlyConfigured
 from django.dispatch import Signal
 from django.template import TemplateSyntaxError
 from django.template.loader import render_to_string
-from django.urls import include, path, re_path, resolve
+from django.urls import get_script_prefix, include, path, re_path, resolve
 from django.urls.exceptions import Resolver404
 from django.utils.module_loading import import_string
 from django.utils.translation import get_language, override as lang_override
@@ -165,7 +165,8 @@ class DebugToolbar:
         # not have resolver_match set.
         try:
             resolver_match = request.resolver_match or resolve(
-                request.path, getattr(request, "urlconf", None)
+                request.path.replace(get_script_prefix(), "/", 1),
+                getattr(request, "urlconf", None),
             )
         except Resolver404:
             return False
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,6 +4,9 @@ Change log
 Pending
 -------
 
+* Fixed internal toolbar requests being instrumented if the Django setting
+  ``FORCE_SCRIPT_NAME`` was set.
+
 4.4.6 (2024-07-10)
 ------------------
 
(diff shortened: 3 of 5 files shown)


matthiask (commented):
Could you try using request.path_info instead of request.path? The former should do the right thing if Django isn't mounted at '/' without having to use the script prefix explicitly. (I haven't tested it myself.)

dmartin:
> Could you try using request.path_info instead of request.path? The former should do the right thing if Django isn't mounted at '/' without having to use the script prefix explicitly. (I haven't tested it myself.)

Oh, that does work and is a lot cleaner! Thanks for the suggestion. I can make that change, but what would you like to do with the existing tests that just override `request.path`? For example:

```python
@override_settings(ROOT_URLCONF="tests.urls_invalid")
    def test_is_toolbar_request_override_request_urlconf(self):
        """Test cases when the toolbar URL is configured on the request."""
        self.request.path = "/__debug__/render_panel/"
        self.assertFalse(self.toolbar.is_toolbar_request(self.request))
```
I can also add matching request.path_info overrides to each of these, or I suppose they could be outright replaced with request.path_info overrides for all tests except mine which is explicitly checking for when they differ.

tim-schilling:
@dmartin could we switch them with a `RequestFactory.get('/__debug__/render_panel/')` instance?

dmartin:
I made those changes, though I didn't rework every test to use a RequestFactory to keep the scope of the PR under control, only the tests that were manually setting `request.path`.

tim-schilling (approved):
I'm good with this change.

It definitely highlights a problem we have in the test suite where we manipulate a request after initializing the toolbar. It's possible we may run into a bug where the the toolbar does something slightly differently because of this. However, if that were the case, I would have expected it to crop up before now.

matthiask:
Thank you!

Issue #1791, opened 2023-05-31
Missing connection.queries


ulgens:
After upgrading our project to `django-debug-toolbar==4.1`, we started having many test failures. Almost all of `self.assertNumQueries(<query_count>)` checks started failing saying the query count is different now. The queries we checked were actually the same as before, they were returning the same calls and returning the same data, but when I checked `connection.queries`, a majority of the queries were seemingly missed.

I did some testing with different commits and this problem starts with https://github.com/jazzband/django-debug-toolbar/commit/e7575e87dc9e2d2560b87d6fd5a123b9398cbd34

Not sure what else to add but if needed, I can try to create a small project reproducing this issue.

tim-schilling:
Thank you for opening the issue. Creating a reproducible project would be very helpful.

todorvelichkov:
We are facing kind of a similar issue, we have a `logging.FileHandler` that write all queries into a file, when DJDT is ON, no queries are being written into the file, however, the debug toolbar itself "sees" the queries (in the SQL tab).

tim-schilling:
@ulgens @todorvelichkov I can't investigate this without a test case or more information on how to reproduce this.

bluetech:
Probably related to this report: since debug-toolbar 4.1 the `django.db.backends` SQL logger no longer works. I bisected this to commit e7575e87dc9e2d2560b87d6fd5a123b9398cbd34 "Inherit from django.db.backends.utils.CursorWrapper". From a quick look, the `django.db.backends` query logging is implemented in Django using a special `CursorDebugWrapper` that is instantiated instead of the normal `CursorWrapper` if `settings.DEBUG` is set:

https://github.com/django/django/blob/2584783f46922bcb456ceb9700a3726314df65d3/django/db/backends/base/base.py#L279-L288

I'm guessing something in the new commit now makes it use DJDT's cursor wrapper instead of the `CursorDebugWrapper`.

akx:
@elnygren noted that this also breaks `pytest-django`'s `django_assert_num_queries` and `django.test.utils. CaptureQueriesContext`.

elnygren:
Also this person may be having the same problem https://github.com/pytest-dev/pytest-django/issues/1068

A really simple fix for testing is to `INSTALLED_APPS.remove("debug_toolbar")` which I guess makes sense to do anyway 😅

crccheck:
[...]

PR #1971, opened 2024-07-22
add self.assertNumQueries to test

crccheck:
# Description

this information as well.

Fixes # (issue)

# Checklist:

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -58,7 +58,8 @@ class SQLPanelTestCase(BaseTestCase):
     def test_recording(self):
         self.assertEqual(len(self.panel._queries), 0)
 
-        sql_call()
+        with self.assertNumQueries(1):
+            sql_call()
 
         # ensure query was logged
         self.assertEqual(len(self.panel._queries), 1)


tim-schilling:
@crccheck I think this is a WIP. I'm inclined to close it to keep the open PRs minimal until there's more of a direction for the work that requires discussion/review.

matthiask:
I think there's some value in asserting how many quietus Django sees vs how many queries the toolbar sees. I think both functionalities use the same underlying data structure, which makes this additional assertion not very useful. Or maybe I'm missing something?

tim-schilling:
@crccheck I'm going to close this. If you feel like this is important, please let us know why and we can reopen it.

crccheck:
oh sorry, I only opened it to see if CI would run it. I figured out how a minimal command to get tests to run locally since then and I thought draft PRs wouldn't show up

crccheck:
FYI @matthiask there's an issue where Django's queries actually don't match Debug Toolbar #1791 I was trying to recreate

> I think there's some value in asserting how many quietus Django sees vs how many queries the toolbar sees. I think both functionalities use the same underlying data structure, which makes this additional assertion not very useful. Or maybe I'm missing something?

tim-schilling:
Ah, that makes sense. Thanks for following up!

PR #1972, opened 2024-07-24
Support select and explain for UNION queries

friedelwolff:
Some UNION queries can start with "(", causing it to not be classified as a SELECT query, and consequently the select and explain buttons are missing on the SQL panel.

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -729,10 +729,7 @@ class SQLPanelTestCase(BaseTestCase):
         self.assertNotEqual(queries[0]["similar_color"], queries[3]["similar_color"])
         self.assertNotEqual(queries[0]["duplicate_color"], queries[3]["similar_color"])
 
-    @unittest.skipUnless(
-        connection.vendor == "postgresql", "Test valid only on PostgreSQL"
-    )
-    def test_explain(self):
+    def test_explain_with_union(self):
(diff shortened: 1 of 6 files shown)


tests/panels/test_sql.py
```diff
@@ -729,6 +729,16 @@ def test_similar_and_duplicate_grouping(self):
         self.assertNotEqual(queries[0]["similar_color"], queries[3]["similar_color"])
         self.assertNotEqual(queries[0]["duplicate_color"], queries[3]["similar_color"])
 
+    @unittest.skipUnless(
+        connection.vendor == "postgresql", "Test valid only on PostgreSQL"
```
tim-schilling, line 733: How come this is only valid for postgres? I feel like the other db drivers should also pass this test.

friedelwolff: From a quick look at the Django code, this can only happen if the DB backend has `supports_slicing_ordering_in_compound = True`. But I see now it is more widely supported than I thought. Shall I rather use
```python
@unittest.skipUnless(connection.vendor != "sqlite", "...")
```
?

matthiask: I'm not sure why the `skipUnless` is even necessary? The test should pass on all database engines anyway, or not?

tim-schilling:
Thank you @friedelwolff!

PR #1975, opened 2024-07-30
Quick hack for including csp_nonces from requests into script tags

### LINKED ISSUE #1723: Content-Security-Policy and strict-dynamic
Django Debug Toolbar is blocked by CSP if we are using strict-dynamic.
Django already support it from the box if we can include `nonce="{{request.csp_nonce}}"` in the template

karolyi:
This should fix #1723 as it optionally includes the `csp_nonce` from the [`django-csp`](/mozilla/django-csp) module.

### DIFF
--- a/tests/test_csp_rendering.py
+++ b/tests/test_csp_rendering.py
@@ -8,7 +8,7 @@ from html5lib.html5parser import HTMLParser
 
 from debug_toolbar.toolbar import DebugToolbar
 
-from .base import BaseTestCase
+from .base import IntegrationTestCase
 
 
 def get_namespaces(element: Element) -> Dict[str, str]:
@@ -22,10 +22,11 @@ def get_namespaces(element: Element) -> Dict[str, str]:
 
 
 @override_settings(DEBUG=True)
-class CspRenderingTestCase(BaseTestCase):
+class CspRenderingTestCase(IntegrationTestCase):
     """Testing if `csp-nonce` renders."""
 
     def setUp(self):
+        super().setUp()
         self.parser = HTMLParser()
 
     def _fail_if_missing(
--- a/tests/test_csp_rendering.py
+++ b/tests/test_csp_rendering.py
@@ -1,0 +1,103 @@
+from typing import Dict
+from xml.etree.ElementTree import Element
+
+from django.conf import settings
+from django.http.response import HttpResponse
+from django.test.utils import ContextList, override_settings
+from html5lib.constants import E
+from html5lib.html5parser import HTMLParser
+
+from .base import BaseTestCase
+
+
+def _get_ns(element: Element) -> dict[str, str]:
+    """
+    Return the default `xmlns`. See
+    https://docs.python.org/3/library/xml.etree.elementtree.html#parsing-xml-with-namespaces
+    """
+    if not element.tag.startswith('{'):
+        return dict()
+    return {'': element.tag[1:].split('}', maxsplit=1)[0]}
+
+
+class CspRenderingTestCase(BaseTestCase):
+    'Testing if `csp-nonce` renders.'
+    panel_id = "StaticFilesPanel"
+
+    # def setUp(self):
+    #     self.factory = RequestFactory()
+    #     self.async_factory = AsyncRequestFactory()
+
+    def _fail_if_missing(
+            self, root: Element, path: str, namespaces: Dict[str, str],
+            nonce: str):
+        """
+        Search elements, fail if a `nonce` attribute is missing on them.
+        """
+        elements = root.findall(path=path, namespaces=namespaces)
+        for item in elements:
+            if item.attrib.get('nonce') != nonce:
+                raise self.failureException(f'{item} has no nonce attribute.')
+
+    def _fail_if_found(
+            self, root: Element, path: str, namespaces: Dict[str, str]):
+        """
+        Search elements, fail if a `nonce` attribute is found on them.
+        """
+        elements = root.findall(path=path, namespaces=namespaces)
+        for item in elements:
+            if 'nonce' in item.attrib:
+                raise self.failureException(f'{item} has no nonce attribute.')
+
+    def _fail_on_invalid_html(self, content: bytes, parser: HTMLParser):
+        'Fail if the passed HTML is invalid.'
+        if parser.errors:
+            default_msg = ['Content is invalid HTML:']
+            lines = content.split(b'\n')
+            for position, errorcode, datavars in parser.errors:
+                default_msg.append('  %s' % E[errorcode] % datavars)
+                default_msg.append('    %r' % lines[position[0] - 1])
+            msg = self._formatMessage(None, '\n'.join(default_msg))
+            raise self.failureException(msg)
+
+    @override_settings(
+        DEBUG=True, MIDDLEWARE=settings.MIDDLEWARE + [
+            'csp.middleware.CSPMiddleware'
+        ])
+    def test_exists(self):
+        'A `nonce` should exists when using the `CSPMiddleware`.'
+        response = self.client.get(path='/regular/basic/')
+        if not isinstance(response, HttpResponse):
+            raise self.failureException(f'{response!r} is not a HttpResponse')
+        self.assertEqual(response.status_code, 200)
+        parser = HTMLParser()
+        el_htmlroot: Element = parser.parse(stream=response.content)
+        self._fail_on_invalid_html(content=response.content, parser=parser)
+        self.assertContains(response, 'djDebug')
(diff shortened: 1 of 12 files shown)


matthiask:
Thanks. Looks good. I'd like to have some tests for this; django-csp could be made a dependency for the testsuite so that we can actually verify that the code works as intended.

karolyi:
What bothers me more is that normally, link/style tags would require the nonces too. But since at its current state, [`'strict-dynamic'`-enabled `script-src` scripts can't inject styles](https://github.com/w3c/webappsec-csp/issues/625) which a lot of frameworks do), everyone just enables inlining styles and sameorigins in practice.

If this ever changes and someone enables `'strict-dynamic'` for `style-src`, that will override `'self'` on style CSPs and every `link[rel=stylesheet]`/`style` tag will require a nonce too from that point on.

I might add that preventative measure. The tests can be organized.

matthiask:
Sure! Feel free to include that.

karolyi:
So, tests are there, all working.

I'm not totally against the enforced opinionated formatting on someone's code (it's like no one is allowed to have their own code formatting), but so be it. I can just as much format the code back automatically when doing _free and voluntary contributions_ to other projects.

In fact, I timed the `dict()` vs `{}` call with `timeit` and the latter seemed faster, so as per empirical evidence, I'll went with that.

Please merge and do a timely release.

karolyi:
So, tests are there, all working.

I'm totally against the enforcing of opinionated formatting on someone else's code (it's like no one is allowed to have their own code formatting), but so be it. I can just as much format the code back automatically when doing free and voluntary contributions to other projects.

In fact, I timed the `dict()` vs `{}` call with `timeit` and the latter seemed faster, so as per empirical evidence, I went with that.

Please merge and do a timely release.

matthiask:
Thanks for your contribution!

We're using code formatters so that we do not have to discuss code style over and over, as do many other projects. Also, we're not being paid for the work we're doing on this project either, free and voluntary here too. 

The PR looks good to me at first sight, thanks. I'll take a closer look probably tomorrow.

tim-schilling:
@matthiask I'm unsure about the type definition changes here. If you're good with them, I'll let it be.

My specific concern is that we don't have a formal policy for them and I'm not sure how this current state improves things. I'd prefer to be consistent in the library so it's easier for folks to navigate around.

matthiask:
@tim-schilling My general opinion is that type annotations can be noisy, and not really worth it unless a big part of the code base is typed, but we have already started moving on this https://github.com/jazzband/django-debug-toolbar/issues/1705 so I'm neutral to slightly fine with it.

karolyi:
For context, I'm using pyright and with these added typings, my test and everything I did, checks out perfectly.

Don't be such an enemy of typing that will point out errors early with static analysis.

tim-schilling:
But the project doesn't use pyright, so it's possible for this to get out of date and cause confusion down the line.

karolyi:
You are confusing python typing and static analysis with pyright. The former will be valid even when the latter changes, as it's defined in python itself, and not the module.

There is only one comment in my code related to pyright, since I didn't want to go and add a missinfg type annotation in Django itself. The rest is standard python typing stuff.

tim-schilling:
Sorry, I shouldn't have even started is down that discussion. Thank you for the PR, this will be great to have.

tim-schilling:
@matthiask when you have time again, take a look at the revised PR. I'm now good with merging this.

tests/test_csp_rendering.py
```diff
@@ -0,0 +1,148 @@
...
+        self._fail_if_missing(
+            root=html_root, path=".//script", namespaces=namespaces, nonce=nonce
+        )
+
+    @override_settings(
+        MIDDLEWARE=settings.MIDDLEWARE + ["csp.middleware.CSPMiddleware"]
+    )
+    def test_panel_content_nonce_exists(self):
+        response = self.client.get("/regular/basic/")
+        if not isinstance(response, HttpResponse):
```
tim-schilling, line 117: Can you help me understand the purpose of this check? My understanding is that the test client only returns http responses.

tests/test_csp_rendering.py
```diff
@@ -0,0 +1,148 @@
...
+            raise self.failureException(f"{response!r} is not a HttpResponse")
+        self.assertEqual(response.status_code, 200)
+
+        html_root: Element = self.parser.parse(stream=response.content)
+        self._fail_on_invalid_html(content=response.content, parser=self.parser)
+        self.assertContains(response, "djDebug")
+
+        namespaces = get_namespaces(element=html_root)
+        context: ContextList = response.context  # pyright: ignore[reportAttributeAccessIssue]
+        nonce = str(context["toolbar"].request.csp_nonce)
```
tim-schilling, line 104:
```suggestion
        toolbar = list(DebugToolbar._store.values())[0]
        nonce = str(toolbar.request.csp_nonce)
```

tests/test_csp_rendering.py
karolyi: It does yes, but the typing is wrong if you don't do this check:

Without, the source lights up in a type checking editor with errors:
image

With:
image

Basically, type checkers rightly think that walking back on the `self.client.get`, it returns a WSGIRequest and not a HttpResponse. Which is a safe assumption, since the typing is wrong in Django itself: if you go back up to [`django/test/client.py:914`](https://github.com/django/django/blob/7b1a76f899f7a7acb7d70b433cb0064c2184186b/django/test/client.py#L914), you'll see that the type inference rightly assumes that it'll a WSGIRequest, which is not the practical case. 

I didn't delve into it deeper now, but like I said earlier, it is easier to right a wrong here than to go through a very bureucratic process to get this fixed. The syntax itself is valid and ensures that type checking also is right.

tests/test_csp_rendering.py
karolyi: Same issue here as above.

It might be inconvenient for others, but I can assure you I have to fight with these kinda errors throughout everywhere when working with django, even when having the stubs installed.

It is not perfect but it's the best we have. I have already made the compromise to have my code "autocorrected" (in the sense of might makes right), so I suggest others make compromises here too.

tests/test_csp_rendering.py
matthiask: I'm not a typing expert at all but wouldn't something like `response = typing.cast(self.client.get(...), HttpResponse)` also work, maybe with a comment which says that we're helping the type checker understand the correct type? Otherwise we have code in the testsuite which is (probably) never run. Almost as bad as untested code in the project itself.

karolyi: Ah, so that's your main concern here, the coverage ... I stopped caring about coverage in these minuscule percents a long time ago.

But, be that as it may, I'll change it to using `cast`.

matthiask: No, it's not really about coverage, it's about communicating the intent more clearly. The check as it's written currently communicates that `client.get` might return something which isn't a `HttpResponse`; `typing.cast` communicates that the type checker needs some help here. At least that's my reading of it.

Thanks for your patience. It's not my (our) intention to make the contribution hard. Please assume good intent as well.

karolyi: No offense taken :)

I've already expressed my concerns earlier, not repeating myself meaninglessly.

karolyi:
Whoa, ruff was pretty bold here removing the `HttpRequest` import :D Hence why everything's failing.

matthiask:
Thanks!

karolyi:
@matthiask you're welcome.

Can I have a release please, so I could use the actual official package, instead of my fork?

matthiask:
@karolyi We're currently in the middle of a GSoC project adding async support to the toolbar, so I'm not sure if we want to do a release in the next few days without some additional testing first.

I generally think releasing quickly and trusting the test suite is the way to go, but I'm a bit wary now because of this larger project.

karolyi:
I see thanks. I've already set a notification on the releases of this project, so I'll be notified if there is a new one.

Until then, I'll keep using my fork.

karolyi:
Congratulations on the release. :)

Zerotask:
When updating from 4.4.6 to 5.0.1 some pages stop working properly (e.g. charts won't be rendered) and in the browser console we get:
image

When I uncomment `CSPMiddleware` in our settings, it works again. So this release somehow added cap related issues.

I can't find any information about this breaking change but just a reference in the release notes. How can I make it work again?

karolyi:
It is kinda offtopic here imo since it's because of your particular CSP setting, I think.

The error lines are descriptive (if you read them a couple times again and again) and should point out the error sufficiently.

PR #1976, opened 2024-07-31
Async compatible redirect panel

salty-ivy:
# Description

async compatible `RedirectsPanel`.

# Checklist:

### DIFF
--- a/debug_toolbar/panels/redirects.py
+++ b/debug_toolbar/panels/redirects.py
@@ -38,12 +38,17 @@ class RedirectsPanel(Panel):
                 response.render()
         return response
 
-    async def aprocess_request(self, request):
-        response = await super().process_request(request)
+    async def aprocess_request(self, request, response_coroutine):
+        """
+        Async version of process_request. used for accessing the response
+        by awaiting it when running in ASGI.
+        """
+
+        response = await response_coroutine
         return self._process_response(response)
 
     def process_request(self, request):
         response = super().process_request(request)
         if iscoroutine(response):
-            return self.aprocess_request(request)
+            return self.aprocess_request(request, response)
         return self._process_response(response)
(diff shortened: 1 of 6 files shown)


debug_toolbar/panels/redirects.py
```diff
@@ -9,23 +11,34 @@ class RedirectsPanel(Panel):
...
+        """
+        Common response processing logic.
+        """
         if 300 <= response.status_code < 400:
             redirect_to = response.get("Location")
             if redirect_to:
                 status_line = f"{response.status_code} {response.reason_phrase}"
                 cookies = response.cookies
                 context = {"redirect_to": redirect_to, "status_line": status_line}
-                # Using SimpleTemplateResponse avoids running global context processors.
```
tim-schilling, line 25: Let's not remove this comment.

debug_toolbar/panels/redirects.py
```diff
@@ -9,23 +11,34 @@ class RedirectsPanel(Panel):
...
         return response
+
+    async def aprocess_request(self, request):
+        response = await super().process_request(request)
+        return self._process_response(response)
+
+    def process_request(self, request):
+        response = super().process_request(request)
+        if iscoroutine(response):
+            return self.aprocess_request(request)
```
tim-schilling, line 43: This flow will call super().process_request twice. We probably want to avoid that. I have some ideas, but none that I know are great. What do you think?

salty-ivy: since `aprocess_request` is exclusive to `RedirectsPanel`
1. we can pass response from `process_request` in `aprocess_request` as an argument.
2. or else we change the check from coroutine to checking request being an instance `ASGIRequest`.

tim-schilling: My preference is option 1

salty-ivy: I think 2. won't work either : )

tim-schilling (approved):

tim-schilling:
Looks good! Thanks for working with me to arrive at this solution.

Issue #1918, opened 2024-05-26
Move Django Debug Toolbar to Django Commons


tim-schilling:
I'd like to transfer the project over to https://github.com/django-commons

Matthias and I both plan on helping with the organization and are the leads on the project here with Jazzband. I don't believe it's a reduction in maintenance quality. There may be a few growing pains with releasing as there are no actual projects with Django Commons. The other issue is that links to https://github.com/django-debug-toolbar/django-debug-toolbar/ may not resolve with the latest destination.

It should also allow us to bypass https://github.com/jazzband/help/issues/360 and revert #1916

matthiask:
I'm seconding this request.

matthiask:
By the way, I only now see that this is in our project; maybe this would better be opened in https://github.com/jazzband/help/issues

On the other hand, since it's possible that we will wait for a long time we should have a backup plan. We could look into migrating issues and comments (or maybe only links to old issues) using the GitHub API and replace jazzband/django-debug-toolbar with only a bare README which points everyone to the new location. A work colleague has transferred hundreds of issues from our decomissioned https://taiga.io/ instance into GitHub and I could probably ask him for pointers or code on how to do this.

tim-schilling:
Oh, we definitely need to open a roadies help question to move it out. But I also wanted to have an issue here for a history of the project.

jezdez:
Is this still something you'd like doing, is Django Commons ready to take it on? Let's do a proper transfer so that redirects keep working.

tim-schilling:
Hi Jannis, yes it is. Django Commons is nearly ready. I suspect early August we can move the toolbar over if that works for you.

jezdez:
Sounds good to me, thanks for the quick response

tim-schilling:
@jezdez alright, we're formally ready to do the transition: https://github.com/jazzband/help/issues/369

There isn't a major rush on this. @cunla is going through the process on September 4th for django-fsm2. There may be some bumps for our first few transitions.

tim-schilling:
This process is in motion. Updates should be followed on https://github.com/jazzband/help/issues/369 and https://github.com/django-commons/membership/issues/34

tim-schilling:
@elineda @VeldaKiara @living180 Do you mind joining Django Commons so we can grant you your elevated permissions on the project when it moves over?

https://github.com/django-commons/membership/issues/new/choose

elineda:
done

[...]

PR #1978, opened 2024-08-02
Switch to Django Commons code of conduct

tim-schilling:
When we transfer over, we should add a line to the changes.rst file.

References #1918

# Checklist:


matthiask (approved):
Let's do this!

PR #1979, opened 2024-08-05
Add a paragraph describing our stance on Python typing

matthiask:
I think we should an official paragraph to the contributing docs regarding our stance on typing. I tried describing what we're doing currently since I think it makes sense, but I'm definitely open to changing this. I don't think we should open a big round of discussions on this though.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -112,10 +112,10 @@ For MySQL/MariaDB in a ``mysql`` shell::
 Style
 -----
 
-The Django Debug Toolbar uses `black <https://github.com/psf/black>`__ to
-format code and additionally uses ruff. The toolbar uses
-`pre-commit <https://pre-commit.com>`__ to automatically apply our style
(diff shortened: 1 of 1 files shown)


tim-schilling (approved):
Seems fine. My issue is about adding code to silence/please people's personal environments because they have it configured differently than the baseline.

matthiask:
Do you think we should add something to address this concern? I totally agree with you but I have a hard time finding ways to describe this without referring to particular cases.

tim-schilling:
I don't know. It feels like it's a thing we need to commit to or not. Otherwise we end up with code like [this](https://github.com/jazzband/django-debug-toolbar/blob/main/tests/test_csp_rendering.py#L69), which I'd argue is adding code to please a configuration.

I think merging this PR is a step forward regardless.

matthiask:
OK let's do it step by step. Thank you!

PR #1982, opened 2024-08-06
Slightly increase opacity of debug toolbar button

### LINKED ISSUE #1981: Debug Toolbar Button has too low contrast
When the page background is white (obviously very common), the opacity and colouring of the debug toolbar button is such that the colour contrast of the text is too low as per accessibility rules (it's 2.61, it should be >=3). If the opacity is increased from 0.5 to 0.6 then this issue will be fixed.

(The low-contrast is an accessibility issue in and of itself, but also it means that when developers are performing accessibility testing during development, they will get warnings about the low contrast of the button, which are distracting and spurious since the button will of course not be present on the production site.)

jribbens:
# Description

Avoids an accessibility issue (low-contrast text) when the page behind the button is white.

Fixes #1981

# Checklist:

Too minor a change for tests or changelog.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -301,7 +301,7 @@
     font-size: 22px;
     font-weight: bold;
     background: #000;
-    opacity: 0.5;
+    opacity: 0.6;
 }
 
 #djDebug #djShowToolBarButton:hover {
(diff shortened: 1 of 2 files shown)


tim-schilling:
Before:
Screenshot from 2024-08-06 06-56-59


After:
Screenshot from 2024-08-06 06-56-52

tim-schilling (approved):
Thank you for the change!

PR #1991, opened 2024-08-20
Async compatible `HistoryPanel`

salty-ivy:
# Description

adding a special check  on `should_render_panel` method in `DebugToolbar` that bypasses `request.multiprocess` check and  allows rendering at async true

Relevant issues: https://github.com/jazzband/django-debug-toolbar/issues/1413

# Checklist:

### DIFF
--- a/debug_toolbar/toolbar.py
+++ b/debug_toolbar/toolbar.py
@@ -12,6 +12,7 @@ from typing import OrderedDict
 from django.apps import apps
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+from django.core.handlers.asgi import ASGIRequest
 from django.dispatch import Signal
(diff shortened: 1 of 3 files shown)


debug_toolbar/toolbar.py
```diff
@@ -106,7 +107,10 @@ def should_render_panels(self):
             # incompatible with the toolbar until
             # https://github.com/jazzband/django-debug-toolbar/issues/1430
             # is resolved.
-            render_panels = self.request.META.get("wsgi.multiprocess", True)
+            if isinstance(self.request, ASGIRequest):
+                render_panels = False
```
tim-schilling, line 111: This conditional doesn't align with the comment above it. If the comment is to be trusted, then we can change the default value to False and avoid this check on `ASGIRequest`.

Issue #932, opened 2017-04-18
It doesnot count query while making query from another thread


amankumarjain:
How to debug database query when query is executed from separate thread.

diegobill:
Same problem +1

matthiask:
This may be because Django ensures that all database connections are cursors differ between threads. 

I think it may be very hard to implement cross-thread tracking of queries; there probably aren't easy ways to determine the originating thread in the general case; patching the `threading` module may work but maybe it won't. Also, the work would probably have to be redone for the async case. So, if anyone has a good idea how to implement this, please contribute (code or ideas) but for now I'm leaning towards closing this as not planned.

PR #1993, opened 2024-08-20
Async compatible `SQLPanel`

salty-ivy:
# Description

By instrumenting  the `SQLPanel` in `sync_to_async` we can monkey patch all the connection in a dedicated thread and store those connections there meanwhile all cross compatible views and operations in async context will take place in that threaded thus they will use those patched db connections which will help us store all async and sync operations.

more explanation of `sync_to_async`: https://docs.djangoproject.com/en/4.2/topics/async/#sync-to-async

Fixes https://github.com/jazzband/django-debug-toolbar/issues/1430
Fixes  https://github.com/jazzband/django-debug-toolbar/issues/932

# Checklist:

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -6,7 +6,7 @@ import re
 import socket
 from functools import lru_cache
 
-from asgiref.sync import iscoroutinefunction, markcoroutinefunction
+from asgiref.sync import iscoroutinefunction, markcoroutinefunction, sync_to_async
 from django.conf import settings
 from django.utils.module_loading import import_string
 
@@ -112,7 +112,10 @@ class DebugToolbarMiddleware:
 
         # Activate instrumentation ie. monkey-patch.
         for panel in toolbar.enabled_panels:
-            panel.enable_instrumentation()
+            if panel.panel_id == "SQLPanel":
+                await sync_to_async(panel.enable_instrumentation)()
+            else:
+                panel.enable_instrumentation()
         try:
             # Run panels like Django middleware.
             response = await toolbar.process_request(request)
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -6,7 +6,7 @@ import re
 import socket
 from functools import lru_cache
 
-from asgiref.sync import iscoroutinefunction, markcoroutinefunction, sync_to_async
+from asgiref.sync import iscoroutinefunction, markcoroutinefunction
 from django.conf import settings
 from django.utils.module_loading import import_string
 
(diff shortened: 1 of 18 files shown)


salty-ivy:
@tim-schilling  I think this is ready as per the discussion tests have been covered by #1835

debug_toolbar/middleware.py
```diff
@@ -112,7 +112,10 @@ async def __acall__(self, request):
 
         # Activate instrumentation ie. monkey-patch.
         for panel in toolbar.enabled_panels:
-            panel.enable_instrumentation()
+            if panel.panel_id == "SQLPanel":
+                await sync_to_async(panel.enable_instrumentation)()
```
tim-schilling, line 116: Is it possible to do this within the `SQLPanel.enable_instrumentation` itself?

tim-schilling: It's a bit odd to see specific logic for a single panel in the middleware. It'll be easier to understand the middleware if we don't have specific panel logic within it. It's also easier to explain why we do that in the panel and have it be in the place a person is most likely to be looking for it.

tim-schilling: Oh, I like that.

salty-ivy: we only have one panel as of now that would utilise this pattern, though it seems really intuitive.

elineda: yep but if a people create a third party panel he need to change the code of the core for make work. That's not a great thing.

docs/spelling_wordlist.txt
```diff
@@ -6,7 +6,9 @@ Pympler
 Roboto
 Transifex
 Werkzeug
+aenable_instrumentation
```
tim-schilling, line 9:
```suggestion
aenable
```

docs/spelling_wordlist.txt
```diff
@@ -15,6 +17,7 @@ checkbox
 contrib
 dicts
 django
+enable_instrumentation
```
tim-schilling, line 20:
```suggestion
```

debug_toolbar/panels/__init__.py
```diff
@@ -154,6 +154,9 @@ def enable_instrumentation(self):
 
         Unless the toolbar or this panel is disabled, this method will be
         called early in ``DebugToolbarMiddleware``. It should be idempotent.
+
+        Add aenable_instrumentation method to the panel class to support async
+        instrumentation logic.
```
elineda, line 159: maybe an "if needed" ? it's not mandatory for async to have a aenable ?

salty-ivy:
@tim-schilling  Apparently `instrumentation` is causing the lint fail, even if I have added it to word list.

debug_toolbar/panels/__init__.py
salty-ivy: yup, although `async instrumentation logic` itself means that `await` is required in that case `aenable_instrumentation` would be the only choice?

tim-schilling (approved):

PR #2001, opened 2024-09-02
Async integration tests

salty-ivy:
#### Description

integrations tests to test the async toolbar.

Fixes # (issue)

#### Checklist:

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -43,6 +43,8 @@ class AsyncToolbarTestClient(AsyncClient):
     async def request(self, **request):
         # Use a thread/async task context-local variable to guard against a
         # concurrent _created signal from a different thread/task.
+        # In cases testsuite will have both regular and async tests or
+        # multiple async tests running in an eventloop making async_client calls.
         data_contextvar.set(None)
 
         def handle_toolbar_created(sender, toolbar=None, **kwargs):
--- a/tests/base.py
+++ b/tests/base.py
@@ -1,12 +1,22 @@
+import contextvars
 from typing import Optional
 
 import html5lib
 from asgiref.local import Local
 from django.http import HttpResponse
-from django.test import Client, RequestFactory, TestCase, TransactionTestCase
+from django.test import (
+    AsyncClient,
+    AsyncRequestFactory,
+    Client,
+    RequestFactory,
+    TestCase,
+    TransactionTestCase,
+)
 
 from debug_toolbar.panels import Panel
 from debug_toolbar.toolbar import DebugToolbar
+
+data_contextvar = contextvars.ContextVar("djdt_toolbar_test_client")
 
 
 class ToolbarTestClient(Client):
@@ -29,11 +39,33 @@ class ToolbarTestClient(Client):
         return response
 
 
+class AsyncToolbarTestClient(AsyncClient):
+    async def request(self, **request):
+        # Use a thread/async task context-local variable to guard against a
+        # concurrent _created signal from a different thread/task.
+        data_contextvar.set(None)
+
+        def handle_toolbar_created(sender, toolbar=None, **kwargs):
+            data_contextvar.set(toolbar)
+
+        DebugToolbar._created.connect(handle_toolbar_created)
+        try:
+            response = await super().request(**request)
+        finally:
+            DebugToolbar._created.disconnect(handle_toolbar_created)
+        response.toolbar = data_contextvar.get()
+
+        return response
+
+
 rf = RequestFactory()
+arf = AsyncRequestFactory()
 
 
 class BaseMixin:
+    _is_async = False
     client_class = ToolbarTestClient
+    async_client_class = AsyncToolbarTestClient
 
     panel: Optional[Panel] = None
     panel_id = None
@@ -42,7 +74,11 @@ class BaseMixin:
         super().setUp()
         self._get_response = lambda request: HttpResponse()
         self.request = rf.get("/")
-        self.toolbar = DebugToolbar(self.request, self.get_response)
+        if self._is_async:
+            self.request = arf.get("/")
+            self.toolbar = DebugToolbar(self.request, self.get_response_async)
+        else:
+            self.toolbar = DebugToolbar(self.request, self.get_response)
         self.toolbar.stats = {}
 
         if self.panel_id:
@@ -57,6 +93,9 @@ class BaseMixin:
         super().tearDown()
 
     def get_response(self, request):
+        return self._get_response(request)
+
+    async def get_response_async(self, request):
         return self._get_response(request)
 
     def assertValidHTML(self, content):
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -11,7 +11,7 @@ from django.core.cache import cache
 from django.db import connection
 from django.http import HttpResponse
 from django.template.loader import get_template
-from django.test import RequestFactory
+from django.test import AsyncRequestFactory, RequestFactory
 from django.test.utils import override_settings
 
 from debug_toolbar.forms import SignedDataForm
@@ -125,6 +125,22 @@ class DebugToolbarTestCase(BaseTestCase):
 
         request.META.pop("wsgi.multiprocess")
         self.assertTrue(toolbar.should_render_panels())
+
+    def test_should_render_panels_asgi(self):
+        """
+        The toolbar not should render the panels on each request when wsgi.multiprocess
+        is True or missing in case of async context rather than multithreaded
+        wsgi.
+        """
+        async_request = AsyncRequestFactory().get("/")
+        # by default ASGIRequest will have wsgi.multiprocess set to True
+        # but we are still assigning this to true cause this could change
+        # and we specifically need to check that method returns false even with
+        # wsgi.multiprocess set to true
+        async_request.META["wsgi.multiprocess"] = True
+        toolbar = DebugToolbar(async_request, self.get_response)
+        toolbar.config["RENDER_PANELS"] = None
+        self.assertFalse(toolbar.should_render_panels())
 
     def _resolve_stats(self, path):
         # takes stats from Request panel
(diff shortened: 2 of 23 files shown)


tests/base.py
```diff
@@ -29,11 +39,33 @@ def handle_toolbar_created(sender, toolbar=None, **kwargs):
         return response
 
 
+class AsyncToolbarTestClient(AsyncClient):
+    async def request(self, **request):
+        # Use a thread/async task context-local variable to guard against a
+        # concurrent _created signal from a different thread/task.
```
matthiask, line 45: Can you expand the comment to explain when this happens? E.g. when running the test suite using multiple threads or something?

tests/views.py
```diff
@@ -15,7 +15,15 @@ def execute_sql(request):
 
 
 async def async_execute_sql(request):
-    await sync_to_async(list)(User.objects.all())
+    list_stor = []
```
matthiask, line 18:
```suggestion
    list_store = []
```

Probably?

tests/test_integration_async.py
```diff
@@ -0,0 +1,617 @@
...
+    def content(self):
+        raise Exception
+
+
+@override_settings(DEBUG=True)
+class DebugToolbarTestCase(BaseTestCase):
+    _is_async = True
+
+    async def test_show_toolbar(self):
+        self.assertTrue(show_toolbar(self.request))
```
matthiask, line 48: `show_toolbar` doesn't have any async behavior, so I'm not sure why it should be tested?

tests/test_integration_async.py
```diff
@@ -0,0 +1,617 @@
...
+    #     stats = self._resolve_stats("/resolving3/a/")
+    #     self.assertEqual(stats["view_args"], ("a",))
+    #     self.assertEqual(stats["view_kwargs"], {"arg2": "default"})
+
+    # def test_url_resolving_bad(self):
+    #     stats = self._resolve_stats("/non-existing-url/")
+    #     self.assertEqual(stats["view_urlname"], "None")
+    #     self.assertEqual(stats["view_args"], "None")
+    #     self.assertEqual(stats["view_kwargs"], "None")
+    #     self.assertEqual(stats["view_func"], "<no view>")
```
matthiask, line 148: I assume you want to remove all of that? Then do it :)

tests/test_integration_async.py
```diff
@@ -0,0 +1,617 @@
...
+        # may run earlier and cause fewer cache calls.
+        cache.clear()
+        response = await self.async_client.get("/cached_view/")
+        self.assertEqual(len(response.toolbar.get_panel_by_id("CachePanel").calls), 3)
+        response = await self.async_client.get("/cached_view/")
+        self.assertEqual(len(response.toolbar.get_panel_by_id("CachePanel").calls), 2)
+
+    @override_settings(ROOT_URLCONF="tests.urls_use_package_urls")
+    async def test_include_package_urls(self):
+        """Test urlsconf that uses the debug_toolbar.urls in the include call"""
```
matthiask, line 178: Maybe that comment was already there, but I don't understand it.

tests/test_integration_async.py
```diff
@@ -0,0 +1,617 @@
...
+
+# Concurrent database queries are not fully supported by Django's backend with
+# psycopg2 and psycopg3's async support isn't integrated yet. As a result, in
+# ASGI environments(especially with the Django Debug Toolbar), either
+# deadlocks can occure in certain cases like when running
+# tests/views/async_execute_sql_concurrently in ASGI environment
+# or queries will execute synchronously. Check ou
+# https://forum.djangoproject.com/t/are-concurrent-database-queries-in-asgi-a-thing/24136/2
+# https://forum.djangoproject.com/t/are-concurrent-database-queries-in-asgi-a-thing/24136/2
+# https://github.com/django/deps/blob/main/accepted/0009-async.rst#the-orm
```
matthiask, line 276: All of that is of course true, but it sounds as if we only looked at PostgreSQL and not at sqlite, mysql etc.

Either mention that the PostgreSQL support is used as an example, or research what happens with other database engines. (The former is fine for me!)

tests/test_integration_async.py
```diff
@@ -0,0 +1,617 @@
...
+                    "alias": "default",
+                    "duration": "0",
+                }
+            )
+        }
+
+        response = await self.async_client.post(url, data)
+        self.assertEqual(response.status_code, 200)
+        response = await self.async_client.post(
+            url, data, headers={"x-requested-with": "XMLHttpRequest"}
```
matthiask, line 433: By the way, the `x-requested-with` header is not needed anymore. Django has removed `request.is_ajax()` and we do not depend on it either.

matthiask (commented):
I read the Django docs regarding `async_client` and now it seems to me that the tests should do what I think they do. 

I really wish the reviewing interface would show what actually changed compared to the sync integration tests. Maybe we could copy the file in a separate pull request and start changing it later the next time we do something similar so that the reviewing interface is more helpful.

tests/test_integration_async.py
salty-ivy: yup I just kept it in order for easy review purpose to let reviewers see what's been involved and what's not.

tests/test_integration_async.py
salty-ivy: Originally not truly an async behaviour but `self.request` what would be passed into it will become an `ASGIRequest` in async context that check can be good to have in any case ? somewhat future proofing as well? cause some changes in show toolbar could be of `WSGIRequest` and in those cases it can be useful.

matthiask: Ah well, that's true. 

Maybe remove the `async` though since there's no `await`-ing happening, and add a docstring:

```
"""show_toolbar() works with an ASGIRequest too"""
```

Or something like this?

tests/test_integration_async.py
salty-ivy: Should I remove them from both integration tests?

tests/test_integration.py
```diff
@@ -288,7 +288,7 @@ def test_sql_page(self):
     def test_async_sql_page(self):
         response = self.client.get("/async_execute_sql/")
         self.assertEqual(
-            len(response.toolbar.get_panel_by_id("SQLPanel").get_stats()["queries"]), 1
+            len(response.toolbar.get_panel_by_id("SQLPanel").get_stats()["queries"]), 2
```
tim-schilling, line 291: Why did this change?

tim-schilling: Whoops, should have read the whole PR first.

tests/test_integration_async.py
matthiask: Remove them from the new tests for sure, but maybe a different pull request for the existing tests to keep this PR small(er). Thanks!

tests/base.py
matthiask: Ah I thought so. In a way outdated or confusing comments are worse than no comments at all. It would certainly help a lot if you could clarify this!

tim-schilling (changes requested):
It's looking good! I've got a few requests for changes, but some of them are discussions.

tests/test_integration_async.py
```diff
@@ -0,0 +1,560 @@
...
+
+from debug_toolbar.forms import SignedDataForm
+from debug_toolbar.middleware import DebugToolbarMiddleware, show_toolbar
+from debug_toolbar.panels import Panel
+from debug_toolbar.toolbar import DebugToolbar
+
+from .base import BaseTestCase, IntegrationTestCase
+from .views import regular_view
+
+rf = RequestFactory()
```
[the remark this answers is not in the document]
salty-ivy: technically yes, in case if we counter a situation where we need both but yea should remove this.

tests/test_integration_async.py
```diff
@@ -0,0 +1,560 @@
...
+
+
+# Concurrent database queries are not fully supported by Django's backend with
+# current integrated database drivers like psycopg2
+# (considering postgresql as an example) and
+# support for async drivers like psycopg3 isn't integrated yet.
+# As a result, in ASGI environments(especially with the Django Debug Toolbar),
+# either deadlocks can occure in certain cases like when running
+# tests/views/async_execute_sql_concurrently in ASGI environment
+# or queries will execute synchronously. Check out
```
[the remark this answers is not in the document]
salty-ivy: I did some more testing and I think the deadlock situation was quite deadlock was more a particular case that occurred in that issues, main thing is query will run synchronously.

I am editing the comment

tests/test_integration_async.py
```diff
@@ -0,0 +1,560 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "djDebug")
+
+    @override_settings(DEFAULT_CHARSET="iso-8859-1")
+    async def test_non_utf8_charset(self):
+        response = await self.async_client.get("/regular/ASCII/")
+        self.assertContains(response, "ASCII")  # template
+        self.assertContains(response, "djDebug")  # toolbar
+
+        response = self.client.get("/regular/LÀTÍN/")
```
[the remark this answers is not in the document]
salty-ivy: Ah, I missed it, though it seems to fail at `async_client` ; )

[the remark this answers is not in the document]
salty-ivy: I am not sure why assert is failing cause manually it seems to work. for now I am changing `LATIN` to `ASCII` for this too.

matthiask (approved):
I think this is a good change and we should merge it, or do you have any reservations @tim-schilling ?

tim-schilling:
@matthiask I didn't review it thoroughly, but it looks like test changes. I'm good with merging them if you're happy. I'll be less busy in ~2 weeks

tim-schilling (approved):
Approval based on Matthias' approval :grin: (I skimmed over the changes quickly)

matthiask:
Let's do this then :) Thanks!

PR #2005, opened 2024-09-17
add support for LoginRequiredMiddleware with login_not_required decor…

dulmandakh:
#### Description

This PR adds support for LoginRequiredMiddleware introduced in Django 5.1, or views provided by this package don't require login.

#### Checklist:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,7 @@ Pending
   `Google Summer of Code Project 2024
   <https://summerofcode.withgoogle.com/programs/2024/projects/iXVvyGYp>`__.
 * Added Django 5.1 to the CI matrix.
+* Added support for the ``LoginRequiredMiddleware`` introduced in Django 5.1.
 * Support select and explain buttons for ``UNION`` queries on PostgreSQL.
 * Fixed internal toolbar requests being instrumented if the Django setting
   ``FORCE_SCRIPT_NAME`` was set.
--- a/tests/test_login_not_required.py
+++ b/tests/test_login_not_required.py
@@ -1,0 +1,39 @@
+import unittest
+
+import django
+from django.test import SimpleTestCase, override_settings
+from django.urls import reverse
+
+
+@unittest.skipIf(
+    django.VERSION < (5, 1),
+    "Valid on Django 5.1 and above, requires LoginRequiredMiddleware",
+)
+@override_settings(
+    DEBUG=True,
+    MIDDLEWARE=[
+        "django.contrib.sessions.middleware.SessionMiddleware",
+        "django.contrib.auth.middleware.AuthenticationMiddleware",
+        "django.contrib.auth.middleware.LoginRequiredMiddleware",
+        "debug_toolbar.middleware.DebugToolbarMiddleware",
+    ],
+)
+class LoginNotRequiredTestCase(SimpleTestCase):
+    def test_panels(self):
+        for uri in (
+            "history_sidebar",
+            "history_refresh",
+            "sql_select",
+            "sql_explain",
+            "sql_profile",
+            "template_source",
+        ):
+            with self.subTest(uri=uri):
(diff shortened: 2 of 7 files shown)


tests/urls.py
```diff
@@ -26,5 +26,5 @@
     path("ajax/", views.ajax_view),
     path("login_without_redirect/", LoginView.as_view(redirect_field_name=None)),
     path("admin/", admin.site.urls),
-    path("__debug__/", include("debug_toolbar.urls")),
+    path("__debug__/", include("debug_toolbar.urls", namespace="djdt")),
```
matthiask, line 29: What's the reason for this change?

debug_toolbar/decorators.py
```diff
@@ -5,9 +5,17 @@
...
+    from django.contrib.auth.decorators import login_not_required
+except ImportError:
+
+    def login_not_required(func):
+        return func
+
 
 def require_show_toolbar(view):
     @functools.wraps(view)
+    @login_not_required
```
matthiask, line 18: I don't think `require_show_toolbar` necessarily implies `login_not_required`. In practice it does, but I'd prefer if the decorator was added to the view function themselves, even though you have to scatter the `@login_not_required` invocations over the code base. 

I have added `authlib/_compat.py` (with a leading underscore to signify that this is purely internal and shouldn't be relied on), the same thing may make sense here.

docs/changes.rst
```diff
@@ -11,6 +11,7 @@ Pending
   `Google Summer of Code Project 2024
   <https://summerofcode.withgoogle.com/programs/2024/projects/iXVvyGYp>`__.
 * Added Django 5.1 to the CI matrix.
+* Added support for LoginRequiredMiddleware introduced in Django 5.1.
```
matthiask, line 14:
```suggestion
* Added support for the ``LoginRequiredMiddleware`` introduced in Django 5.1.
```

tests/test_login_not_required.py
```diff
@@ -0,0 +1,31 @@
...
+    def test_login_not_required(self):
+        self.assertEqual(self.client.get(reverse("djdt:render_panel")).status_code, 200)
+
+
+# DEBUG_TOOLBAR_CONFIG = {
+#     # Django's test client sets wsgi.multiprocess to True inappropriately
+#     "RENDER_PANELS": False,
+#     # IS_RUNNING_TESTS must be False even though we're running tests because we're running the toolbar's own tests.
+#     "IS_RUNNING_TESTS": False,
+# }
```
matthiask, line 31:
```suggestion
```

matthiask (commented):
Thanks for working on this!

Here's an example of an app adding support for `login_not_required` while actually testing the behavior:
https://github.com/matthiask/django-authlib/commit/2239a1c1d313e1f5a8db14531d0458fbc19f5c6e

tests/urls.py
dulmandakh: just to use reverse in tests

dulmandakh:
> Thanks for working on this!
> 
> Here's an example of an app adding support for `login_not_required` while actually testing the behavior: [matthiask/django-authlib@2239a1c](https://github.com/matthiask/django-authlib/commit/2239a1c1d313e1f5a8db14531d0458fbc19f5c6e)

I contributed same support for Django OAuth Toolkit in https://github.com/jazzband/django-oauth-toolkit/pull/1454. Thanks for the reference.

debug_toolbar/decorators.py
dulmandakh: changed the implementation as suggested

tests/test_login_not_required.py
dulmandakh: removed comments

dulmandakh:
@matthiask please review and merge if this PR looks good

matthiask:
Thanks!

PR #2014, opened 2024-10-09
Adding in support for Python 3.13

abeed-avayu:
#### Description

This adds support for Python 3.13. I _think_ I've got all the references, but there's probably somewhere else that I've overlooked.

Shout-out to London Django Meetup for the Hacktoberfest Hackathon inspiration. CC @adamchainz et al.

#### Testing notes

I ran this under docker (python:3.13-slim) locally and `make test` and `make coverage` passed all the tests. Would have also tried with a virtualenv but pyenv didn't have 3.13 (non-RC) at the time of testing.

#### Checklist:

- [ x ] I have added the relevant tests for this change.
- [ x ] I have added an item to the Pending section of ``docs/changes.rst``.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -84,8 +84,7 @@ jobs:
             - python-version: '3.12'
               database: psycopg3
             - python-version: '3.13'
-              # psycopg installs version 3+; psycopg3 not found via pip in python 3.13
-              database: psycopg
+              database: psycopg3
 
     services:
       postgres:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
 
     services:
       mariadb:
@@ -73,7 +73,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
         database: [postgresql, postgis]
         # Add psycopg3 to our matrix for modern python versions
         include:
@@ -148,7 +148,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
 
     steps:
     - uses: actions/checkout@v4
@@ -195,7 +195,7 @@ jobs:
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
       with:
-        python-version: 3.8
+        python-version: 3.9
 
     - name: Get pip cache dir
       id: pip-cache
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,7 +73,8 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        # Skip 3.13 here, it needs the psycopg3 / postgis3 database
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         database: [postgresql, postgis]
         # Add psycopg3 to our matrix for modern python versions
         include:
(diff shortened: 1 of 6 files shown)


docs/changes.rst
```diff
@@ -4,6 +4,8 @@ Change log
 Pending
 -------
 
+* Added support for Python 3.13 and to the CI matrix.
```
matthiask, line 7: Typo?

```suggestion
* Added support for Python 3.13 to the CI matrix.
```

tox.ini
```diff
@@ -3,8 +3,8 @@ isolated_build = true
 envlist =
     docs
     packaging
-    py{38,39,310,311,312}-dj{42}-{sqlite,postgresql,postgis,mysql}
-    py{310,311,312}-dj{42,50,51,main}-{sqlite,postgresql,psycopg3,postgis,mysql}
+    py{38,39,310,311,312,313}-dj{42}-{sqlite,postgresql,postgis,mysql}
+    py{310,311,312,313}-dj{42,50,51,main}-{sqlite,postgresql,psycopg3,postgis,mysql}
```
matthiask, line 7: No version of Django before 5.1.3 support Python 3.13 if I interpret this recent change correctly:

https://github.com/django/django/commit/2e3bc59fd3760de87952ec8fd6cd3694e8d9dc1c

So, 313 should really only be added to dj51 (once 5.1.3 is actually released) and djmain for now.

matthiask:
Thanks! The failures are probably due to psycopg2 not having 3.13-compatible wheels yet. (https://github.com/psycopg/psycopg2/pull/1729)

I added some comments.

tox.ini
matthiask: https://forum.djangoproject.com/t/backport-python-3-13-support-in-django-5-0/34671 better source.

abeed-avayu: I've separated out 3.13 to match up with expected Django compatibility.

abeed-avayu:
I've probably overstepped the mark by removing 3.8 as it's end of life, but as the next version release looks to be a major one I thought it would be compatible with that. Let me know if that's a problem and I'll reinstate it.

matthiask:
> I've probably overstepped the mark by removing 3.8 as it's end of life, but as the next version release looks to be a major one I thought it would be compatible with that. Let me know if that's a problem and I'll reinstate it.

On a pedantic day I'd like this to happen with a separate pull request. I think supporting 5 minor Python versions at the same time is enough and I'm therefore fine with it. CI isn't free energy wise after all.

docs/changes.rst
```diff
@@ -4,6 +4,9 @@ Change log
 Pending
 -------
 
+* Added support for Python 3.13 to the CI matrix.
+* Removed support for Python 3.8 as it is end of life.
```
matthiask, line 8: Another suggestion. The first line is optional, I feel more strongly about the second line.

```suggestion
* Added Python 3.13 to the CI matrix.
* Removed support for Python 3.8 as it has reached end of life.
```

tox.ini
```diff
@@ -49,28 +50,28 @@ pip_pre = True
 commands = python -b -W always -m coverage run -m django test -v2 {posargs:tests}
 
 
-[testenv:py{38,39,310,311,312}-dj{42,50,51,main}-{postgresql,psycopg3}]
+[testenv:py{38,39,310,311,312,313}-dj{42,50,51,main}-{postgresql,psycopg3}]
```
matthiask, line 53: You can remove 38 here and below as well.

.github/workflows/test.yml
```diff
@@ -83,6 +83,9 @@ jobs:
               database: psycopg3
             - python-version: '3.12'
               database: psycopg3
+            - python-version: '3.13'
+              # psycopg installs version 3+; psycopg3 not found via pip in python 3.13
+              database: psycopg
```
matthiask, line 88: Can you explain why you're using `psycopg` instead of `psycopg3` here and below? This doesn't look correct to me. psycopg3 doesn't refer to the PyPI package, it's just an internal name we're using to differentiate between the old psycopg2 and the new psycopg v3; the package name itself is defined here: https://github.com/jazzband/django-debug-toolbar/blob/244d4fd7f94caf77c4c18058420fa03f21751195/tox.ini#L16

abeed-avayu: Yeah, I got that wrong, sorry.

docs/changes.rst
abeed-avayu: Added.

tox.ini
abeed-avayu: So still struggling a bit to get the matrix running correctly. I think we only want py313 with dj51 and main, and then postgres and postgis DB backends but using psycopg as the adapter library.

matthiask:
Thank you for your contribution!

PR #2017, opened 2024-10-16
Convert to Django Commons pypi-github release process

tim-schilling:
#### Description

This switches us to the Django Commons release process.

References: https://github.com/django-commons/membership/issues/34
References: #1918

#### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,40 +1,120 @@
-name: Release
+name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
 
-on:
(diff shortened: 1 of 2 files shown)


matthiask (approved):
Is there a reason why we have both TestPyPI *and* PyPI in there?

I'm not using trusted publishing yet (unfortunately), so I cannot say if this would work, but it certainly looks good to me.

tim-schilling:
This comes from the Django Commons example and is from the pypi tutorial. I'm about 99% sure it'll work for us.

PR #2021, opened 2024-10-27
The static files panel shouldn't choke on unexpected data types

### LINKED ISSUE #2002: TypeError __str__ returned non-string (type PosixPath) /debug_toolbar/templates/debug_toolbar/panels/staticfiles.html
django-debug-toolbar: 4.4.6

I'm just providing a solution to an event I came across. Line 29 for /debug_toolbar/templates/debug_toolbar/panels/staticfiles.html needs to have {{ staticfile }} updated to {{ staticfile|stringformat:"s" }}

Below is the event that happened.

When running uvicorn project.asgi:application --host 127.0.0.1 --port 8001 and then loading 127.0.0.1:8001 in a new web browser, the TypeError __str__ returned non-string (type PosixPath). Specifically focusing on {{ staticfile }}. I updated {{ staticfile }} to {{ staticfile|stringformat:"s" }} and that resolved the TypeError.

Thanks!

matthiask:
#### Description

Using `Path()` instances in `static()` is unexpected but we shouldn't crash. 

While debugging I found that our `DebugStaticFilesStorage.url` instantiates a `StaticFile` which in turn calls `DebugStaticFilesStorage.url` again, so rendering `self.panel.content` in my test wouldn't end ever. What was tricky about it is that I didn't get an infinite recursion error because we're using a signal. (To be clear: I'm not sure this is what happens, I only think it is.)

I'm confused why this wouldn't have appeared earlier, but I'm happy with the fix.

Fixes #2002 

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -21,7 +21,7 @@ class StaticFile:
         self.path = path
 
     def __str__(self):
-        return self.path
+        return str(self.path)
 
     def real_path(self):
         return finders.find(self.path)
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -17,17 +17,18 @@ class StaticFile:
     Representing the different properties of a static file.
     """
 
-    def __init__(self, path):
+    def __init__(self, *, path, url):
         self.path = path
+        self._url = url
 
     def __str__(self):
-        return str(self.path)
+        return self.path
 
     def real_path(self):
         return finders.find(self.path)
 
     def url(self):
-        return storage.staticfiles_storage.url(self.path)
+        return self._url
 
 
 # This will record and map the StaticFile instances with its associated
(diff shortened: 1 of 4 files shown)


tim-schilling (approved):
The code makes sense. I don't have ideas on why this would break either but I'm happy to see a problem solved

salty-ivy:
>  What was tricky about it is that I didn't get an infinite recursion error because we're using a signal. (To be clear: I'm not sure this is what happens, I only think it is.)

If that were to be the case,  wouldn't we will have numerous signals for a single request, if call `staticfile.url()`

salty-ivy:
> What was tricky about it is that I didn't get an infinite recursion error because we're using a signal. 

Just checked.

we can still hit the `maximum recursion depth` if we make a call to `StaticFile(path).url()` within `DebugStaticFilesStorage.url()`

I think that concludes that `signal` is not the reason and `staticfile.url()` is not being called within `DebugStaticFilesStorage.url()`  in any typical case.

salty-ivy (approved):

PR #2025, opened 2024-10-31
Apply UP006 and UP035 changes from pyupgrade pre-commit hook

ulgens:
UP006: https://docs.astral.sh/ruff/rules/non-pep585-annotation/
UP035: https://docs.astral.sh/ruff/rules/deprecated-import/

#### Description

`pre-commit` output on the main branch wasn't clean. Run `pre-commit run --all-files` and applied the resulting fixes.

The ideal solution here would be to have `pre-commit` running in the CI. I could any recommendations if there is any easy way to achieve that with `tox`

### DIFF
--- a/debug_toolbar/_stubs.py
+++ b/debug_toolbar/_stubs.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, List, NamedTuple, Optional, Tuple
+from typing import Any, NamedTuple, Optional
 
 from django import template as dj_template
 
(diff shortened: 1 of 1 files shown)


matthiask:
Makes sense, thanks.

I thought the pre-commit CI runs, but it obviously does not!

matthiask:
I reconnected pre-commit.ci with the repository since the CI thing was lost when moving away from Jazzband.

ulgens:
@matthiask Thanks for the fix. How can we see the build output for pre-commit hooks, or at least a log that shows it's working?

matthiask:
@ulgens GitHub doesn't seem to link to the output anymore after merging, but there you go:

https://results.pre-commit.ci/run/github/46939/1730393311.cldDtEpORou_PEjJom970A
https://github.com/django-commons/django-debug-toolbar/pull/2026

ulgens:
@matthiask Thanks for the link.

I don't use the external service at all, so I have no idea how to fix these issues but please let me know if it's something I can help.

PR #2040, opened 2024-12-14
Adopt a basic security policy

tim-schilling:
This informs users to submit security reports through GitHub's private vulnerability mechanism.

#### Description

This formalizes a basic security policy for the project. I'm unsure about the supported version aspect. I think it's factual. If there's a bug in an older version that's not in the newer, we're not backporting it. However, we have done this in the past with a significant vulnerability.

#### Checklist:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,9 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


matthiask (approved):
Excellent, thank you!

elineda (approved):
As we have no LTS we aren't bound to update older version. So it's ok to do only  lastest..

ok for me. ty

PR #2047, opened 2024-12-30
Updated Troubleshooting documentation: simpler mimetype workaround for .js file

### LINKED ISSUE #2046: Documentation: simplify troubleshooting of incorrect mime types for .js static files
The troubleshooting documentation makes it difficult for new Django developers like me who follow the Django tutorial and run into an issue:
https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#troubleshooting

- First, it should tell you to check the console window for message about incorrect MIME type when loading toolbar.js (was text/plain for me, running on Windows O.S.). For me the error message was (sorry, in French):
`Le chargement du module à l’adresse « http://localhost:8000/static/debug_toolbar/js/toolbar.js » a été bloqué en raison d’un type MIME interdit (« text/plain »).`
- Second, the workarounds are complexe, involve global system configuration changes, are missing detailed steps and need to be performed on each developer's machine. 

A much simpler work-around is to add the following to your settings.py, which [forces the MIME type of .js files](https://docs.python.org/3/library/mimetypes.html#mimetypes.add_type):
```
import mimetypes
mimetypes.add_type("application/javascript", ".js", True)
```

This resolved the issue for me without system-wide change. 
May even want to add these steps as part of the installation instruction, to ensure it runs on any system.

blep:
#### Description

Updated Troubleshooting documentation:

- Added an introduction advising users to check the developer console log for errors.
- Provided an example of an error message in the developer console log for an incorrect MIME type for a .js file.
- Provided a simple solution to fix the incorrect MIME type for a .js file that does not require an O.S. level change.
- Restructured the existing MIME type documentation to improve readability.

Fixes #2046 (issue)

#### Checklist:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -9,6 +9,7 @@ Pending
 * Converted to Django Commons PyPI release process.
 * Fixed a crash which occurred when using non-``str`` static file values.
 * Documented experimental async support.
+* Improved troubleshooting doc for incorrect mime types for .js static files
 
 5.0.0-alpha (2024-09-01)
 ------------------------
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -195,15 +195,41 @@ option.
 Troubleshooting
 ---------------
 
-On some platforms, the Django ``runserver`` command may use incorrect content
-types for static assets. To guess content types, Django relies on the
-:mod:`mimetypes` module from the Python standard library, which itself relies
-on the underlying platform's map files. If you find improper content types for
-certain files, it is most likely that the platform's map files are incorrect or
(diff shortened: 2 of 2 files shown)


tim-schilling (approved):
I'm not a big fan of putting the python workaround before the more correct, fix your OS configuration, solution, but I can live with it. It suggests that's our recommended approach and if it is, we should change how the library works. I'm unsure if we want to do that, so that's why I'm not sold on the order.

That said, the rest of the changes are excellent, so I'm overall really happy to have this in.

tim-schilling:
Rebased, then squashed and merged. Thank you @blep!

PR #2049, opened 2025-01-03
Track and display middleware time in the toolbar

### LINKED ISSUE #1903: CPU Time includes DJDT's own time
The CPU Time displayed in the DJDT shows the total time taken to render the request, including the time to render the DJDT itself. This is not what I care about, because the DJDT will not be there when running in production. 

Is it possible to get the time elapsed of the actual rendered view on-screen somewhere?

earthcomfy:
#### Description

CPU time includes the debug toolbar's own time so this PR aims to track how long the middleware takes to run and display that time on the panel.

Fixes #1903

#### Checklist:


robhudson:
Would it make sense to include a stat with the middleware time deducted from the total for the user? Or deduct it from the existing total?

tim-schilling:
@robhudson the thinking I had earlier was that the toolbar changes the application [so we shouldn't try to extract the toolbar itself from the timing](https://github.com/django-commons/django-debug-toolbar/issues/1903#issuecomment-2047100405).

debug_toolbar/middleware.py
```diff
@@ -78,6 +79,8 @@ def __init__(self, get_response):
             markcoroutinefunction(self)
 
     def __call__(self, request):
+        # Start tracking middleware execution time
+        middleware_start_time = perf_counter()
```
tim-schilling, line 83: We probably want to rename this to `toolbar_` and update the text for it to be specific to the toolbar. Middleware may imply all of the middlewares together.

earthcomfy: Yes, that makes sense. I have renamed it.

debug_toolbar/middleware.py
```diff
@@ -78,6 +79,8 @@ def __init__(self, get_response):
             markcoroutinefunction(self)
 
     def __call__(self, request):
+        # Start tracking middleware execution time
+        toolbar_start_time = perf_counter()
```
tim-schilling, line 83: Sorry, I didn't go through this well enough earlier. What do you think about putting this start time in `TimerPanel.enable_instrumentation`? That way we put all of the timing logic in that specific panel and we avoid leaking it into the toolbar itself.

debug_toolbar/middleware.py
```diff
@@ -98,10 +101,14 @@ def __call__(self, request):
             # regardless of the response. Keep 'return' clauses below.
             for panel in reversed(toolbar.enabled_panels):
                 panel.disable_instrumentation()
+            toolbar_end_time = perf_counter()
```
tim-schilling, line 104: The other thing I was thinking about is how to handle measuring the full time of the toolbar's processing. This is actually stopping before all the `Panel.generate_stats()` methods are called which contains quite a bit of logic. I think ideally we'd capture that in this timing.

The problem is, how do we capture that? I have a few ideas, but I'm not happy with them.

I think we could measure the end time in `TimerPanel.generate_stats()` like we do with the rest and store it. This means any other panels that generate their stats after the `TimerPanel` would not be included. To mitigate that, we could find a way to make sure that `TimerPanel` is first to enable and last to generate stats. I think having it be first in `enabled_panels` would do the trick.

The other idea I had was to create a templatetag that would do the diff between the stored time and the current time at the time the toolbar is rendered. The problem here is that it wouldn't be compatible with the serializable branch where we want to store all the stats before rendering.

So I think trying to consolidate all the logic into `TimerPanel` with a change to force `TimerPanel` to be first for processing (not appearance on the idea) is ideal.

@salty-ivy @elineda any thoughts?

debug_toolbar/middleware.py
earthcomfy: Yes, added it to `aenable_instrumentation` as well

debug_toolbar/middleware.py
earthcomfy: I have updated the `enabled_panels` property to ensure the `TimerPanel` is first in order. This is I think what you meant by in the first approach. The order of enabled panels is reversed in `_postprocess` ensuring the end time is captured at the end.

tim-schilling: Excellent, thank you @earthcomfy! This is looking really nice. The last bit we need are the tests.

earthcomfy: I added a test to ensure the timer panel is first. Are there any other tests that need to be included

debug_toolbar/toolbar.py
```diff
@@ -63,7 +63,15 @@ def enabled_panels(self):
...
-        return [panel for panel in self._panels.values() if panel.enabled]
+        panels = [panel for panel in self._panels.values() if panel.enabled]
+        # Ensure TimerPanel is first in order to measure the full time of the toolbar's processing.
+        timer_panel = next(
+            (panel for panel in panels if panel.panel_id == "TimerPanel"), None
+        )
+        if timer_panel:
+            panels.remove(timer_panel)
+            panels.insert(0, timer_panel)
+        return panels
```
tim-schilling, line 74: @earthcomfy I'm going to need to take a bit to process this. This is a bit of a change and I want to make sure I understand it well. For example, I think this impacts the order of the panels on the side which means it impacts the UX of the toolbar.

Currently by default the Timer panel is 3rd from the top, while the top is the History panel. I think the History panel being at the top is between mildly important and important.

@django-commons/django-debug-toolbar any thoughts here?

debug_toolbar/middleware.py
tim-schilling: @earthcomfy ideally we'd have tests covering all the changes on the Timer panel.

debug_toolbar/toolbar.py
matthiask: Yes, I agree that the history panel should stay at the top because it affects basically everything show below.

Maybe we could unconditionally (without checking if the timer panel is even used) save the start time of the toolbar in the middleware or in the `DebugToolbar` instance. The timer panel could then use `self.toolbar` or something to access this timing information if it's enabled.

tim-schilling: I've been going back and forth on that. I was hoping we could contain this entirely within the `TimerPanel`, but that may be more problematic than we'd like.

matthiask: Yes, it would be great if there was a way to keep it contained. But if the price we pay for that is that we either have to introduce a pre-init API or some way of ordering panels processing (`priority`? `weight`?) then I'd slightly prefer to keep the amount of code required down while uglifying things a bit. This isn't a strongly held conviction though.

tim-schilling: Yeah, that adds a fair amount of complexity. I'm leaning towards agreeing with you. @earthcomfy what do you think?

elineda: The order on the ux is not affected by these changes isn't ?
The display order is calculated by the DebugToolbar.panels (debug_toolbar/toolbar.py:33), not DebugToolbar.enabled_panels (debug_toolbar.toolbar.DebugToolbar.enabled_panels)

tim-schilling: I believe it is. They both end up coming from `settings.get_panels()`

elineda: But the change isn't in the settings, it's ont the panel enabled.

tim-schilling:
@earthcomfy do you have time to pick this back up? If not, @robhudson could ask a Djangonaut from the next session of Djangonaut Space to get it over the finish line.

earthcomfy:
@tim-schilling Thanks for following up! I don’t think I’ll have time to return to this, so feel free to reassign it:)

earthcomfy:
Hey @tim-schilling, I'd like to pick this back up if that's okay.

I checked this again, and I believe what @elineda pointed out is correct. I have verified manually as well that the change doesn't affect the order of the panels. The template ([base.html](https://github.com/django-commons/django-debug-toolbar/blob/main/debug_toolbar/templates/debug_toolbar/base.html#L28)) iterates over `toolbar.panels`, which returns `self._panels.values()` directly (the original order from settings). The `enabled_panels` property creates a separate list for processing order only. doesn't mutate `_panels`.

That said, if the team still prefers the other approach of storing timing on the DebugToolbar instance directly, I'm happy to go that route instead.

PS: fixed the conflict and updated my branch to the latest.

tim-schilling:
Awesome, glad to hear you can work on it again! I've got this on my list of things to review. I don't have a timetable though.

debug_toolbar/panels/timer.py
```diff
@@ -109,6 +113,9 @@ def generate_stats(self, request, response):
             #        stats['urss'] = self._end_rusage.ru_idrss
             #        stats['usrss'] = self._end_rusage.ru_isrss
 
+        if hasattr(self, "_toolbar_start_time"):
+            stats["toolbar_time"] = (perf_counter() - self._toolbar_start_time) * 1000
```
codingjoe, line 117: Consider storing the duration with an explicit type using `datetime.timedelta`. It's always easier than  remembering that `toolbar_time` is in milliseconds.

debug_toolbar/panels/timer.py
```diff
@@ -109,6 +113,9 @@ def generate_stats(self, request, response):
             #        stats['urss'] = self._end_rusage.ru_idrss
             #        stats['usrss'] = self._end_rusage.ru_isrss
 
+        if hasattr(self, "_toolbar_start_time"):
```
codingjoe, line 116: EAFP: Maybe consider catching the AttributeError instead of the default if-branch.

debug_toolbar/panels/timer.py
```diff
@@ -120,7 +127,16 @@ def generate_server_timing(self, request, response):
...
 
     @staticmethod
     def _elapsed_ru(start, end, name):
         return end.get(name) - start.get(name)
+
+    def enable_instrumentation(self):
+        self._toolbar_start_time = perf_counter()
+
+    async def aenable_instrumentation(self):
+        self._toolbar_start_time = perf_counter()
```
codingjoe, line 142: Both are synchronous, and instrumentation usually lives in the CPU and has no IO. Thus, I don't think asyncIO is all that helpful even in the future.

debug_toolbar/toolbar.py
```diff
@@ -73,7 +73,15 @@ def enabled_panels(self) -> list[Panel]:
         """
         Get a list of panels enabled for the current request.
         """
-        return [panel for panel in self._panels.values() if panel.enabled]
+        panels = [panel for panel in self._panels.values() if panel.enabled]
+        # Ensure TimerPanel is first in order to measure the full time of the toolbar's processing.
+        timer_panel = next(
+            (panel for panel in panels if panel.panel_id == "TimerPanel"), None
+        )
+        if timer_panel:
```
codingjoe, line 81: Great place for a walrus:
```suggestion
        if timer_panel := next(
            (panel for panel in panels if panel.panel_id == "TimerPanel"), None
        )
```

debug_toolbar/toolbar.py
```diff
@@ -73,7 +73,15 @@ def enabled_panels(self) -> list[Panel]:
...
+            panels.remove(timer_panel)
+            panels.insert(0, timer_panel)
```
codingjoe, line 83: Maybe replace the item instead of the stack operation.
```suggestion
            panels[0] = timer_panel
```

debug_toolbar/toolbar.py
```diff
@@ -73,7 +73,15 @@ def enabled_panels(self) -> list[Panel]:
         """
         Get a list of panels enabled for the current request.
         """
-        return [panel for panel in self._panels.values() if panel.enabled]
+        panels = [panel for panel in self._panels.values() if panel.enabled]
+        # Ensure TimerPanel is first in order to measure the full time of the toolbar's processing.
+        timer_panel = next(
+            (panel for panel in panels if panel.panel_id == "TimerPanel"), None
+        )
```
codingjoe, line 80: You are looping over the panels twice, O(n)=2*n. I think with a little rearrangement, this can be done in a single loop.

tests/test_integration.py
```diff
@@ -620,6 +624,7 @@ def test_server_timing_headers(self):
             r'TimerPanel_stime;dur=(\d)*(\.(\d)*)?;desc="System CPU time", ',
             r'TimerPanel_total;dur=(\d)*(\.(\d)*)?;desc="Total CPU time", ',
             r'TimerPanel_total_time;dur=(\d)*(\.(\d)*)?;desc="Elapsed time", ',
+            r'TimerPanel_toolbar_time;dur=(\d)*(\.(\d)*)?;desc="Toolbar time", ',
```
codingjoe, line 627: Toolbar is ambiguous outside this repo. Maybe stick with DebugToolbar, like it's currently phrased in most places.
```suggestion
            r'TimerPanel_dj_toolbar_time;dur=(\d)*(\.(\d)*)?;desc="DebugToolbar time", ',
```

debug_toolbar/panels/timer.py
```diff
@@ -47,6 +47,10 @@ def content(self):
             (_("System CPU time"), _("%(stime)0.3f ms") % stats),
             (_("Total CPU time"), _("%(total)0.3f ms") % stats),
             (_("Elapsed time"), _("%(total_time)0.3f ms") % stats),
+            (
+                _("Toolbar time"),
```
codingjoe, line 51: Toolbar is ambiguous outside this repo. Maybe stick with DebugToolbar, like it's currently phrased in most places.
```suggestion
                _("DebugToolbar time"),
```

codingjoe (commented):
Hi @earthcomfy, I invested some time to give this a thorough code review. Please tell me if you want me to clarify some of my comments. Cheers! Joe

debug_toolbar/panels/timer.py
earthcomfy: I could do that, but `generate_stats` already uses `hasattr(self, "_start_time")` on line 86. elsewhere the codebase also uses this pattern so I was following that..

debug_toolbar/panels/timer.py
earthcomfy: Every other timing stat in this file is a millisecond float, and downstream code consumes `ms`. so I think this is a bit out of scope here.

tests/test_integration.py
tim-schilling: :shrug: I don't think it matters a ton really. Especially since it's after "TimerPanel" which the developer needs to understand that the toolbar exists and what it contains.

debug_toolbar/panels/timer.py
earthcomfy: Removed:)

debug_toolbar/toolbar.py
earthcomfy: Refactored this one. The above 2 comments are thus no longer relevant.

tests/test_integration.py
earthcomfy: Happy to revert this back to `TimerPanel`. I think personally `TimerPanel` reads simpler but honestly fine with either.

debug_toolbar/toolbar.py
tim-schilling: You're right @elineda, this isn't an issue. Thank you

tests/test_integration.py
```diff
@@ -620,6 +624,7 @@ def test_server_timing_headers(self):
             r'TimerPanel_stime;dur=(\d)*(\.(\d)*)?;desc="System CPU time", ',
             r'TimerPanel_total;dur=(\d)*(\.(\d)*)?;desc="Total CPU time", ',
             r'TimerPanel_total_time;dur=(\d)*(\.(\d)*)?;desc="Elapsed time", ',
+            r'TimerPanel_toolbar_time;dur=(\d)*(\.(\d)*)?;desc="DebugToolbar time", ',
```
tim-schilling: I'm going to change the desc here to use "Including DebugToolbar time" to be clear it's not just the toolbar. Reviewing this after such a long time I expected this to just be the toolbar for a while 😅

tim-schilling (approved):
Thank you @earthcomfy for sticking with this one and tolerating my slowness!

PR #2051, opened 2025-01-11
added clipboard copy button

### LINKED ISSUE #1840: Make copying context easier (for pasting into an editor or shell).
A lot of the time I'm looking at the context of templates, the next step is top paste it into an editor for reasons[1] - especially including pasting into a python shell or notebook to explore.

At the moment it's inconvenient: You have to manually select the data - making it easy to select things that aren't the context data, and the data isn't exactly in a useable format ready to paste into a python list.

See this example - the first and second items are not comma separated and there is no start and end bracket.:
image

It would be nice to have a "Copy to clipboard" button, perhaps that floats in the top right of the context.
It would also be good if the text was in a format that was likely to work in a python shell, but maybe that is a slightly different issue (at the 

[1] - scrolling up and down in the browser is less convenient with everything there, searching is more fiddly etc

thesayfulla:
#### Description
Added a 'Copy' button for template context with unicode handling.

Fixes #1840

Hi @tim-schilling, please give me your feedback, thank you!


thesayfulla:
> This is a nice start! I think we'd want a few more things before this could be merged.
> 
> * The original issue calls out the fact that the context lists several dictionaries one after another with no delimiter. This PR may need to do something about that. The goal is to have something that can be copied and pasted into a shell (looking at the original issue). Please consider that workflow and let me know what you think.
> * Some selenium tests which may be annoying, but we need something to automatically test this

Got it, I'll make the changes. Thanks for the feedback!

thesayfulla:
> * The original issue calls out the fact that the context lists several dictionaries one after another with no delimiter. This PR may need to do something about that. The goal is to have something that can be copied and pasted into a shell (looking at the original issue). Please consider that workflow and let me know what you think.

What do you think 'Copy' button would wrap these dictionaries inside a list format?

So, instead of copying:
```json
{"data": "smth"}
{"data2": "smth2"}
```

It would copy:
```json
[
    {"data": "smth"},
    {"data2": "smth2"}
]
```

tim-schilling:
I think that's better. The alternative is allowing them to copy each dict one at a time. This would require a small UI change to add some more vertical spacing between the dictionaries.

---

I wonder if we should use a dictionary with keys identifying where each of the contexts have come from.

That's a bit bigger of a change and should be a separate issue.

---

Last aside, some of the values are not valid primitives or common classes. They are things like `"my_class" : <MyClass prop=Hello world other=10 ...>` what do you suggest we do when that's the case? That value can not be pasted into a shell as is.

thesayfulla:
**1. I think that's better. The alternative is allowing them to copy each dict one at a time. This would require a small UI change to add some more vertical spacing between the dictionaries.**

Yes, I agree with you. My next question is: What should we do with the 'context processors'? Should I add a 'Copy' button as well? Initially, I thought we had the main context data, and I planned to implement 'Copy' as a list. That's why I'm bringing this up. Please let me know.

Screenshot 2025-01-12 at 20 12 38


**2. I wonder if we should use a dictionary with keys identifying where each of the contexts have come from. That's a bit bigger of a change and should be a separate issue.**

I'll search about it, i'm still learning about `django-debug-toolbar` :)

I'll try my best implement middleware for this, please give me your suggestion, how can I do it?)

**3. Last aside, some of the values are not valid primitives or common classes. They are things like "my_class" : <MyClass prop=Hello world other=10 ...> what do you suggest we do when that's the case? That value can not be pasted into a shell as is.**
Maybe `repr()` can help us. What do you think? Skipping data is another option, but I don't think it's a good idea.

tim-schilling:
> Yes, I agree with you. My next question is: What should we do with the 'context processors'? Should I add a 'Copy' button as well? Initially, I thought we had the main context data, and I planned to implement 'Copy' as a list. That's why I'm bringing this up. Please let me know.

Oh dang. I'm sorry, I definitely didn't understand things properly. I thought the multiple dictionaries were coming from the context processors. They aren't. This is new to me.

Apparently the context for a template is stacked, which is why we have the list here. Though when the template is rendered, it just shows the flattened version. So, yes I think a list of dictionaries is appropriate here.

Though I think we may want to create an issue to inform the developer about this. The developer should be aware that multiple contexts are used for the view and what the flattened context dictionary was effectively used to render the template. 


> Maybe repr() can help us. What do you think? Skipping data is another option, but I don't think it's a good idea.

I actually think we can punt on this part. When the serializable branch gets merged, it's going to be converted to JSON and at that point, this will be "solved". Or at least decided for us.

---

It seems like the main change is the copy button (done, but needs tests) and displaying the contexts as a list.

@matthiask any opinions?

thesayfulla:
@tim-schilling Hello! I hope you're doing well.
I've added a "Copy" button to each context item and also created a list view. Now, I'd like to write tests for this feature. Could you please share some resources or guidance on how to approach testing this?

Thank you!


### List View
Screenshot 2025-03-19 at 05 24 15

### Context Item View
Screenshot 2025-03-19 at 05 24 33

matthiask:
@thesayfulla We have some selenium integration tests in the repo, they could be extended to do a real end to end test of this new feature. Otherwise, I'd have a look at the template panel testcases here, and add another testcase which verifies that the generated HTML has the expected format and escaping:
https://github.com/django-commons/django-debug-toolbar/blob/main/tests/panels/test_template.py

thesayfulla:
hello @matthiask, thank you a lot. I'll try.

thesayfulla:
Sorry guys, I'm going to close this PR. I'll make new one as soon as possible.

tim-schilling:
All good. Take your time. I hope you're doing well!

PR #2066, opened 2025-01-30
Make show toolbar callback function async/sync compatible.

### LINKED ISSUE #2061: Version 5.0 breaks show_toolbar callback accessing database
## Description

Since version 5.0, setting `SHOW_TOOLBAR_CALLBACK` to a function that uses the ORM to access the database raises the following exception:

> SynchronousOnlyOperation
> You cannot call this from an async context - use a thread or sync_to_async.

An example function that triggers this issue:

```python
def show_toolbar(request):
    return request.user.is_staff
```

As `request.user` is a lazy object, it'll be retrieved during the `show_toolbar` execution. As some panels are asynchronous now, Django Debug Toolbar will try to access to the database from an asynchronous context.

Converting the callback to asynchronous does **not** work either, because:
* The `DebugToolbarMiddleware.__acall__` method does not await for the `show_toolbar(request)` call.
* The `DebugToolbarMiddleware.__call__` method calls `show_toolbar(request)`, which will always be true, as Python will return a coroutine object.

## Possible solution

Allow both sync and async `SHOW_TOOLBAR_CALLBACK` functions, and:
* In `DebugToolbarMiddleware.__call__` method, cast the callback using `async_to_sync` if the provided function is asynchronous.
* In `DebugToolbarMiddleware.__acall__` method, cast the callback using `sync_to_async` if the provided function is synchronous, and add the `await` to its execution.

tim-schilling:
#### Description

This checks if the SHOW_TOOLBAR_CALLBACK is a coroutine if we're in async mode and the reverse if it's not. It will automatically wrap the function with sync_to_async or async_to_sync when necessary.

Fixes #2061

#### Checklist:


adamantike:
I can confirm these changes completely fix the issue reported in #2061, both when using a synchronous callback or an asynchronous one!

tim-schilling:
> I don't like very much to for the use sync_to_async all the time

To be clear, if you're using a sync callback with `async_mode=False`, it doesn't wrap it. Similarly, if you're using an async callback with `async_mode=True`, it doesn't wrap it. It only wraps if the callback is written for the opposite of `async_mode`.

Is that what you understood @elineda?

elineda:
> > I don't like very much to for the use sync_to_async all the time
> 
> To be clear, if you're using a sync callback with `async_mode=False`, it doesn't wrap it. Similarly, if you're using an async callback with `async_mode=True`, it doesn't wrap it. It only wraps if the callback is written for the opposite of `async_mode`.
> 
> Is that what you understood @elineda?

yeah I missed that. In that case we can't have both sync and async fonction in the same time, but for a dev dependencies I think it's ok.

ok for me

debug_toolbar/decorators.py
```diff
@@ -11,7 +11,7 @@ def require_show_toolbar(view):
     def inner(request, *args, **kwargs):
         from debug_toolbar.middleware import get_show_toolbar
 
-        show_toolbar = get_show_toolbar()
```
salty-ivy, line 14: Shouldn't we pass `async_mode` based on the type of request `WSGIRequest` or `ASGIRequest` here ? I mean I am not sure whether `require_show_toolbar` would ever be exposed to a custom `async view` in async context or vice versa but this would save the conversion as it slows down  the overall process. thoughts?

debug_toolbar/decorators.py
```diff
@@ -11,7 +11,7 @@ def require_show_toolbar(view):
...
+        show_toolbar = get_show_toolbar(async_mode=False)
```
salty-ivy, line 14: Shouldn't we pass `async_mode` based on the type of request `WSGIRequest` or `ASGIRequest` here ? I mean I am not sure whether `require_show_toolbar` would ever be exposed to a custom `async view` in async context or vice versa but this would save the conversion as it slows down  the overall process. thoughts?

tim-schilling: I was hoping `require_show_toolbar` was an undocumented API, but it is documented. Yes, we need to pass in a better value than always `False`. Good catch @salty-ivy!

It feels like we'd want `async_mode` defined by whether `view` is a coroutine though rather than based on WSGIRequest vs ASGIRequest. Not 100% sure though.

salty-ivy: It can possible break in a situation where everything is in async context
including `show_toolbar` call back and if I use `require_show_toolbar` on an `async view` it will forcefully convert it into a sync func and we would get into similar error given that `require_show_toolbar` is exposed to such conditions :D

salty-ivy: > It feels like we'd want `async_mode` defined by whether `view` is a coroutine though rather than based on WSGIRequest vs ASGIRequest. Not 100% sure though.

may be checking both but it seems like that would cause a rabbit hole : (

tim-schilling: Oh. I totally missed that this decorator needs to be async friendly.

salty-ivy: > It feels like we'd want `async_mode` defined by whether `view` is a coroutine though rather than based on WSGIRequest vs ASGIRequest. Not 100% sure though.

This is correct, we should check whether view is a coroutine or not since django also converts the request object based on that, just checked : D

elineda: yeah missed that too and yeah a iscoroutinefunction is the way other django decorator work.

salty-ivy:
@tim-schilling Any idea why tox tests are failing in all the recent PRs ?

matthiask:
@salty-ivy The reason for this is #2082 , I have now pushed a change pinning django-csp to versions before 4 for now. This allows us to continue working on everything else. Please resubmit the PR (merge main and push again) so that tests have a chance to run again? Thank you!

salty-ivy:
@matthiask Thanks!  @tim-schilling @elineda I think I think this is ready for another review.

matthiask (approved):

salty-ivy:
I have also pushed documentation changes Its taking some time to reflect on github :D

elineda (approved):

Issue #2068, opened 2025-02-05
Version 5.0 breaks collectstatic  when using some storage backends


machakux:
When using Google Cloud Storage backend via [django-storages](https://github.com/jschneier/django-storages) running `./manage.py collectstatic` fails with some `ValueError`. Based on my analysis it seems the error has something to do with `DebugStaticFilesStorage` objects from debug toolbar 5. Downgrading to version 4.4.* seems to solve the issue.

The issue can be encounteres when using
`django-debug-toolbar==5.0.1`
`Django==5.1.5`
`django-storages[google]==1.14.4`

A sample error returned by `./manage.py collectstatic` ...
```
  File "./manage.py", line 22, in <module>
    main()
  File "./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File ".virtualenvs/test/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File ".virtualenvs/test/lib/python3.10/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File ".virtualenvs/test/lib/python3.10/site-packages/django/core/management/base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
  File ".virtualenvs/test/lib/python3.10/site-packages/django/core/management/base.py", line 459, in execute
    output = self.handle(*args, **options)
  File ".virtualenvs/test/lib/python3.10/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 209, in handle
    collected = self.collect()
  File ".virtualenvs/test/lib/python3.10/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 135, in collect
    handler(path, prefixed_path, storage)
  File ".virtualenvs/test/lib/python3.10/site-packages/django/contrib/staticfiles/management/commands/collectstatic.py", line 378, in copy_file
    self.storage.save(prefixed_path, source_file)
  File ".virtualenvs/test/lib/python3.10/site-packages/django/core/files/storage/base.py", line 49, in save
    name = self._save(name, content)
  File ".virtualenvs/test/lib/python3.10/site-packages/storages/backends/gcloud.py", line 192, in _save
    file_object = GoogleCloudFile(name, "rw", self)
  File ".virtualenvs/test/lib/python3.10/site-packages/storages/backends/gcloud.py", line 45, in __init__
[...]

PR #2069, opened 2025-02-06
Refs #2068: Do not reinstantiate staticfiles storage classes

matthiask:
#### Description

this information as well.

Fixes # (issue)

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -3,10 +3,8 @@ import uuid
 from contextvars import ContextVar
 from os.path import join, normpath
 
-from django.conf import settings
 from django.contrib.staticfiles import finders, storage
 from django.dispatch import Signal
-from django.utils.functional import LazyObject
 from django.utils.translation import gettext_lazy as _, ngettext
 
 from debug_toolbar import panels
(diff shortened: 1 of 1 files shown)


matthiask:
@tim-schilling Yes, I think we could directly monkey-patch the `url` method on the static files storage instance. Do you think that would be preferrable to the `Mixin` injection? I can see how a straightforward monkeypatch might be easier to understand.

tim-schilling:
I think either case works well. I haven't manipulated `__mro__` often, but if it works I'm good with it.

matthiask:
I have done too many hacks with `__bases__`, and monkey patching the method seems to be a better way to go. 

If we want a quick bugfix release for #2068 I can finish this, but maybe this would also be a good fit for our Djangonauts? Especially if we can wait a few days or weeks for a fix.

tim-schilling:
I'm good with waiting.

dr-rompecabezas:
I assigned #2068 to myself, thinking all this PR needed was tests, but now I see you were considering a different approach (monkey patching). I'm trying to understand the proposal to see if I can help.

matthiask:
So, monkeypatching would mean something like this:

```python
def patched_url(self, path):
    url = orig_url(path)
    ... our own customizations
    return url

orig_url = storage.staticfiles_storage.url
storage.staticfiles_storage.url = patched_url
``` 

It's a less tricky way to achieve basically the same thing my MRO hack did. I'm not immediately aware of any downsides either way. Also, runtime patching is frowned upon, but we have to do something ugly to add our instrumentation.

tim-schilling:
I suspect the MRO hack may be less intrusive to other devs. If we monkey patch things, we'd want to consider using `functools.wraps` or something similar to keep the method interface the same.

PR #2088, opened 2025-02-26
Refactor on csp_nonce usage with django-csp

### LINKED ISSUE #2082: Fix failing tests in CspRenderingTestCase
Related to a `django-csp` package incompatibility, as reported in [a comment](https://github.com/django-commons/django-debug-toolbar/pull/2081#issuecomment-2664909933) for PR #2081.

Edit: see also seemingly related conversation in issue #2058.

tim-schilling:
#### Description

This refactors how the CSP nonce is fetched. It's now done as a toolbar property and wraps the attribute `request.csp_nonce`

This avoids the toolbar from generating a nonce that gets injected into the CSP header when the view doesn't expect it to. It also supports using a nonce that is generated from any other point while processing the request, including other middleware.

Fixes #2082 

#### Checklist:


tests/views.py
```diff
@@ -42,6 +42,11 @@ def regular_view(request, title):
     return render(request, "basic.html", {"title": title})
 
 
+def csp_view(request):
+    """Use request.csp_nonce to inject it into the headers"""
+    return render(request, "basic.html", {"title": f"CSP {request.csp_nonce}"})
```
tim-schilling, line 47: The nonce needed to be rendered in the view. I chose to do it here.

tests/test_csp_rendering.py
```diff
@@ -63,70 +70,97 @@ def _fail_on_invalid_html(self, content: bytes, parser: HTMLParser):
...
-        toolbar = list(DebugToolbar._store.values())[0]
-        nonce = str(toolbar.request.csp_nonce)
-        self._fail_if_missing(
-            root=html_root, path=".//link", namespaces=namespaces, nonce=nonce
-        )
-        self._fail_if_missing(
-            root=html_root, path=".//script", namespaces=namespaces, nonce=nonce
-        )
+        for middleware in [MIDDLEWARE_CSP_BEFORE, MIDDLEWARE_CSP_LAST]:
+            with self.settings(MIDDLEWARE=middleware):
```
tim-schilling, line 76: It become important to test with both configurations of middleware. Or at least it will be more important if we stop relying on `request._csp_nonce`

tests/test_csp_rendering.py
```diff
@@ -13,6 +13,13 @@
 
 from .base import IntegrationTestCase
 
+MIDDLEWARE_CSP_BEFORE = settings.MIDDLEWARE[:]
```
robhudson, line 16: Python lists have a `.copy()` method which does the same but maybe a bit clearer?

tim-schilling: Makes sense!

robhudson:
The latest commit on django-csp changes the way nonce access is handled that may help this situation and not have to use the private attribute.

If the nonce was used in the content it will still be available after the middleware has processed the response so other middleware can reference it.

If the nonce was NOT used in the content it will raise the error, but also checking the nonce via `if request.csp_nonce` will return `False`.

You can see the changes in the PR: https://github.com/mozilla/django-csp/pull/269

jwhitlock:
OK, what if in step 4 `request.csp_nonce` is falsy? Would that work for you?

1. CSPMiddleware pre-view, configures the lazy `request.csp_nonce` with `FalseLazyObject`
2. DebugToolbarMiddleware pre-view, does nothing relevant
3. View doesn't use `request.csp_nonce`, but has static assets in the rendered template
4. DebugToolbarMiddleware post view, finds `request.csp_nonce` as falsly, does not generate a nonce or use it for the toolbar's static assets
5. CSPMiddleware sees that `request.csp_nonce` is still unset and omits it from the header

tim-schilling:
@jwhitlock yup, yup. That would work well. A read-only csp_nonce is what the toolbar needs and why relying on the private attribute worked. I don't know how much of `django-csp`'s API should be catered to this particular problem. I do appreciate the efforts you've gone to already.

jwhitlock:
Thanks, I should be able to get that PR open today or tomorrow.

Until then, maybe the solution is to exclude django-csp [4.0b4](https://github.com/mozilla/django-csp/releases/tag/4.0b4)? It seems `4.0b3` was OK, and hopefully `4.0b5` will as well.

tim-schilling:
This has been updated to properly handle the new csp middleware as discussed in https://github.com/mozilla/django-csp/issues/268

debug_toolbar/toolbar.py
```diff
@@ -65,6 +65,20 @@ def enabled_panels(self):
...
+    def csp_nonce(self):
+        """
+        Look up the Content Security Policy nonce if there is one.
+
+        This is built specifically for django-csp, which may not always
+        have a nonce associated with the request. Use the private attribute
+        because the lazy object wrapped value can generate a nonce by
+        accessing it. This isn't ideal when the toolbar is injecting context
+        into the response because it may set a nonce not used with
+        other assets.
```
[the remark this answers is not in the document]
tim-schilling: Thank you. Good catch!

docs/spelling_wordlist.txt
```diff
@@ -16,6 +16,7 @@ backported
 biome
 checkbox
 contrib
+csp
```
matthiask, line 19: Do we need this?

matthiask (approved):
LGTM! Thank you.

I didn't know `self.settings()`, there's always something new.

docs/spelling_wordlist.txt
tim-schilling: Yes, I think my note in the changelog triggered it as a word despite being `django-csp`

robhudson (approved):
This looks good to me. django-csp==4.0b7 is out now. Hopefully that's the last one before the official 4.0 release.

PR #2089, opened 2025-02-28
Pull translations from transifex

matthiask:
I don't know if I want to do this. Everything is political after all.


tim-schilling (approved):
I think we should merge this. This benefits all Russian speakers, even those who are not based in Russia. I don't feel like by merging this we are making a statement in favor of the war or Russia's politics in general.

matthiask:
Thank you. I can agree to that.

PR #2091, opened 2025-03-01
Enable a few of the disabled biome rules 

matthiask:
#### Description

During the initial migration to biome I disabled a few rules to keep the code
changes minimal. I reenable some of the less controversial rules here.

The remaining rules are discouraging `forEach` in favor of `for ... of` (not done 
for now because of the amount of change it introduces) and the discouragement
of assignments in expressions. I want to have a closer look there because the
code reads strange (but it probably isn't).

#### Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,6 +27,7 @@ repos:
     rev: v1.9.4
     hooks:
       - id: biome-check
+        verbose: true
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: 'v0.9.7'
     hooks:
--- a/debug_toolbar/static/debug_toolbar/js/timer.js
+++ b/debug_toolbar/static/debug_toolbar/js/timer.js
@@ -9,9 +9,8 @@ function insertBrowserTiming() {
             return (
                 ((performance.timing[stat] - timingOffset) / totalTime) * 100.0
             );
-        } else {
-            return 0;
         }
+        return 0;
     }
     function getCSSWidth(stat, endStat) {
         let width = 0;
--- a/debug_toolbar/static/debug_toolbar/js/timer.js
+++ b/debug_toolbar/static/debug_toolbar/js/timer.js
@@ -27,36 +27,31 @@ function insertBrowserTiming() {
         } else {
             width = 0;
         }
-        return width < 1 ? "2px" : width + "%";
+        return width < 1 ? "2px" : `${width}%`;
     }
     function addRow(tbody, stat, endStat) {
         const row = document.createElement("tr");
+        const elapsed = performance.timing[stat] - timingOffset;
         if (endStat) {
+            const duration =
+                performance.timing[endStat] - performance.timing[stat];
             // Render a start through end bar
-            row.innerHTML =
-                "<td>" +
-                stat.replace("Start", "") +
-                "</td>" +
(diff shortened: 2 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -27,6 +27,8 @@ repos:
     rev: v1.9.4
     hooks:
       - id: biome-check
+        # args: [--unsafe]
```
tim-schilling, line 30: I'm not sure if this is a WIP or some debugging code.

debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -116,48 +116,44 @@ const djdt = {
...
             this.closest(".djDebugPanelContent")
-                .querySelectorAll(".djToggleDetails_" + id)
-                .forEach(function (e) {
+                .querySelectorAll(`.djToggleDetails_${id}`)
+                .forEach((e) => {
                     if (openMe) {
                         e.classList.add("djSelected");
                         e.classList.remove("djUnselected");
-                        self.textContent = toggleClose;
+                        this.textContent = toggleClose;
```
tim-schilling, line 132: Good catch

debug_toolbar/static/debug_toolbar/js/timer.js
```diff
@@ -28,36 +27,20 @@ function insertBrowserTiming() {
...
-                "<td>" +
-                stat.replace("Start", "") +
-                "</td>" +
-                '<td><svg class="djDebugLineChart" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 100 5" preserveAspectRatio="none"><rect y="0" height="5" fill="#ccc" /></svg></td>' +
-                "<td>" +
-                (performance.timing[stat] - timingOffset) +
-                " (+" +
-                (performance.timing[endStat] - performance.timing[stat]) +
-                ")</td>";
+            row.innerHTML = `<td>${stat.replace("Start", "")}</td><td><svg class="djDebugLineChart" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 100 5" preserveAspectRatio="none"><rect y="0" height="5" fill="#ccc" /></svg></td><td>${performance.timing[stat] - timingOffset} (+${performance.timing[endStat] - performance.timing[stat]})</td>`;
```
tim-schilling, line 36: I think this change and the one on line 43 are a step backwards in terms of maintainability.

.pre-commit-config.yaml
matthiask: Thanks! I left that it because activating it temporarily may be useful when modernizing code, but I should either add a comment or remove it again.

debug_toolbar/static/debug_toolbar/js/toolbar.js
matthiask: Not a bugfix by the way, just something enabled by the use of arrow functions instead of function() {}.

debug_toolbar/static/debug_toolbar/js/timer.js
matthiask: You're not objecting to the use of the template literal I hope, just to the fact that we're using too much inline code here and have lost the line wrapping? I'll fix that.

tim-schilling: I didn't even notice the template literal :sweat_smile: Yes, my issue is that the line of code is really long. I think the multiple lines helped convey what logic was occurring, then letting long static text exist where it needed to be.

tim-schilling (approved):
:rocket:

PR #2099, opened 2025-03-08
Consolidate csp_nonce usages to a single property on the toolbar.

tim-schilling:
#### Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,7 +25,7 @@ deps =
     pygments
     selenium>=4.8.0
     sqlparse
-    django-csp<4
+    django-csp
 passenv=
     CI
     COVERAGE_ARGS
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -25,7 +25,7 @@ deps =
     pygments
     selenium>=4.8.0
     sqlparse
-    django-csp<4
+    django-csp
```
robhudson, line 28: Since the django-csp v4 releases are beta, a pre-release, will this pick up v3.8, the last published release? I'm not up on if the intention here is to test against v4 or v3.x but I believe django-csp has solved this issue for both versions anyway by adding the `CheckableLazyObject` in the v4 betas.

tim-schilling: Yes, we test against pre-releases in ci. I think that's why we caught this early.

PR #2101, opened 2025-03-10
Fix for exception-unhandled "forked" Promise chain

karolyi:
See https://github.com/django-commons/django-debug-toolbar/pull/2100

I took the liberty of reformatting a bit of code here, I'm sure the autoformatter will have its way with it.

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -330,11 +330,12 @@ const djdt = {
                 if (response.headers.get("djdt-store-id") !== null) {
                     try {
                         handleAjaxResponse(
-                            response.headers.get("djdt-store-id"));
+                            response.headers.get("djdt-store-id")
+                        );
                     } catch (err) {
                         throw new Error(
-                            `A(n) "${err.name}" happened within `
-                            `django-debug-toolbar: ${err.message}`)
+                            `A(n) "${err.name}" happened within ``django-debug-toolbar: ${err.message}`
+                        );
                     }
                 }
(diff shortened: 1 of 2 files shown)


debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -321,16 +321,25 @@ const djdt = {
...
+            return promise.then((response) => {
                 if (response.headers.get("djdt-store-id") !== null) {
-                    handleAjaxResponse(response.headers.get("djdt-store-id"));
+                    try {
+                        handleAjaxResponse(
+                            response.headers.get("djdt-store-id")
+                        );
+                    } catch (err) {
+                        throw new Error(
+                            `A(n) "${err.name}" happened within ``django-debug-toolbar: ${err.message}`
```
tim-schilling, line 337:
```suggestion
                            `"${err.name}" occurred within django-debug-toolbar: ${err.message}`
```

karolyi:
I don't find splitting hairs on wording useful, but so be it.

tim-schilling (approved):
This looks reasonable to me. If someone else can pull it down and confirm it works, we can release it. I'll have time later this week otherwise.

karolyi:
Does it count when I say this change fixed my original problem outlined? After all, I was the OG PR creator.

matthiask:
I've tested it locally and what worked before still seems to work fine. Will do more tests later. The reason provided in the other PR why writing tests is hard makes sense to me and the comment in the code is sufficient to hopefully be aware of this issue when we have to change the code again in the future, for whatever reason.

PR #2103, opened 2025-03-12
Add data sanitization to RequestPanel

### LINKED ISSUE #2074: Sanitize `RequestPanel` data behind a setting
We should sanitize the `RequestPanel` stats behind a setting. This would prevent us from storing passwords from monitored `/login/` requests and other sensitive data. While this isn't a major concern while the toolbar runs in memory, when we start storing data this becomes a big concern.

This is somewhat related to https://github.com/django-commons/django-debug-toolbar/issues/2048#issuecomment-2574215884. See https://github.com/django/django/blob/main/django/views/debug.py#L115-L130 for an example on how to do this.

dr-rompecabezas:
#### Description

This implementation:

1. Makes sanitization of sensitive values enabled by default but configurable
2. Uses a similar approach to--and the same default key patterns as--Django's sanitization 
3. Allows customizing the patterns used to identify sensitive data

- Added `SANITIZE_REQUEST_DATA` setting to control sanitization of request data.
- Added `REQUEST_SANITIZATION_PATTERNS` setting to define `key` patterns to sanitize the respective `value`.
- Implemented `sanitize_value` function to sanitize sensitive data based on `key` regex patterns.
- Updated `get_sorted_request_variable` to support sanitization.
- Added tests for sanitization of GET and POST parameters, cookies, and session data.
- Updated documentation to include new settings.

Fixes #2074

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/request.py
+++ b/debug_toolbar/panels/request.py
@@ -3,7 +3,11 @@ from django.urls import resolve
 from django.utils.translation import gettext_lazy as _
 
 from debug_toolbar.panels import Panel
-from debug_toolbar.utils import get_name_from_obj, get_sorted_request_variable
+from debug_toolbar.utils import (
+    get_name_from_obj,
+    get_sorted_request_variable,
+    sanitize_value,
+)
 
 
 class RequestPanel(Panel):
@@ -24,6 +28,8 @@ class RequestPanel(Panel):
         return view_func.rsplit(".", 1)[-1]
 
     def generate_stats(self, request, response):
+        from debug_toolbar import settings as dt_settings
+
         self.record_stats(
             {
                 "get": get_sorted_request_variable(request.GET),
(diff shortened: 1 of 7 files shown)


debug_toolbar/settings.py
```diff
@@ -45,6 +45,17 @@
...
+    "REQUEST_SANITIZATION_PATTERNS": (
+        "API",
+        "AUTH",
+        "TOKEN",
+        "KEY",
+        "SECRET",
+        "PASS",
+        "SIGNATURE",
+        "HTTP_COOKIE",
+    ),
```
matthiask, line 58: I don't think this list has to be configurable. The equivalent list isn't configurable in Django either.

matthiask (commented):
It's a great start! Thank you.

I'd like you to investigate whether it's possible to use the `SafeExceptionReporterFilter` instead of duplicating the filtering and cleaning logic ourselves, especially the list of patterns. The list of patterns isn't updated often in Django but it does happen from time to time, and if we can just reuse the functionality in there we immediately profit as well without doing anything.

https://github.com/django/django/blob/955b7c6ba105b328f387a9d63540dbabd4a05828/django/views/debug.py#L108

debug_toolbar/panels/request.py
```diff
@@ -24,6 +28,8 @@ def nav_subtitle(self):
         return view_func.rsplit(".", 1)[-1]
 
     def generate_stats(self, request, response):
+        from debug_toolbar import settings as dt_settings
```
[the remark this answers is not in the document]
matthiask: You're right, that's the reason for doing imports inside functions. It's not a problem if doing it is necessary, but it's better if we can move imports to the top.

debug_toolbar/panels/request.py
```diff
@@ -58,14 +64,35 @@ def generate_stats(self, request, response):
...
+                    session_list = []
+                    for k in request.session.keys():
+                        v = request.session.get(k)
+                        sanitized_v = sanitize_value(k, v)
+                        session_list.append((k, sanitized_v))
+                else:
+                    session_list = [
+                        (k, request.session.get(k)) for k in request.session.keys()
+                    ]
+
```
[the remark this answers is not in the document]
dr-rompecabezas: Absolutely. I'll refactor for DRYer code.

debug_toolbar/settings.py
dr-rompecabezas: For the toolbar, shouldn't we let the user decide what patterns to sanitize? for example, I might not care to sanitize crsftoken (captured by the TOKEN pattern) but I want to sanitize my API keys. Perhaps there is context that I am missing here.

matthiask: Tim (@tim-schilling) raised the issue, maybe he has a different view on this. My thinking is that if we do not have a good reason to deviate from Django's lead we shouldn't, but maybe there's something *I* am missing?

tim-schilling: Agreed with Matthias, though I would be open to allowing a customized list in the future if folks request it.

dr-rompecabezas: As agreed, there will be no configurable settings for this feature.

dr-rompecabezas:
I will be opening a new PR soon with a new approach.

PR #2104, opened 2025-03-12
Enhance RedirectsPanel with customizable redirect response hook

### LINKED ISSUE #1679: Original response is lost when redirection is intercepted
The `RedirectsPanel` creates a new response, preserving only the cookies of the original response. Then any additional information in the original response is lost and cannot be made visible in the toolbar’s panels. In our case, each response has authorization information and I would like to be able to verify that this information is captured correctly also when the user is ultimately redirected. The fix looks quite straightforward, within the `if`:
https://github.com/jazzband/django-debug-toolbar/blob/f1387801e84a343fbce5c7955423f87589790a26/debug_toolbar/panels/redirects.py#L20
modify the code creating the new response to
```python
    original_response = response
    response = SimpleTemplateResponse( ... )
    response.original_response = original_response
```
Everything else will be taken care of by custom code in the relevant panels...

blingblin-g:
#### Description

This PR introduces a hook in the RedirectsPanel to allow subclasses to customize the redirect response. This enhancement enables developers to extend the panel's functionality without modifying the base class, providing more flexibility in handling redirect responses.

The motivation behind this change is to improve the extensibility of the RedirectsPanel and provide a way for developers to capture and process additional information from the original response during redirects.

Fixes #1679

#### Checklist:


tim-schilling:
As per Djangonaut Space call:
- Matthias suggested adding a note to the RedirectsPanel documentation

matthiask:
Thank you!

Issue #2096, opened 2025-03-04
Dark mode conflicts with pygments


jefftriplett:
While DDT served the right template and saved the day for me, Tim pointed out that dark mode conflicts with pygments and he asked me to open an issue. See the screenshot. 

context: https://mastodon.social/@CodenameTim@fosstodon.org/114105865490684530 

![Image](https://github.com/user-attachments/assets/99277041-3497-4f7f-8550-743e0ace8814)

[Off topic] It was also pointed out in that thread that there was a Dracula theme for pygments in case that's a useful data point. 

https://draculatheme.com/pygments

matthiask:
Thanks @andoriyaprashant 

It seems that the issue hasn't been fixed; maybe you hadn't installed the pygments library and syntax highlighting wasn't active at all?

matthiask:
I think you have to do something like the attached patch. Currently, the CSS has to be included twice -- once for the preference (media query) and once for the explicit switch, that would maybe be fixed also by only evaluating the media query in the JavaScript code.

[patch.txt](https://github.com/user-attachments/files/19338711/patch.txt)

matthiask:
Hey @andoriyaprashant 

I want you to succeed, can I do anything to help? Maybe concentrate on what's in patch.txt first, the `matchMedia` stuff mentioned above can be cleaned up later.

PR #2116, opened 2025-03-31
Refs #2096: The theme selector now controls all colors

matthiask:
#### Description

Previously, if the system preference was dark mode and the user explicitly selected the light theme, the @media block still interferred with the styling. This is fixed by only evaluating the color scheme preference when initializing the toolbar and later only looking at our own selected theme.

Also, this pull request drops the `DEFAULT_THEME` setting, and instead prefers looking directly at the preferences of users' systems. Removing settings is most often a plus in my book. The main reason for removing it was to make the code easier to follow, because there's one less theme preference source to look at.

Refs #2096 

#### Checklist:

### DIFF
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -45,7 +45,6 @@ CONFIG_DEFAULTS = {
     "TOOLBAR_LANGUAGE": None,
     "IS_RUNNING_TESTS": "test" in sys.argv,
     "UPDATE_ON_FETCH": False,
-    "DEFAULT_THEME": "auto",
 }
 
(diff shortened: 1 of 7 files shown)


tim-schilling (approved):
I'm good with this change. We'll need to remember this when doing the release to explicitly call it out. Maybe we should cut another one just so we don't forget?

matthiask:
Yep, I think we should do a release soon-ish, especially since the release notes of 5.1 mentioned that we fixed the dark mode issue (but it has only been fixed for real with #2114 and this change).

PR #2117, opened 2025-03-31
Fix #2109: Recursively unwrap loaders to support template partials

### LINKED ISSUE #2109: django-template-partials and the templates panel do not work well together it seems
Loading template sources crashes:

```
[19/Mar/2025 11:06:46] "GET /__debug__/render_panel/?store_id=f32af068303746c588911011cbf7da3f&panel_id=TemplatesPanel HTTP/1.1" 200 249515
Internal Server Error: /__debug__/template_source/
  File ".../.venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
               ^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../projects/django-debug-toolbar/debug_toolbar/decorators.py", line 34, in inner
    return view(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../projects/django-debug-toolbar/debug_toolbar/decorators.py", line 46, in inner
    return view(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../projects/django-debug-toolbar/debug_toolbar/panels/templates/views.py", line 45, in template_source
    source = loader.get_contents(origin)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.12/site-packages/django/template/loaders/cached.py", line 27, in get_contents
    return origin.loader.get_contents(origin)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_contents'
```

(The strange debug toolbar paths are there because I use an editable install)

matthiask:
#### Description

The current template source view only supports unwrapping one level of 
template loaders depending on other loaders.

When using django-template-partials with cached templates we have to
unwrap the loader twice to get to the concrete loaders.

Fixes #2109

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/templates/views.py
+++ b/debug_toolbar/panels/templates/views.py
@@ -27,15 +27,18 @@ def template_source(request):
     template_name = request.GET.get("template", template_origin_name)
 
     final_loaders = []
-    loaders = Engine.get_default().template_loaders
+    loaders = list(Engine.get_default().template_loaders)
 
-    for loader in loaders:
+    while loaders:
+        loader = loaders.pop(0)
+
         if loader is not None:
-            # When the loader has loaders associated with it,
-            # append those loaders to the list. This occurs with
-            # django.template.loaders.cached.Loader
+            # Recursively unwrap loaders until we get to loaders which do not
+            # themselves wrap other loaders. This occurs
+            # with django.template.loaders.cached.Loader and the
+            # django-template-partials loader (possibly among others)
             if hasattr(loader, "loaders"):
-                final_loaders += loader.loaders
+                loaders.extend(loader.loaders)
             else:
                 final_loaders.append(loader)
 
--- a/debug_toolbar/panels/templates/views.py
+++ b/debug_toolbar/panels/templates/views.py
@@ -34,8 +34,8 @@ def template_source(request):
 
         if loader is not None:
             # Recursively unwrap loaders until we get to loaders which do not
-            # themselves wrap other loaders. This occurs
-            # with django.template.loaders.cached.Loader and the
+            # themselves wrap other loaders. This adds support for
+            # django.template.loaders.cached.Loader and the
             # django-template-partials loader (possibly among others)
             if hasattr(loader, "loaders"):
                 loaders.extend(loader.loaders)
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -28,6 +28,7 @@ INSTALLED_APPS = [
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "debug_toolbar",
+    "template_partials",
     "tests",
 ]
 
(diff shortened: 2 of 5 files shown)


debug_toolbar/panels/templates/views.py
```diff
@@ -27,15 +27,18 @@ def template_source(request):
...
 
-    for loader in loaders:
         if loader is not None:
-            # When the loader has loaders associated with it,
-            # append those loaders to the list. This occurs with
-            # django.template.loaders.cached.Loader
+            # Recursively unwrap loaders until we get to loaders which do not
+            # themselves wrap other loaders. This occurs
+            # with django.template.loaders.cached.Loader and the
+            # django-template-partials loader (possibly among others)
```
matthiask, line 39:
```suggestion
            # themselves wrap other loaders. This adds support for
            # django.template.loaders.cached.Loader and the
            # django-template-partials loader (possibly among others)
```

tests/settings.py
```diff
@@ -27,6 +28,7 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "debug_toolbar",
+    "template_partials",
```
matthiask, line 31: This
```suggestion
    # We are not actively using template-partials; we just want more nesting
    # in our template loader configuration, see
    # https://github.com/django-commons/django-debug-toolbar/issues/2109
    "template_partials",
```

tests/settings.py
```diff
@@ -27,6 +28,10 @@
     "django.contrib.messages",
     "django.contrib.staticfiles",
     "debug_toolbar",
+    # We are not actively using template-partials; we just want more nesting
+    # in our template loader configuration, see
+    # https://github.com/django-commons/django-debug-toolbar/issues/2109
+    "template_partials",
```
tim-schilling, line 34: Should we always be testing with template partials? I think with the django-csp tests, we have specific tests that install the package and test against it. That said, this is likely to be a short-time thing. Both django-csp and template-partials are very likely to be merged to core in the 6.x series.

tim-schilling (approved):
Thank you for fixing this issue! My other comment isn't a required change, but a discussion topic for us.

tests/settings.py
matthiask: It does feel a bit bad, I agree. Another way to test this would have been to add a mock loader with a `loaders` attribute containing the cached loader configuration and testing the panel like that. I thought it easier and more understandable to add django-template-partials as a test dependency than doing the more involved mocking. 

We will have to revisit this when template-partials is merged (hopefully), either by switching to a different package which also wraps the cached loader or by adding the relevant testing code to our project.

TomHall2020:
Bit late to catch this here, but when I was trying to follow the contributing guide and use `make test` this was breaking on me as `django-template-partials` was only added to the tox config, and not to `requirements_dev.txt` as well

PR #2118, opened 2025-04-01
Only try resolving the docker gateway IP once per server startup and add some diagnostics

### LINKED ISSUE #2113: host.docker.internal DNS lookup causes 15s delay in offline environments
In environments without internet access and without a DNS resolver (fully offline setups), the use of socket.gethostbyname("host.docker.internal") causes a 15-second delay on every request due to DNS resolution timeout. This is particularly problematic when DEBUG = True is required during development or internal deployments.

Details:
I encountered this issue in a system that must remain completely offline (air-gapped). There is no DNS resolver available, and thus, any DNS lookup (such as for host.docker.internal) results in a timeout of ~15 seconds. The following code snippet demonstrates the issue:

```
    if not settings.DEBUG:
        return False

    # Test: settings
    if request.META.get("REMOTE_ADDR") in settings.INTERNAL_IPS:
        return True

    # Test: Docker
    try:
        # This is a hack for docker installations. It attempts to look
        # up the IP address of the docker host.
        # This is not guaranteed to work.
        docker_ip = (
            # Convert the last segment of the IP address to be .1
            ".".join(socket.gethostbyname("host.docker.internal").rsplit(".")[:-1])
            + ".1"
        )
```
This logic gives no way to avoid the DNS lookup if DEBUG = True. The delay occurs on every request, making development/debugging extremely painful in such setups.

Additionally, tools like the Django Debug Toolbar exacerbate the issue by triggering this DNS lookup repeatedly, which compounds the delay.

Suggested Solution:
Introduce a configurable way to bypass the host.docker.internal resolution in known offline environments. Perhaps a try-except with a timeout or a setting that disables Docker IP detection entirely.

Environment:

    Offline setup (no DNS resolver, no internet)

    Docker-based deployment

    Django (with Debug Toolbar in development)

matthiask:
#### Description

this information as well.

Fixes #2113.

#### Checklist:

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -19,6 +19,32 @@ from debug_toolbar import settings as dt_settings
 from debug_toolbar.toolbar import DebugToolbar
 from debug_toolbar.utils import clear_stack_trace_caches, is_processable_html_response
 
+_resolved_gateway_ip = None
+
+
+def _gateway_ip():
+    global _resolved_gateway_ip
+    if _resolved_gateway_ip is None:
+        print(
+            "[django-debug-toolbar] Trying to configure determine the Docker gateay address for autoconfiguration... ",
+            end="",
+        )
+        try:
+            # This is a hack for docker installations. It attempts to look
+            # up the IP address of the docker host.
+            # This is not guaranteed to work.
+            _resolved_gateway_ip = (
+                # Convert the last segment of the IP address to be .1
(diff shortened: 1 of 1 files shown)


matthiask:
Thank you @dr-rompecabezas 

We discussed this in the Djangonauts call and come to the conclusion that we're following a different implementation strategy to fix this. I'm closing the PR.

PR #2120, opened 2025-04-03
Format the templates using djade

matthiask:
#### Description

The djade template formatter standardizes `load` tags, translation tags and some other things.

Using autoformatters is practically always good in my opinion.

#### Checklist:

### DIFF
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -655,8 +655,8 @@ class SQLPanelTestCase(BaseTestCase):
         template_info = query["template_info"]
         template_name = os.path.basename(template_info["name"])
         self.assertEqual(template_name, "flat.html")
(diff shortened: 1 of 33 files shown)


tests/panels/test_sql.py
```diff
@@ -655,8 +655,8 @@ def test_flat_template_information(self):
         template_info = query["template_info"]
         template_name = os.path.basename(template_info["name"])
         self.assertEqual(template_name, "flat.html")
-        self.assertEqual(template_info["context"][2]["content"].strip(), "{{ users }}")
-        self.assertEqual(template_info["context"][2]["highlight"], True)
+        self.assertEqual(template_info["context"][3]["content"].strip(), "{{ users }}")
+        self.assertEqual(template_info["context"][3]["highlight"], True)
```
matthiask, line 659: The reason for the offset is that we have another empty line between `{% extends %}` and `{% block %}`.

PR #2121, opened 2025-04-03
Fixes #2073 -- Added DatabaseStore for persistent debug data storage.

### LINKED ISSUE #2073: Implement database backed store
_This work should occur off of the serializable branch._

The goal here is to implement a `DatabaseStore` that inherits from `toolbar.store.BaseStore` which is similar to `MemoryStore`.

Some considerations:
- We should be able to run the test suite with the `DatabaseStore` easily
- When running `make example`, we should be able to kill the process, start it up and see requests from the first `make example` run. This proves we're able to fetch data from the database.
- We should clear the data that's older than X hours or days in our `AppConfig.ready()`. We don't need to retain data from forever. This should be exposed as a setting and documented.

Sub-issues should be created from this to track functonality

dr-rompecabezas:
#### Description

- Introduced `DatabaseStore` to store debug toolbar data in the database.
- Added `DebugToolbarEntry` model and migrations for persistent storage.
- Updated documentation to include configuration for `DatabaseStore`.
- Added tests for `DatabaseStore` functionality, including CRUD operations and cache size enforcement.

Fixes #2073

#### Checklist:

### DIFF
--- a/debug_toolbar/models.py
+++ b/debug_toolbar/models.py
@@ -1,14 +1,15 @@
 from django.db import models
+from django.utils.translation import gettext_lazy as _
 
 
-class DebugToolbarEntry(models.Model):
+class HistoryEntry(models.Model):
     request_id = models.UUIDField(primary_key=True)
     data = models.JSONField(default=dict)
     created_at = models.DateTimeField(auto_now_add=True)
 
     class Meta:
-        verbose_name = "Debug Toolbar Entry"
-        verbose_name_plural = "Debug Toolbar Entries"
+        verbose_name = _("debug toolbar entry")
+        verbose_name_plural = _("debug toolbar entries")
         ordering = ["-created_at"]
 
     def __str__(self):
--- a/debug_toolbar/models.py
+++ b/debug_toolbar/models.py
@@ -1,0 +1,15 @@
+from django.db import models
+
+
+class DebugToolbarEntry(models.Model):
+    request_id = models.UUIDField(primary_key=True)
+    data = models.JSONField(default=dict)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        verbose_name = "Debug Toolbar Entry"
+        verbose_name_plural = "Debug Toolbar Entries"
+        ordering = ["-created_at"]
+
+    def __str__(self):
+        return str(self.request_id)
--- a/debug_toolbar/models.py
+++ b/debug_toolbar/models.py
@@ -8,8 +8,8 @@ class HistoryEntry(models.Model):
     created_at = models.DateTimeField(auto_now_add=True)
 
     class Meta:
-        verbose_name = _("debug toolbar entry")
-        verbose_name_plural = _("debug toolbar entries")
+        verbose_name = _("history entry")
+        verbose_name_plural = _("history entries")
         ordering = ["-created_at"]
 
     def __str__(self):
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -148,13 +148,8 @@ class DatabaseStore(BaseStore):
         Enforce the cache size limit - keeping only the most recently used entries
         up to RESULTS_CACHE_SIZE.
         """
-        # Get the cache size limit from settings
-        cache_size = dt_settings.get_config()["RESULTS_CACHE_SIZE"]
-
-        # Determine which entries to keep (the most recent ones up to cache_size)
-        keep_ids = list(
-            HistoryEntry.objects.all()[:cache_size].values_list("request_id", flat=True)
-        )
+        # Determine which entries to keep
+        keep_ids = cls.request_ids()
 
         # Delete all entries not in the keep list
         if keep_ids:
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -172,10 +172,11 @@ class DatabaseStore(BaseStore):
     def set(cls, request_id: str):
         """Set a request_id in the store and clean up old entries"""
         # Create the entry if it doesn't exist (ignore otherwise)
-        HistoryEntry.objects.get_or_create(request_id=request_id)
+        _, created = HistoryEntry.objects.get_or_create(request_id=request_id)
 
-        # Enforce the cache size limit to clean up old entries
-        cls._cleanup_old_entries()
+        # Only enforce cache size limit when new entries are created
+        if created:
+            cls._cleanup_old_entries()
 
     @classmethod
     def clear(cls):
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -193,9 +193,6 @@ class DatabaseStore(BaseStore):
     @classmethod
     def save_panel(cls, request_id: str, panel_id: str, data: Any = None):
         """Save the panel data for the given request_id"""
-        # First ensure older entries are cleared if we exceed cache size
-        cls.set(request_id)
-
         with transaction.atomic():
             obj, _ = HistoryEntry.objects.get_or_create(request_id=request_id)
             store_data = obj.data
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -153,9 +153,7 @@ class DatabaseStore(BaseStore):
 
         # Determine which entries to keep (the most recent ones up to cache_size)
         keep_ids = list(
-            HistoryEntry.objects.all()[:cache_size].values_list(
-                "request_id", flat=True
-            )
+            HistoryEntry.objects.all()[:cache_size].values_list("request_id", flat=True)
         )
 
         # Delete all entries not in the keep list
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -151,13 +151,16 @@ class DatabaseStore(BaseStore):
         # Get the cache size limit from settings
         cache_size = dt_settings.get_config()["RESULTS_CACHE_SIZE"]
 
-        # Get the IDs to keep as a subquery
-        recent_ids = HistoryEntry.objects.all()[:cache_size].values_list(
-            "request_id", flat=True
+        # Determine which entries to keep (the most recent ones up to cache_size)
+        keep_ids = list(
+            HistoryEntry.objects.all()[:cache_size].values_list(
+                "request_id", flat=True
+            )
         )
 
         # Delete all entries not in the keep list
-        HistoryEntry.objects.exclude(request_id__in=recent_ids).delete()
+        if keep_ids:
+            HistoryEntry.objects.exclude(request_id__in=keep_ids).delete()
 
     @classmethod
     def request_ids(cls):
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -6,10 +6,12 @@ from collections.abc import Iterable
 from typing import Any
 
 from django.core.serializers.json import DjangoJSONEncoder
+from django.utils import timezone
 from django.utils.encoding import force_str
 from django.utils.module_loading import import_string
 
 from debug_toolbar import settings as dt_settings
+from debug_toolbar.models import DebugToolbarEntry
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


matthiask:
Tests are all failing, you'd probably have to do a merge of `main` to fix this, but that's hard. So let's concentrate on the branch.

debug_toolbar/models.py
```diff
@@ -0,0 +1,15 @@
+from django.db import models
+
+
+class DebugToolbarEntry(models.Model):
```
matthiask, line 4: Very much [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding): I'd name this `HistoryEntry` or something.

debug_toolbar/models.py
```diff
@@ -0,0 +1,15 @@
...
+    request_id = models.UUIDField(primary_key=True)
+    data = models.JSONField(default=dict)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    class Meta:
+        verbose_name = "Debug Toolbar Entry"
+        verbose_name_plural = "Debug Toolbar Entries"
```
matthiask, line 11:
```suggestion
        verbose_name = _("Debug Toolbar Entry")
        verbose_name_plural = _("Debug Toolbar Entries")
```

Not sure about capitalization. I mostly write everything in lowercase.

debug_toolbar/store.py
```diff
@@ -140,5 +142,100 @@ def panels(cls, request_id: str) -> Any:
...
+        # Determine which entries to keep (the most recent ones up to cache_size)
+        keep_ids = list(
+            DebugToolbarEntry.objects.order_by("-created_at")[:cache_size].values_list(
+                "request_id", flat=True
+            )
+        )
+
+        # Delete all entries not in the keep list
+        if keep_ids:
+            DebugToolbarEntry.objects.exclude(request_id__in=keep_ids).delete()
```
matthiask, line 164: Unimportant/nitpicky: I think blindly running the `delete` query would be fine; you could remove the `list()` call above then. I haven't tested if that works, and if the `list()` call is necessary for some reason then I'm fine with this.

debug_toolbar/store.py
```diff
@@ -140,5 +142,100 @@ def panels(cls, request_id: str) -> Any:
...
+
+    @classmethod
+    def request_ids(cls):
+        """Return all stored request ids within the cache size limit"""
+        cache_size = dt_settings.get_config()["RESULTS_CACHE_SIZE"]
+        return list(
+            DebugToolbarEntry.objects.order_by("-created_at")[:cache_size].values_list(
```
matthiask, line 171: The `.order_by` isn't necessary when it's the default for the model. You should take advantage of the default ordering I think.

dr-rompecabezas:
The only failing tests are unrelated to this PR. Both tests have already been fixed, but the fixes have not yet been merged into the serializable branch.

```text
   ======================================================================
  ERROR: test_exists (tests.test_csp_rendering.CspRenderingTestCase)
  A `nonce` should exist when using the `CSPMiddleware`.
```

```text
   ======================================================================
  ERROR: test_redirects_exists (tests.test_csp_rendering.CspRenderingTestCase)
```

debug_toolbar/store.py
```diff
@@ -140,5 +142,100 @@ def panels(cls, request_id: str) -> Any:
...
+        return DebugToolbarEntry.objects.filter(request_id=request_id).exists()
+
+    @classmethod
+    def set(cls, request_id: str):
+        """Set a request_id in the store and clean up old entries"""
+        # Create or update the entry
+        obj, created = DebugToolbarEntry.objects.get_or_create(request_id=request_id)
+        if not created:
+            # Update timestamp to mark as recently used
+            obj.created_at = timezone.now()
```
matthiask, line 188: This doesn't do anything; `auto_now_add` overrides whatever you're setting.

If you want to set the `created_at` yourself you'd have to use `default=timezone.now`. 

But even then, updating the `created_at` timestamp when modifying the record doesn't make a lot of sense to me. After all, it's a *modification*, not the *creation*. I'd simplify the code by just ignoring the `not created` case here.

debug_toolbar/store.py
```diff
@@ -140,5 +142,100 @@ def panels(cls, request_id: str) -> Any:
...
+        """Delete the stored request for the given request_id"""
+        DebugToolbarEntry.objects.filter(request_id=request_id).delete()
+
+    @classmethod
+    def save_panel(cls, request_id: str, panel_id: str, data: Any = None):
+        """Save the panel data for the given request_id"""
+        # First ensure older entries are cleared if we exceed cache size
+        cls.set(request_id)
+
+        # Ensure the request exists
```
matthiask, line 210:
```suggestion
```

this comment doesn't seem too useful.

debug_toolbar/store.py
```diff
@@ -140,5 +142,100 @@ def panels(cls, request_id: str) -> Any:
...
+    @classmethod
+    def delete(cls, request_id: str):
+        """Delete the stored request for the given request_id"""
+        DebugToolbarEntry.objects.filter(request_id=request_id).delete()
+
+    @classmethod
+    def save_panel(cls, request_id: str, panel_id: str, data: Any = None):
+        """Save the panel data for the given request_id"""
+        # First ensure older entries are cleared if we exceed cache size
+        cls.set(request_id)
```
matthiask, line 208: I wonder what the `set` method is for. @tim-schilling , do you remember what the thinking behind it was when we have `save_panel` to really do saving? 

The call can probably be removed here because we're immediately doing `get_or_create` afterwards, so if the entry doesn't exist already saving the panel data works anyway.

docs/configuration.rst
```diff
@@ -186,6 +187,24 @@ Toolbar options
...
+
+  The DatabaseStore provides persistence and automatically cleans up old
+  entries based on the ``RESULTS_CACHE_SIZE`` setting.
+
+  Note: For full functionality, DatabaseStore requires migrations for
+  the debug_toolbar app:
+
+  .. code-block:: bash
+
+      python manage.py migrate debug_toolbar
```
dr-rompecabezas, line 203: My intuition is not to mention running migrations with `python manage.py migrate debug_toolbar`, but I left this here just in case you think it doesn't hurt. Django will recognize the migration hasn't run and throw a warning when running the development server.

matthiask (approved):
I like this a lot. Sorry for the nitpicky review comments.

docs/configuration.rst
matthiask: If we're adding models (and we are!) mentioning `migrate` is the thing to do, I think.

debug_toolbar/models.py
matthiask: @dr-rompecabezas I only see the emoji, I hope the confusion isn't too strong.

`DebugToolbarEntry` doesn't really tell a story about the job of this model, while `HistoryEntry` somewhat does -- since it's associated with the debug toolbar's `HistoryPanel`.

I really don't want to make it hard for you, feel free to engage with me on comments you disagree with or where you're unsure.

dr-rompecabezas: I like `HistoryEntry` but I also I do not dislike `DebugToolbarEntry`. This is a hard model for me to name. Unlike the model for a blog or just about any model I have created before with a clear business purpose, the naming convention is less obvious here to me. This is almost like a _metamodel_ of sorts that we don't even add to the admin (should we?). 

The smiley is because I had never heard of bikeshedding or Parkinson's law of triviality. Made me smile when I read the article. 

So, this is a change I am happy to make, and will do so. I trust and respect your experience and intuition!

debug_toolbar/models.py
dr-rompecabezas: Yes, missed that. I also prefer lowercase.

debug_toolbar/store.py
dr-rompecabezas: True.

debug_toolbar/store.py
dr-rompecabezas: Clearly an oversight on my part. I'll fix this. Thank you.

debug_toolbar/store.py
dr-rompecabezas: I see the option to commit the change here, which is convenient, but I'll just make the change locally together with the rest of edits and push again.

debug_toolbar/store.py
dr-rompecabezas: I'll push the rest of changes for now, while this is being discussed. 

For reference, here's our the `save_panel` method was implemented in MemoryStore, including the `set` method call.

```python
    @classmethod
    def save_panel(cls, request_id: str, panel_id: str, data: Any = None):
        """Save the panel data for the given request_id"""
        cls.set(request_id)
        cls._request_store[request_id][panel_id] = serialize(data)
```

We can also call `_cleanup_old_entries` directly rather than `set` to call `_cleanup_old_entries`, if I understand the purpose correctly.

debug_toolbar/models.py
dr-rompecabezas: Thinking about this more deeply now, I remember I have in the past changed model names to avoid repetition in the string representation of app_model. I now strongly prefer not to include DebugToolbar in the model name and align with your view. Great suggestion. This change was very easy to make.

dr-rompecabezas: Oh, @matthiask now I see that the emoji is a frowny face! I meant the opposite (it looks so tiny on my screen I couldn't tell until I opened it in VS Code). Sorry! Thanks again for the great feedback.

matthiask:
By the way, you could try pinning django-csp<4 in this branch again, this would probably allow the tests to pass. And you don't have to worry about merging main for now.

dr-rompecabezas:
Good. I'll try pinning CSP. That way, I can also rest assured that's the only cause of the failing tests.

dr-rompecabezas:
I'll look into the failing (MySQL) tests later today. Good that we pinned django-csp. Otherwise, I would have assumed that was the only reason for the failing tests.

matthiask:
The MariaDB failure has to do with using `LIMIT` and subqueries. It seems my suggestion to use `.exclude(...[:25]).delete()` broke the MariaDB test. It seems there's a reason for using `list()` after all.

debug_toolbar/store.py
tim-schilling: `cls.set()` is meant to create the record for the request in the toolbar's store. For the `MemoryStore`, there's a separate collection of ids from the actual store. I suspect it's for performance more than anything.

debug_toolbar/store.py
```diff
@@ -140,5 +141,91 @@ def panels(cls, request_id: str) -> Any:
...
+        Enforce the cache size limit - keeping only the most recently used entries
+        up to RESULTS_CACHE_SIZE.
+        """
+        # Get the cache size limit from settings
+        cache_size = dt_settings.get_config()["RESULTS_CACHE_SIZE"]
+
+        # Determine which entries to keep (the most recent ones up to cache_size)
+        keep_ids = list(
+            HistoryEntry.objects.all()[:cache_size].values_list("request_id", flat=True)
+        )
```
tim-schilling, line 157:
```suggestion
        # Determine which entries to keep
        keep_ids = cls.request_ids()
```

There's some code we can re-use. This should reduce what the test needs to do.

debug_toolbar/store.py
```diff
@@ -140,5 +141,91 @@ def panels(cls, request_id: str) -> Any:
...
+        return HistoryEntry.objects.filter(request_id=request_id).exists()
+
+    @classmethod
+    def set(cls, request_id: str):
+        """Set a request_id in the store and clean up old entries"""
+        # Create the entry if it doesn't exist (ignore otherwise)
+        HistoryEntry.objects.get_or_create(request_id=request_id)
+
+        # Enforce the cache size limit to clean up old entries
+        cls._cleanup_old_entries()
```
tim-schilling, line 183:
```suggestion
        with transaction.atomic():
            # Create the entry if it doesn't exist (ignore otherwise)
            _, created = HistoryEntry.objects.get_or_create(request_id=request_id)
            if created:
                # Enforce the cache size limit to clean up old entries
                cls._cleanup_old_entries()
```
This does two things. It wraps the operation in a transaction so that we only drop create and drop old entries if everything works as expected. That may be overkill though.

The other part is that we only cleanup old entries if we're adding an entry. It reduces two queries.

debug_toolbar/store.py
```diff
@@ -140,5 +141,91 @@ def panels(cls, request_id: str) -> Any:
...
+        """Remove all requests from the store"""
+        HistoryEntry.objects.all().delete()
+
+    @classmethod
+    def delete(cls, request_id: str):
+        """Delete the stored request for the given request_id"""
+        HistoryEntry.objects.filter(request_id=request_id).delete()
+
+    @classmethod
+    def save_panel(cls, request_id: str, panel_id: str, data: Any = None):
```
tim-schilling, line 196: Similar to `set()`, do we want to use `transaction.atomic()` here as well?

debug_toolbar/store.py
dr-rompecabezas: The transaction.atomic() wrapper seems overkill to me but I see the point about running cleanup only when new records are created, which we could do in the set method:


```python
    @classmethod
    def set(cls, request_id: str):
        """Set a request_id in the store and clean up old entries"""
        # Create the entry if it doesn't exist (ignore otherwise)
        _, created = HistoryEntry.objects.get_or_create(request_id=request_id)

        # Only enforce cache size limit when new entries are created
        if created:
            cls._cleanup_old_entries()
```

What do you think?

dr-rompecabezas: Hmmm... this comment is showing under save_panel(), but looking at your next comment, I wonder if you were commenting on the set() method.

dr-rompecabezas:
I applied some of the changes suggested by @tim-schilling without attempting to simplify tests yet to see if all tests pass in CI.

matthiask:
@tim-schilling I'm good with this, and I hope we can get `serialized` merged soon? Maybe I should do a quick release containing everything else first.

debug_toolbar/store.py
tim-schilling: I meant this comment for both methods.

tim-schilling: As per our call, there's a concern about panels updating this concurrently and needing to avoid clearing another panel's data when saving.

debug_toolbar/store.py
tim-schilling: I agree @matthiask's original statement. I'm going to adjust this accordingly.

tim-schilling (approved):

PR #2128, opened 2025-04-26
Keep pannel close button accessible with external styles

### LINKED ISSUE #2083: Panel close button can be hidden behind the panel title
When using DjDT on a local project I found that the close button was not working. After digging into it I found the stylesheet I was using for the project ([missing.css](https://missing.style/)) was interfering with the positioning on the panel title such that it now fully overlaps the close button and I couldn't click through to the button beneath the heading.

Fixing was as simple as swapping two lines in the template so that the button is rendered after the title h3 element. Is this something that can be included upstream? I didn't look into anything else like messing with z-index, as this just seemed the more natural way to write it.

```diff
 {% if panel.has_content and panel.enabled %}
   <div id="{{ panel.panel_id }}" class="djdt-panelContent djdt-hidden">
     <div class="djDebugPanelTitle">
-      <button type="button" class="djDebugClose">×</button>
       <h3>{{ panel.title }}</h3>
+      <button type="button" class="djDebugClose">×</button>
     </div>
     <div class="djDebugPanelContent">
       {% if toolbar.should_render_panels %}
```

TomHall2020:
#### Description

External styles such as [missing.css](https://missing.style/) can interfere with the display of the panel header elements, and since these are currently rendered after the button they can appear ontop and block click events from passing through. Fix was fairly straightforward, have just swapped the order of display everwhere I could find these pairs of elements.

Fixes #2083

While the selenium tests are all working, I haven't quite figured out how I could incorporate a regression test for this. I've narrowed the offending css down to this rule:

```css
h1,h2,h3,h4,h5,h6 {
  position:relative;
} 
```

And I've tested manually that this will break the close button before this PR and not after, but I can't quite see how to approach setting up a test case for this. The test cases I've edited do at least assert that the content of the button is rendered after the headers, so it is indirectly tested for, perhaps that is suffient?

#### Checklist:

- [x] I have added the relevant tests for this change (edited existing tests)

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/utils.js
+++ b/debug_toolbar/static/debug_toolbar/js/utils.js
@@ -90,7 +90,7 @@ function ajax(url, init) {
         })
         .catch((error) => {
             const win = document.getElementById("djDebugWindow");
-            win.innerHTML = `<div class="djDebugPanelTitle"><button type="button" class="djDebugClose">»</button><h3>${error.message}</h3></div>`;
+            win.innerHTML = `<div class="djDebugPanelTitle"><h3>${error.message}</h3><button type="button" class="djDebugClose">»</button></div>`;
             $$.show(win);
             throw error;
         });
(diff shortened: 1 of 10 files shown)


matthiask:
Thank you!

PR #2132, opened 2025-05-02
Remove pin for django-csp.

### LINKED ISSUE #2129: requirements_dev.txt and tox.ini do not agree with each other
> Bit late to catch this here, but when I was trying to follow the contributing guide and use `make test` this was breaking on me as `django-template-partials` was only added to the tox config, and not to `requirements_dev.txt` as well

_Originally posted by @TomHall2020 in https://github.com/django-commons/django-debug-toolbar/issues/2117#issuecomment-2832317527_
            

I also noticed that the requirements file still specifies django-csp<4.

tim-schilling:
The toolbar now supports django-csp v4

#### Description

this information as well.

Fixes #2129

#### Checklist:

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -11,6 +11,7 @@ html5lib
 selenium
 tox
 black
+django-template-partials
 django-csp # Used in tests/test_csp_rendering
 
 # Integration support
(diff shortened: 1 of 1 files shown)


matthiask:
Thanks!

PR #2137, opened 2025-05-14
Added check for pytest as test runner for IS_RUNNING_TESTS.

### LINKED ISSUE #2136: `IS_RUNNING_TESTS` config should also detect Pytest runner
One may presume that `pytest` so popular that it is used with Django projects as often as the built-in `manage.py test` framework (usually including plugins, e.g. `pytest-django`). However, the `IS_RUNNING_TESTS` setting detects only the latter by default by looking for `"test"` in `sys.argv`.

Updating:

https://github.com/django-commons/django-debug-toolbar/blob/cf71ded725ddda6124e55762cc2115567d6eec4c/debug_toolbar/settings.py#L46

by:

```python
    "IS_RUNNING_TESTS": ("test" in sys.argv) or "PYTEST_VERSION" in os.environ,
```

NOTE: Using `PYTEST_VERSION` variable is a recommended way for determining whether `pytest` is running; see https://github.com/pytest-dev/pytest/issues/9502#issuecomment-2063572916.

tim-schilling:
Fixes #2136

I could not find a way to reasonably test this in our test suite.

#### Checklist:

### DIFF
--- a/debug_toolbar/settings.py
+++ b/debug_toolbar/settings.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import warnings
 from functools import cache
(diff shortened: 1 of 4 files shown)


paduszyk:
@tim-schilling You'd like to update:

https://github.com/django-commons/django-debug-toolbar/blob/cf71ded725ddda6124e55762cc2115567d6eec4c/tests/test_checks.py#L243

or write a new test?

tim-schilling:
@paduszyk that test manipulates `IS_RUNNING_TEST`'s value, but this change is how that value gets determined by default.

matthiask:
Thanks! Makes sense.

PR #2139, opened 2025-05-16
Disabled document.cookie linter

tim-schilling:
The replacement, CookieStore isn't accessible on all browsers at this time, nor should we assume we always have access to secure environments.

#### Checklist:


matthiask (approved):
Thank you!

Issue #2087, opened 2025-02-26
Remove typing from CspRenderingTestCase


tim-schilling:
I'd be in favor of removing the typing on `CspRenderingTestCase`. I'm not in favor of utilizing type hinting to that extent.

I know we have an issue to bring typing to the library, but I'm not sure I'm in agreement with that any more. The type hinting below seems to only be beneficial to a particular developer set up that we don't have documented anywhere. I think that inconsistency isn't worth it.

```python
response = cast(HttpResponse, self.client.get(path="/regular/basic/"))
self.assertEqual(response.status_code, 200)

html_root: Element = self.parser.parse(stream=response.content)
self._fail_on_invalid_html(content=response.content, parser=self.parser)
self.assertContains(response, "djDebug")

namespaces = get_namespaces(element=html_root)
context: ContextList = response.context  # pyright: ignore[reportAttributeAccessIssue]
```

matthiask:
I'm not profiting from typing at all currently but I also think the Python world is slowly moving towards it. 

If the types keep us from fixing issues we have I agree that removing them is the right thing to do. 

Maybe my stance will change once we get the type checker integrated into ruff, I don't know. I do expect that one to make a difference.

tim-schilling:
> If the types keep us from fixing issues we have I agree that removing them is the right thing to do.

It's not getting in the way, but I can't maintain the typing code for that because I'm not set up for it. It's not something I'm interested in investing a significant amount of time into right now either.

matthiask:
> It's not something I'm interested in investing a significant amount of time into right now either.

Me neither! It seems we really should be the ones profiting from the typing effort. Since we're not I now totally agree with the removal.

karolyi:
Just my 2 cents: if it doesn't bother your environment, why would you lessen someone else's experience with the code, when it comes to writing tests? If nobody forces you to write typed code, why would you force someone else to write *untyped* code?

For the record: I'm using basedpyright and I've learned how to write code immensely more stable just by utilizing proper typing. Yes it is a hassle at first but totally worth it on the long run. Maybe it's the same reason why python is moving towards it.

Feel free to do as you like, I just wanted to put in an opposing opinion.

tim-schilling:
Thank you for your opinion. The earlier discussion should cover the questions you asked. When I make the time to invest into heavily typed python, I'm sure I'll rue this decision.

JohananOppongAmoateng:
@tim-schilling i think this issue should be closed

PR #2144, opened 2025-06-14
Remove Type Hints from CspRenderingTestCase

andoriyaprashant:
#### Description

This PR removes explicit type annotations and unnecessary imports from the CspRenderingTestCase class as suggested in #2087 

- Changes made:
   - Removed all type hints within the CspRenderingTestCase methods (ex - root: Element, namespaces: dict[str, str], etc.).
   - Removed unused cast, Element, and ContextList imports. 
   - Preserved all logic and test behavior - only typing-related changes were made.

Fixes # 2087

#### Checklist:


matthiask (approved):
Slightly sad, but I have to say it looks less noisy. The annotations for basic functions look good, but `cast` really doesn't.

PR #2146, opened 2025-06-18
Add .prof File Download Support to Profiling Panel

### LINKED ISSUE #1798: Download profiling results
The `.prof` files that cProfile generates can be used with interactive programs or to create helpful graphs. If you could download the raw `.prof` file, users could take advantage of the toolbar even if they prefer other visualization methods.

andoriyaprashant:
#### Description

This PR adds a feature to download the raw .prof file generated by cProfile via the Django Debug Toolbar profiling panel.

Users can now export the profiling results to visualize or analyze them externally.

- Added download_prof_file view to serve .prof files
- Updated profiling.html to include a download button
- Included prof_file_path in the panel stats to pass to the template

Status: Draft - encountering a NoReverseMatch error related to debug_toolbar_download_prof_file. Likely an issue with urls.py configuration. Seeking guidance on the correct integration approach.

Fixes #1798 

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/profiling.py
+++ b/debug_toolbar/panels/profiling.py
@@ -2,6 +2,7 @@ import cProfile
 import os
 from colorsys import hsv_to_rgb
 from pstats import Stats
+import tempfile
 
 from django.conf import settings
 from django.utils.html import format_html
@@ -168,8 +169,11 @@ class ProfilingPanel(Panel):
         self.stats = Stats(self.profiler)
         self.stats.calc_callees()
 
-        root_func = cProfile.label(super().process_request.__code__)
+        prof_file_path = os.path.join(tempfile.gettempdir(), next(tempfile._get_candidate_names()) + ".prof")
+        self.profiler.dump_stats(prof_file_path)
+        self.prof_file_path = prof_file_path 
 
+        root_func = cProfile.label(super().process_request.__code__)
         if root_func in self.stats.stats:
             root = FunctionCall(self.stats, root_func, depth=0)
             func_list = []
@@ -182,4 +186,9 @@ class ProfilingPanel(Panel):
                 dt_settings.get_config()["PROFILER_MAX_DEPTH"],
                 cum_time_threshold,
             )
-            self.record_stats({"func_list": func_list})
+            self.record_stats({
+                "func_list": func_list,
+                "prof_file_path": self.prof_file_path 
+            })
+
+
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -142,7 +142,16 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
         # process during the .last_executed_query() call.
         self.db._djdt_logger = None
         try:
-            return self.db.ops.last_executed_query(self.cursor, sql, params)
+            # Handle executemany: take the first set of parameters for formatting
+            if isinstance(params, (list, tuple)) and len(params) > 0 and isinstance(params[0], (list, tuple)):
+                sample_params = params[0]
+            else:
+                sample_params = params
+
+            try:
+                return self.db.ops.last_executed_query(self.cursor, sql, sample_params)
+            except Exception:
+                return sql
         finally:
             self.db._djdt_logger = self.logger
 
--- a/debug_toolbar/templates/debug_toolbar/panels/profiling.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/profiling.html
@@ -1,4 +1,13 @@
 {% load i18n %}
+
+{% if prof_file_path %}
+  <div style="margin-bottom: 10px;">
+    <a href="{% url 'debug_toolbar_download_prof_file' %}?path={{ prof_file_path|urlencode }}" class="djDebugButton">
+    Download .prof file
+  </a>
+  </div>
+{% endif %}
+
 <table>
   <thead>
     <tr>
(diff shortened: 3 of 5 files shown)


andoriyaprashant:
Hey @matthiask @tim-schilling

I’ve started working on the enhancement Download profiling results (Issue #1798) and have implemented the backend and template changes to support .prof file download. However, I’m currently getting the following error:

```bash
django.urls.exceptions.NoReverseMatch: Reverse for 'debug_toolbar_download_prof_file' not found.
```
I’ve defined the download_prof_file view in views.py, but I suspect the issue might be with how I registered it in urls.py or the namespace being used by the toolbar.

Since this involves integrating a custom view into the debug toolbar's internal URLs, I’ve kept the PR in draft so you could please review my approach and suggest how I can resolve this.

Thank you!

debug_toolbar/panels/profiling.py
```diff
@@ -168,8 +169,13 @@ def generate_stats(self, request, response):
         self.stats = Stats(self.profiler)
         self.stats.calc_callees()
 
-        root_func = cProfile.label(super().process_request.__code__)
+        prof_file_path = os.path.join(
+            tempfile.gettempdir(), next(tempfile._get_candidate_names()) + ".prof"
```
matthiask, line 173: `_get_candidate_names()` isn't documented? I think we should only enable this feature when users provide a path where we can store profiles, because we have to ensure we're only serving files from this very path and not arbitrary files.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -142,7 +142,20 @@ def _last_executed_query(self, sql, params):
...
+                and isinstance(params[0], (list, tuple))
+            ):
+                sample_params = params[0]
+            else:
+                sample_params = params
+
+            try:
+                return self.db.ops.last_executed_query(self.cursor, sql, sample_params)
+            except Exception:
+                return sql
```
matthiask, line 158: This looks like an unrelated change?

debug_toolbar/templates/debug_toolbar/panels/profiling.html
```diff
@@ -1,4 +1,13 @@
 {% load i18n %}
+
+{% if prof_file_path %}
+  <div style="margin-bottom: 10px;">
```
matthiask, line 4: We do not have any inline styles in templates until now, if we need styles add them to the `toolbar.css` file.

debug_toolbar/urls.py
```diff
@@ -1,5 +1,14 @@
...
 app_name = APP_NAME
-urlpatterns = DebugToolbar.get_urls()
+
+urlpatterns = DebugToolbar.get_urls() + [
+    path(
+        "download_prof_file/",
+        debug_toolbar_views.download_prof_file,
+        name="debug_toolbar_download_prof_file",
+    ),
+]
```
matthiask, line 14: You should add the URL pattern to `get_urls()`, not here.

debug_toolbar/templates/debug_toolbar/panels/profiling.html
```diff
@@ -1,4 +1,13 @@
...
+    <a href="{% url 'debug_toolbar_download_prof_file' %}?path={{ prof_file_path|urlencode }}" class="djDebugButton">
```
matthiask, line 5: The toolbar uses the `djdt:` namespace for URLs; search the code base for `reverse(` calls to see how that can be used. (See `git grep -C3 'reverse('`)

debug_toolbar/views.py
```diff
@@ -25,3 +28,20 @@ def render_panel(request):
         content = panel.content
         scripts = panel.scripts
     return JsonResponse({"content": content, "scripts": scripts})
+
+
+@require_GET
+def download_prof_file(request):
+    file_path = request.GET.get("path")
+    print("Serving .prof file:", file_path)
+    if not file_path or not os.path.exists(file_path):
```
matthiask, line 37: This can trivially be exploited by using something like

`...?path=/etc/passwd`

This view serves arbitrary files which are readable by the webserver process: Either you have to use signing (`django.core.signing`, also used in the templates panel) to ensure that paths cannot be tampered with and/or use a known root where you put files, and ensure that you only serve files from this root. This isn't totally straightforward because you have to protect against path traversal attacks.

matthiask (commented):
Can you explain more what we can do with the `.prof` file?

Also, I wonder if we shouldn't just write profile files always to a configured path (in the settings) and not make this a part of the interface? (Not saying that's a better idea, just thinking out loud.)

tim-schilling:
@andoriyaprashant do you have time to pick this back up? If not, @robhudson could ask a Djangonaut from the next session of Djangonaut Space to get it over the finish line.

andoriyaprashant:
> @andoriyaprashant do you have time to pick this back up? If not, @robhudson could ask a Djangonaut from the next session of Djangonaut Space to get it over the finish line.

Hey @tim-schilling Thanks for checking in! At the moment I’m unable to give this PR the focus it needs, so I’m fine with someone else picking it up and taking it forward 

Thank you

JohananOppongAmoateng:
@tim-schilling I would like to pick this up

tim-schilling:
Following up here. We're going to hold this in reserve for a few more weeks for Djangonaut Space.

tim-schilling:
@JohananOppongAmoateng please feel free to work on this

JohananOppongAmoateng:
@tim-schilling noted

tim-schilling:
Replaced by #2286

PR #2153, opened 2025-07-08
Add show_toolbar_with_docker function for Docker IP handling

matthiask:
- Introduced a new function, show_toolbar_with_docker, to determine if the toolbar should be displayed when running inside Docker containers.
- Updated installation documentation to reference the new function for Docker configurations.
- Added a changelog entry for the new functionality.

Co-authored-by: Matthias Kestenholz <mk@feinheit.ch>


matthiask:
Thank you :)

PR #2155, opened 2025-07-15
Deduplicate staticfiles

matthiask:
#### Description

Currently, the staticfiles panel contains one entry per `static()` call. That's not the same thing as used staticfiles, since assets can be deduplicated in multiple ways, e.g. through `forms.Media` or because the browser actually understands that it should load `<script type="module">` once only.

I think it makes more sense to show a deduplicated and sorted list of static files used.

Refs https://github.com/matthiask/django-prose-editor/issues/48

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -1,6 +1,7 @@
 import contextlib
 import uuid
 from contextvars import ContextVar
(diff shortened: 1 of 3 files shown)


tim-schilling (approved):
Makes sense and looks good!

matthiask:
Thanks!

PR #2157, opened 2025-07-22
Version 6.0.0

tim-schilling:
#### Description

Release Version 6.0.0 of the toolbar. Now including the serializable changes.

#### Checklist:


tim-schilling:
@matthiask I haven't been running my work project off main, so I'll do that tomorrow. If that looks good, I'm inclined to release this.

tim-schilling:
@django-commons/django-debug-toolbar If anyone has been testing off main (not from the latest in pypi), please let us know.

matthiask:
Yes, that's great! I'm not approving yet because I want to do the same as you and run it in a project or two of mine, but feel free to move ahead with this if you get to it faster than I do.

matthiask:
I have submitted PRs to fix the bugs, but I haven't tested the `DatabaseStore` yet at all.

tim-schilling:
I just ran some tests with `DatabaseStore`.
- https://github.com/django-commons/django-debug-toolbar/issues/2165 I think we could put this off to a future version. Most folks probably aren't going to use this and if they are, they've specifically configured it so it shouldn't be a total surprise. :shrug:
- https://github.com/django-commons/django-debug-toolbar/issues/2166 This is ancillary. We don't have to fix it now, but at some point we should.

I feel much better about releasing this into the wild at this point.

matthiask:
> I feel much better about releasing this into the wild at this point.

Yes, I agree!

matthiask (approved):

PR #2162, opened 2025-07-22
Fixes #2160 -- Corrected links to static files

### LINKED ISSUE #2160: Staticfiles in TemplatesPanel aren't allowing viewing static file content
The static files panel is rendering the used static files with empty `href=""` attributes:

Image

When functioning properly, it should look like:

Image

#### Suggested solution:

`staticfiles.py`
```python
    self.record_stats(
        {
            "num_found": self.num_found,
            "num_used": len(self.used_paths),
            # We can't store a dataclass, so we should convert it to a dictionary when it gets stored.
            "staticfiles": [{**vars(staticfile), "real_path": staticfile.real_path()} for staticfile in sorted(self.used_paths)],
            "staticfiles_apps": self.get_staticfiles_apps(),
            "staticfiles_dirs": self.get_staticfiles_dirs(),
            "staticfiles_finders": self.get_staticfiles_finders(),
        }
    )
```

`staticfiles.html`
```django
<h4>{% blocktranslate count staticfiles_count=staticfiles|length %}Static file{% plural %}Static files{% endblocktranslate %}</h4>
{% if staticfiles %}
  <dl>
    {% for staticfile in staticfiles %}
      {# Ideally we'd use remoteCall to inject the content, but for now let's open the content in a separate tab/window #}
      {# We also need to explicitly render the path as we want to see it. Not use the __str__ method #}
      <dt><strong><a class="toggleTemplate" target="_blank" href="{{ staticfile.url }}">{{ staticfile.path }}</a></strong></dt>
      <dd><samp>{{ staticfile.real_path }}</samp></dd>
    {% endfor %}
  </dl>
{% else %}
  <p>{% translate "None" %}</p>
{% endif %}
```

matthiask:
#### Description

The JSON serialization used by the new serializable mechanism compressed the `StaticFile` dataclass down to a single string without an `.url` attribute. Change the code to use a tuple instead to avoid this problem.

Fixes #2160 

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -1,7 +1,6 @@
 import contextlib
 import uuid
 from contextvars import ContextVar
-from dataclasses import dataclass
 from os.path import join, normpath
 
 from django.contrib.staticfiles import finders, storage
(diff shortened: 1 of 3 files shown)


tim-schilling (approved):
Thanks for cleaning this up!

PR #2163, opened 2025-07-22
Fixes 2151 -- Forwarded args and kwargs in URLMixin.url

### LINKED ISSUE #2151: URLMixin is missing `force` argument
It looks like that the [`force` argument](https://github.com/django/django/blob/main/django/contrib/staticfiles/storage.py#L201) is not passed through in [URLMixin.url](https://github.com/django-commons/django-debug-toolbar/blob/c217334010fa54a8726640519ca29cb77fffda58/debug_toolbar/panels/staticfiles.py#L39), causing the panel staticfiles panel to crash when the caller is passing it in.

matthiask:
#### Description

Some static file storages have additional arguments to `.url()`, for example to return a non-hashed URL in debug mode, see https://github.com/django/django/blob/a8536e33da6365fb7bba901a06a67d163d0d4c8a/django/contrib/staticfiles/storage.py#L201

I think this change doesn't strictly require tests, but I'm happy to postpone this if we don't want to do this without tests.

Fixes #2151 

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/staticfiles.py
+++ b/debug_toolbar/panels/staticfiles.py
@@ -16,8 +16,8 @@ record_static_file_signal = Signal()
 
 
 class URLMixin:
-    def url(self, path):
-        url = super().url(path)
+    def url(self, path, *args, **kwargs):
+        url = super().url(path, *args, **kwargs)
         with contextlib.suppress(LookupError):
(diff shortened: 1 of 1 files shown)


matthiask:
Great, thanks!

PR #2164, opened 2025-07-22
Fixes #2159 -- Do not HTML-escape traces in the cache and profiling panel

### LINKED ISSUE #2159: Profiling panel is rendering the html as text
It appears that the profiling panel's content is rendering the html as text and isn't being treated as html.

#### Reproducing

1. Go to the admin
2. Enable profiling panel
3. Refresh the page
4. Access profiling panel

Image

matthiask:
#### Description

The panel data has to be explicitly reloaded from the store to hit this case. The new ``reload_stats`` utility helps with this.

Fixes #2159

#### Checklist:

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/panels/cache.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/cache.html
@@ -61,7 +61,7 @@
         </tr>
         <tr class="djUnselected djToggleDetails_{{ forloop.counter }}" id="cacheDetails_{{ forloop.counter }}">
           <td colspan="1"></td>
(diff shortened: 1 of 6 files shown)


matthiask:
It's really annoying. The tests pass both with and without the `|safe`, but the code obviously doesn't do the right thing when removing `|safe`.

I have traced the code through the data serialization and it seems to me that the traces are properly serialized. So, I don't understand why the test doesn't show the problem...

matthiask:
I have found the problem. The panel data has to be forcibly reloaded from the store before rendering, otherwise the `SafeString` objects are still around.

tim-schilling (approved):
Thank you for tracking this one down!

PR #2169, opened 2025-07-27
Binary parameter handling for GeoDjango

### LINKED ISSUE #423: Error debugging GeoDjango queries

eduzen:
#### Description

hey folks, this is my very first PR in the django-commons. I did this because I faced the issue myself in one project that I'm working on. I don't know how I can test this in that project. So I will wait for some of you to tell me how I can do that... but in the meantime this is the fix for the SQL Explain functionality for GeoDjango queries with binary parameters.

Fixes #423 

SQL Explain feature throws a 500 error when used with GeoDjango queries containing binary parameters. This affects any query using PostGIS functions like ST_GeomFromEWKB(), ST_Distance_Sphere(), etc.
I'm trying to create a robust binary parameter handling system:

  1. Enhanced Parameter Encoding (tracking.py)

  - Detects binary data (bytes, bytearray) during parameter logging
  - Encodes binary data as base64 with a special marker: {"__djdt_binary__": "base64_data"}
  - Preserves exact binary integrity through JSON serialization

  2. Parameter Reconstruction (forms.py)

  - Added _reconstruct_params() function to decode stored parameters
  - Recursively handles nested structures (lists, dicts) containing binary data
  - Restores original binary data from base64 encoding

  3. Updated SQL Operations

  - Modified select(), explain(), and profile() methods to use parameter reconstruction
  - Added helper method _get_query_params() to eliminate code duplication
  
#### Checklist:


debug_toolbar/panels/sql/forms.py
```diff
@@ -69,10 +87,15 @@ def clean(self):
         cleaned_data["query"] = query
         return cleaned_data
 
+    def _get_query_params(self):
+        """Get reconstructed parameters for the current query"""
+        query = self.cleaned_data["query"]
+        return _reconstruct_params(json.loads(query["params"]))
```
tim-schilling: Do you think this would be improved by making use of the [`json.loads(..., cls)` argument](https://docs.python.org/3/library/json.html#json.load)?

eduzen: If I understand you correctly, @tim-schilling, you're suggesting that we create our own DebugToolbarJSONDecoder and use it like this: json.loads(query["params"], cls=DebugToolbarJSONDecoder), so we can encapsulate all the base64 logic inside of `_reconstruct_params` in custom JSON decoder class?

eduzen:
@tim-schilling I added some commits following what you  were pointing out, let me know what do you think!

debug_toolbar/panels/sql/decoders.py
```diff
@@ -0,0 +1,26 @@
...
+        elif isinstance(params, dict):
+            if "__djdt_binary__" in params:
+                return base64.b64decode(params["__djdt_binary__"])
+            else:
+                return {
+                    key: self._reconstruct_params(value)
+                    for key, value in params.items()
+                }
+        else:
+            return params
```
tim-schilling:
```suggestion
        match params:
            case list():
                return [self._reconstruct_params(param) for param in params]
            case dict() if "__djdt_binary__" in params:
                return base64.b64decode(params["__djdt_binary__"])
            case dict():
                return {
                    key: self._reconstruct_params(value)
                    for key, value in params.items()
                }
            case _:
                return params
```
We don't need to use it, but match-case is an option here. What do you think @eduzen?

tests/panels/test_sql_geodjango_fix.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class MockLogger:
+    """Mock logger for testing"""
+
+    def record(self, **kwargs):
+        pass
+
+
+class TestCursor(NormalCursorMixin):
```
tim-schilling:
```suggestion
class StubCursor(NormalCursorMixin):
```

[As a heads up, these types of classes are called "stubs".](https://stackoverflow.com/a/463305/1637351)

debug_toolbar/panels/sql/decoders.py
```diff
@@ -0,0 +1,26 @@
+import base64
+import json
+
+
+class DebugToolbarJSONDecoder(json.JSONDecoder):
```
tim-schilling: Rather than the approach you've taken here, you may want to consider going with something like:

```python
import base64
import json

def object_hook(obj):
    for key, value in obj.items():
        if isinstance(value, str) and value.startswith("__djdt_binary__"):
            _, encoded = value.split("__djdt_binary__", maxsplit=1)
            obj[key] = base64.b64decode(encoded)
    return obj


class DebugToolbarJSONDecoder(json.JSONDecoder):
    """Custom JSON decoder that reconstructs binary data during parsing."""
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        if self.object_hook is None:
            self.object_hook = object_hook
```

This will require you to change the tests. The decoding will only work if there's an actual object being passed in. Most of the tests are passing in a list of values. Using an object/dictionary is more representative of what will actually be used since this is for parameters for a SQL query which is a dictionary.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -126,6 +127,11 @@ def _decode(self, param):
         if isinstance(param, dict):
             return {key: self._decode(value) for key, value in param.items()}
 
+        # Handle binary data (e.g., GeoDjango EWKB geometry data)
+        if isinstance(param, (bytes, bytearray)):
+            # Mark as binary data for later reconstruction
+            return {"__djdt_binary__": base64.b64encode(param).decode("ascii")}
```
tim-schilling: We should move this logic into a shared utility that way we don't have the magic value `"__djdt_binary__"` floating around the code-base. Plus it would be clearer how the logic gets paired together.

I'm also curious if we should be using this with the Store logic. If we do, then we could potentially ignore this and handle it within `DebugToolbarJSONEncoder` and then have `DebugToolbarJSONDecoder` as above. What do you think?

tests/panels/test_sql_geodjango_fix.py
```diff
@@ -0,0 +1,137 @@
...
+
+        expected_b64 = base64.b64encode(binary_data).decode("ascii")
+        self.assertEqual(encoded["__djdt_binary__"], expected_b64)
+
+        json_params = json.dumps([encoded])
+        reconstructed = json.loads(json_params, cls=DebugToolbarJSONDecoder)
+
+        self.assertEqual(len(reconstructed), 1)
+        self.assertEqual(reconstructed[0], binary_data)
+        self.assertIsInstance(reconstructed[0], bytes)
```
tim-schilling: I don't think there's a benefit to being this verbose. Try to be more concise such as:

```suggestion
        self.assertEqual(reconstructed, [binary_data])
```

tim-schilling (changes requested):
Thank you for continuing to work on this! I have a few directional changes. If you want more specific feedback, I can provide it. I didn't want to be overly prescriptive.

debug_toolbar/panels/sql/decoders.py
eduzen: Oh, I got it, I did not know about this `object_hook` possibility.

debug_toolbar/panels/sql/tracking.py
eduzen: Thank you for the feedback @tim-schilling ! I think an encoders.py or moving even everything into utils.py is a more consistent way.
About your second question, what do you mean with using the store logic?

eduzen:
> Thank you for continuing to work on this! I have a few directional changes. If you want more specific feedback, I can provide it. I didn't want to be overly prescriptive.

No please! Tell me more, because I see that you are pointing to a deeper idea that I did not catch yet.

debug_toolbar/panels/sql/tracking.py
tim-schilling: The Store logic converts all the panels data into JSON. So this binary parameter would eventually be converted to JSON to be added to the Store. So if the Store handles binary data properly, it may remove the need for the SQL panel to deal with it itself. Does that make more sense?

tim-schilling:
@eduzen oh, I meant I was writing out the solution when I did the last round of review. It seemed like you may be interested in working through the issues to arrive at your own implementation and I didn't want to deprive you of the opportunity. So I should have said was, "if you'd like me to make these edits myself, please let me know."

eduzen:
Super clear @tim-schilling let me take another look and I let you know!

tim-schilling:
Hey @eduzen no rush on this! Just wanted to check-in to see how things are going. Take care!

eduzen:
@tim-schilling thank you for asking. Busy days 😅 let me see if I can take a look the upcoming days!

eduzen:
Hey @tim-schilling, I’m adding a page to the example project so I can see the issue and the fix with my own eyes (in addition to the tests). Do you think I should include that in this MR as well?

image

I’m still cooking the solution. While playing with the UI, I ran into a couple of issues that I’m working on now. Thank you!

eduzen:
@tim-schilling  I've addressed your feedback 🤞. I also updated is_select_query in `utils.py` because I noticed the view and explain buttons weren't working correctly for this query in the sample app.

debug_toolbar/panels/sql/forms.py
```diff
@@ -69,10 +69,15 @@ def clean(self):
         cleaned_data["query"] = query
         return cleaned_data
 
+    def _get_query_params(self):
+        """Get parameters for the current query."""
+        query = self.cleaned_data["query"]
+        return query["params"]
+
```
tim-schilling:
```suggestion
```

debug_toolbar/panels/sql/forms.py
```diff
@@ -69,10 +69,15 @@ def clean(self):
...
     def select(self):
         query = self.cleaned_data["query"]
         sql = query["raw_sql"]
-        params = json.loads(query["params"])
+        params = self._get_query_params()
```
tim-schilling:
```suggestion
        params = query["params"]
```

debug_toolbar/panels/sql/forms.py
```diff
@@ -82,7 +87,7 @@ def select(self):
     def explain(self):
         query = self.cleaned_data["query"]
         sql = query["raw_sql"]
-        params = json.loads(query["params"])
+        params = self._get_query_params()
```
tim-schilling:
```suggestion
        params = query["params"]
```

debug_toolbar/panels/sql/forms.py
```diff
@@ -101,7 +106,7 @@ def explain(self):
     def profile(self):
         query = self.cleaned_data["query"]
         sql = query["raw_sql"]
-        params = json.loads(query["params"])
+        params = self._get_query_params()
```
tim-schilling:
```suggestion
        params = query["params"]
```

docs/changes.rst
```diff
@@ -18,6 +18,10 @@ Pending
 * Added test to confirm Django's ``TestCase.assertNumQueries`` works.
 * Fixed string representation of values in settings panel.
 * Declared support for Django 6.0.
+* Fixed SQL Explain functionality for GeoDjango queries with binary parameters.
+  Binary data (such as EWKB geometry) is now properly handled through base64
+  encoding, preventing "parse error - invalid geometry" errors when using
+  Explain on spatial queries.
```
```suggestion
  explain on spatial queries.
```

tests/panels/test_sql_geodjango_fix.py
```diff
@@ -0,0 +1,97 @@
...
+is properly encoded and decoded through the Store's serialize/deserialize
+functions using DebugToolbarJSONEncoder and DebugToolbarJSONDecoder.
+"""
+
+from debug_toolbar.store import deserialize, serialize
+
+from ..base import BaseTestCase
+
+
+class GeoDjangoBinaryParameterTest(BaseTestCase):
```
tim-schilling: @eduzen I'm assuming you're using an LLM for the tests judging from the fact that this is in it's own file, the doc string comment, and the comments on very obvious lines of code `serialized = serialize(params)`. Can you revisit these tests and remove the redundant pieces? (Comments, extraneous tests, extraneous assertions). We haven't fully adopted the LLM approach of "let it test everything it can think of" yet on this project. Please also move these tests into the `test_sql.py` file.

tim-schilling (changes requested):
Thank you @eduzen for continuing on this. I finally got around to rereviewing. Most of my comments are minimal, though I pulled this down and the tests have a few failures yet.

```
FAIL: test_serialize_logs_on_failure (tests.test_store.SerializationTestCase.test_serialize_logs_on_failure)
  File "/home/schillingt/Projects/debug_toolbar/tests/test_store.py", line 18, in test_serialize_logs_on_failure
    self.assertEqual(
AssertionError: '{"hello": {"foo": {"__djdt_binary__": "YmFy"}}}' != '{"hello": {"foo": "bar"}}'
- {"hello": {"foo": {"__djdt_binary__": "YmFy"}}}
+ {"hello": {"foo": "bar"}}
```

tests/panels/test_sql_geodjango_fix.py
eduzen: Sure, Tim. Sorry for the delay. I was using indeed llms, for asking question for iterating and also for testing. But I also wrote code myself, test it.

eduzen: Let me know what do you think now 😄

tim-schilling:
@eduzen hey I finally got back around to this. It appears the tests are still failing (you can review the logs on GitHub). I also set this up with a project using GeoDjango and the explain button is still causing a 500 error. You can pull that project down and run it with docker and uv here: https://github.com/tim-schilling/geodjango_test

Please let me know if I'm misunderstanding something or holding it wrong.

eduzen:
Thank you for the support @tim-schilling but it's a bit hard for me to find the time for fixing these issues now. I

tim-schilling:
@eduzen That's understandable. Thank you for pushing this forward though!

Issue #1395, opened 2020-11-02
Colorize the minimized DDT icon


d9pouces:
I made an [extension](https://github.com/d9pouces/djt_nvu) to check the HTML validity. 
However, my DDT is minimized and I don't see the result.

Why not add a method `get_warning_level` (0/10/20/30/40/50/… like the `logging` module) to the `Panel` class?
The minimized icon could be colored following the maximum level (across all panels), being red if any panel is above 40, orange if any panel is above 30, and so on.

For example, the SQL panel could raise a warning if there are too many requests or if there are too slow or the Django panel could be red if the current Django version is not supported anymore.

hunzlahmalik:
Hi @matthiask , I'm starting to work on this feature now.

PR #2178, opened 2025-08-10
feat: Add colorization based on health level for panels and toolbar

hunzlahmalik:
- #1395 

## Description

This PR introduces a unified health/alert state for Django Debug Toolbar panels and updates the UI to reflect panel.
The name `health` is debatable, but the `alert` was already taken by the panel.

## Key Changes

  - Adds a `ToolbarHealthLevel` enum (`NONE`, `WARNING`, `CRITICAL`) for numeric health states.
  - Adds a `health_level` property to the base `Panel` class, allowing panels to report their health status.
  - Implements `health_level` in `AlertsPanel` and `SQLPanel` for alert and slow query detection.
  - The toolbar now computes the maximum health level across all panels and uses it to color the hidden toolbar handle.

- **UI Updates:**  
  - Panel buttons and the hidden toolbar handle use new CSS classes and colors based on health level.
  - Adds CSS variables and styles for warning and critical states.

| Warning      | Warning on Hover     | Critical | Critical on Hover
| :-------------: | :-------------: | :-------------: | :-------------: |
| image |  image | image | image |

| Critical on Panel     |  Warning on Panel  |
| :-------------: | :-------------: | 
| image |  image |

## Motivation
- Improves user experience by visually highlighting issues in the toolbar.

## Checklist:

- [ ] Add to Documentation

### DIFF
--- a/debug_toolbar/panels/__init__.py
+++ b/debug_toolbar/panels/__init__.py
@@ -3,7 +3,7 @@ from django.template.loader import render_to_string
 from django.utils.functional import classproperty
 
 from debug_toolbar import settings as dt_settings
-from debug_toolbar.utils import get_name_from_obj
+from debug_toolbar.utils import HealthLevel, get_name_from_obj
 
 
 class Panel:
@@ -128,6 +128,19 @@ class Panel:
         this event to support dynamic functionality.
         """
         return []
+
+    @property
+    def health_level(self):
+        """
+        Returns the health level of the panel as a `ToolbarHealthLevel` enum value.
+
+        This property is used by the toolbar to determine the overall health status of each panel.
+        The default implementation returns `ToolbarHealthLevel.NONE`, indicating no issues.
+
+        Subclasses should override this property to provide custom health logic, returning
+        `ToolbarHealthLevel.WARNING` or `ToolbarHealthLevel.ERROR` as appropriate based on panel-specific conditions.
+        """
+        return HealthLevel.NONE
 
     # Panel early initialization
 
--- a/debug_toolbar/panels/alerts.py
+++ b/debug_toolbar/panels/alerts.py
@@ -97,7 +97,7 @@ class AlertsPanel(Panel):
         """
         Return the health level of the panel based on the alerts.
         """
-        if not self.alerts:
+        if not self.get_stats().get("alerts"):
             return HealthLevel.NONE
 
         return HealthLevel.CRITICAL
--- a/debug_toolbar/panels/alerts.py
+++ b/debug_toolbar/panels/alerts.py
@@ -3,7 +3,7 @@ from html.parser import HTMLParser
 from django.utils.translation import gettext_lazy as _
 
 from debug_toolbar.panels import Panel
-from debug_toolbar.utils import is_processable_html_response
+from debug_toolbar.utils import HealthLevel, is_processable_html_response
 
 
 class FormParser(HTMLParser):
@@ -91,6 +91,16 @@ class AlertsPanel(Panel):
             return f"{len(alerts)} {alert_text}"
         else:
             return ""
+
+    @property
+    def health_level(self):
+        """
+        Return the health level of the panel based on the alerts.
+        """
+        if not self.alerts:
+            return HealthLevel.NONE
+
+        return HealthLevel.CRITICAL
 
     def add_alert(self, alert):
         self.alerts.append(alert)
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -19,7 +19,7 @@ from debug_toolbar.panels.sql.utils import (
     is_select_query,
     reformat_sql,
 )
-from debug_toolbar.utils import render_stacktrace
+from debug_toolbar.utils import HealthLevel, render_stacktrace
 
 
 def get_isolation_level_display(vendor, level):
@@ -185,6 +185,21 @@ class SQLPanel(Panel):
             "SQL queries from %(count)d connections",
             count,
         ) % {"count": count}
+
+    @property
+    def health_level(self):
+        """
+        Return the health level of the SQL panel.
+        This is determined by the number of slow queries recorded.
+        """
+        stats = self.get_stats()
+        slow_queries = len([1 for q in stats.get("queries", []) if q.get("is_slow")])
+        if slow_queries > 10:
+            return HealthLevel.CRITICAL
+        elif slow_queries > 0:
+            return HealthLevel.WARNING
+
+        return super().health_level
 
     template = "debug_toolbar/panels/sql.html"
 
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -34,6 +34,13 @@
     --djdt-button-border-color: var(--djdt-table-border-color);
     --djdt-pre-border-color: var(--djdt-table-border-color);
     --djdt-raw-border-color: var(--djdt-table-border-color);
+
+    --djdt-health-background-color-1: #4b3f1b;
+    --djdt-health-color-1: #ffe761;
+    --djdt-health-border-color-1: #ffcc00;
+    --djdt-health-background-color-2: #5a2327;
+    --djdt-health-color-2: #ffb3b3;
+    --djdt-health-border-color-2: #ff0000;
 }
 
 #djDebug[data-theme="dark"] {
@@ -56,6 +63,13 @@
     --djdt-button-border-color: var(--djdt-table-border-color);
     --djdt-pre-border-color: var(--djdt-table-border-color);
     --djdt-raw-border-color: var(--djdt-table-border-color);
+
+    --djdt-health-background-color-1: #4b3f1b;
+    --djdt-health-color-1: #ffe761;
+    --djdt-health-border-color-1: #ffcc00;
+    --djdt-health-background-color-2: #5a2327;
+    --djdt-health-color-2: #ffb3b3;
+    --djdt-health-border-color-2: #ff0000;
 }
 
 /* Debug Toolbar CSS Reset, adapted from Eric Meyer's CSS Reset */
@@ -375,6 +389,29 @@
     width: 38px;
     height: 38px;
     animation: spin 2s linear infinite;
+}
+
+/* Panel and Toolbar hidden button health states */
+#djDebug .djdt-health-1 {
+    /* The background can be shadowed by #djDebugToolbar li.djdt-active a:hover */
+    background: var(--djdt-health-background-color-1) !important;
+    color: var(--djdt-health-color-1);
+}
+
+#djDebug .djdt-health-2 {
+    /* The background can be shadowed by #djDebugToolbar li.djdt-active a:hover */
+    background: var(--djdt-health-background-color-2) !important;
+    color: var(--djdt-health-color-2);
+}
+
+#djDebug .djdt-toolbarhandle-health-1 {
+    /* The border-color is shadowed by the default #djShowToolBarButton border */
+    border-color: var(--djdt-health-border-color-1) !important;
+}
+
+#djDebug .djdt-toolbarhandle-health-2 {
+    /* The border-color is shadowed by the default #djShowToolBarButton border */
+    border-color: var(--djdt-health-border-color-2) !important;
 }
 
 @keyframes spin {
--- a/debug_toolbar/templates/debug_toolbar/base.html
+++ b/debug_toolbar/templates/debug_toolbar/base.html
@@ -31,7 +31,7 @@
     </ul>
   </div>
   <div class="djdt-hidden" id="djDebugToolbarHandle">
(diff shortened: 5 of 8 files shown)


hunzlahmalik:
@matthiask @tim-schilling 

I'll be writing the testcases for this. It would be great if you could take a look at it before testcases.

Thanks!


Edit: You can also suggest how to compute health(debatable name) across other panel profiling, time etc...

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -34,6 +34,13 @@
     --djdt-button-border-color: var(--djdt-table-border-color);
     --djdt-pre-border-color: var(--djdt-table-border-color);
     --djdt-raw-border-color: var(--djdt-table-border-color);
+
+    --djdt-health-bg-1: #4b3f1b;
+    --djdt-health-color-1: #ffe761;
+    --djdt-health-bc-1: #ffcc00;
```
matthiask: What is `bc`? Background color? I think we prefer using variable names without abbreviations.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -377,6 +391,25 @@
...
+    color: var(--djdt-health-color-2) !important;
+}
+
+#djDebug .djdt-toolbarhandle-health-1 {
+    border-color: var(--djdt-health-bc-1) !important;
+}
+
+#djDebug .djdt-toolbarhandle-health-2 {
+    border-color: var(--djdt-health-bc-2) !important;
+}
```
matthiask, line 415: Can you add a comment why all these `!important` are necessary? And/or remove them when they are not?

debug_toolbar/templates/debug_toolbar/base.html
```diff
@@ -31,7 +31,7 @@
     </ul>
   </div>
   <div class="djdt-hidden" id="djDebugToolbarHandle">
-    <div title="{% translate 'Show toolbar' %}" id="djShowToolBarButton">
+    <div title="{% translate 'Show toolbar' %}" id="djShowToolBarButton" class="{% if toolbar.health_level %} djdt-toolbarhandle-health-{{ toolbar.health_level }}{% endif %}">
```
matthiask, line 34: I generally do not like uppercase letters much in ID and class attributes, but I wonder if we should keep capitalization for internal consistency here? Any thoughts?

debug_toolbar/toolbar.py
```diff
@@ -72,6 +72,14 @@ def csp_nonce(self):
         """
         return getattr(self.request, "csp_nonce", None)
 
+    @property
+    def health_level(self):
+        """
+        Return the maximum health level across all panels.
+        This is used to color the toolbar hidden button.
+        """
+        return max(panel.health_level for panel in self.panels)
```
matthiask: What happens when `self.panels` is empty? Is it possible? I think you might want to provide a default value as a fallback (`max((...), NONE)`)

debug_toolbar/utils.py
```diff
@@ -401,3 +402,9 @@ def is_processable_html_response(response):
         and content_encoding == ""
         and content_type in _HTML_TYPES
     )
+
+
+class ToolbarHealthLevel(IntEnum):
```
matthiask: I think this could be called `HealthLevel` only since it actually doesn't only apply to the toolbar itself but also to individual panels.

matthiask (commented):
I like the idea! I added some nitpicks and comments/general questions.

debug_toolbar/templates/debug_toolbar/base.html
hunzlahmalik: They are inconsistent right now. Some places use camelCase when starting class names with `dj`, while others use lowercase with dashes (mostly when starting with `djdt`). 

Rightnow, I can change them to camelcase if you want, but for consistency the changes in whole project will be required in future.

matthiask: Ah, don't worry about it then. Being as consistent as possible is good enough.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -34,6 +34,13 @@
     --djdt-button-border-color: var(--djdt-table-border-color);
     --djdt-pre-border-color: var(--djdt-table-border-color);
     --djdt-raw-border-color: var(--djdt-table-border-color);
+
+    --djdt-health-background-color-1: #4b3f1b;
+    --djdt-health-color-1: #ffe761;
+    --djdt-health-border-color-1: #ffcc00;
+    --djdt-health-background-color-2: #5a2327;
+    --djdt-health-color-2: #ffb3b3;
+    --djdt-health-border-color-2: #ff0000;
```
tim-schilling, line 43: Could we use a descriptive word rather than 1 and 2 here please?

debug_toolbar/utils.py
```diff
@@ -401,3 +402,18 @@ def is_processable_html_response(response):
         and content_encoding == ""
         and content_type in _HTML_TYPES
     )
+
+
+class HealthLevel(IntEnum):
```
tim-schilling, line 407: @matthiask @hunzlahmalik do we want to bikeshed a bit on the name here? If we build out an API, this will likely be included in that and no longer be a private-ish thing. In my mind, "severity" may be more appropriate, but then it may lack the context of what is the severity in relation to.

Not sure, but want to bring it up since it's likely to become a public interface in the near future (< 1 years).

debug_toolbar/toolbar.py
```diff
@@ -72,6 +73,17 @@ def csp_nonce(self):
...
+    @property
+    def health_level(self):
+        """
+        Return the maximum health level across all panels.
+        This is used to color the toolbar hidden button.
+        """
+        if not self.panels:
+            return HealthLevel.NONE
+
+        return max(panel.health_level for panel in self.panels)
```
tim-schilling:
```suggestion
        return max(panel.health_level for panel in self.enabled_panels)
```
We should only try to fetch the value for the panels that are turned on.

tim-schilling:
@hunzlahmalik this is really neat! How does it work with the history panel? Does the color only show for the request that renders the button? If you switch to a request that has problems, does the color update? Last question, I promise, did you consider adding the health level to the history panel so a person can view the high level issues of all the requests in a summary-esque view? I personally think that would be a fantastic way to leverage this feature.

debug_toolbar/utils.py
hunzlahmalik: Yes we can change the name. I thought about it and was inclined to using the `alert`, but it was already taking in the app for the panel. We can use `severity`, we can see this as independent from the colorization. Later this can be use to get the panel severitylevel to do something else.

hunzlahmalik:
For the panels that do define the health_level logic, after switching the request they should update the color too.  Right now there are only two panels with health_level login, sqlpanel and alerts.

See this:
image
image


Regarding summary-esque view, what do you have in mind. We can colorize the rows in the history panel based on health, or do you want a separate some kind of filter/tab for it. I think coloring the rows looks good to me. Like if there is a request with health_level CRITICAL, then that row will be red, with warning, that row will be yellowish.

debug_toolbar/panels/sql/panel.py
```diff
@@ -186,6 +186,21 @@ def title(self):
...
+        """
+        Return the health level of the SQL panel.
+        This is determined by the number of slow queries recorded.
+        """
+        stats = self.get_stats()
+        slow_queries = len([1 for q in stats.get("queries", []) if q.get("is_slow")])
+        if slow_queries > 10:
+            return HealthLevel.CRITICAL
+        elif slow_queries > 0:
+            return HealthLevel.WARNING
```
codingjoe, line 200: The current implementation is susceptible to the number of queries executed on a page. A mean or median threshold might be more reliable.

Also, this is a great usecase for `matcht..case`.

debug_toolbar/panels/sql/panel.py
```diff
@@ -186,6 +186,21 @@ def title(self):
             count,
         ) % {"count": count}
 
+    @property
+    def health_level(self):
+        """
+        Return the health level of the SQL panel.
+        This is determined by the number of slow queries recorded.
+        """
```
codingjoe, line 194: I don't love overwriting docstrings.  PEP257 would have you document the behavior of a function, not its implementation. And the behavior usually doesn't change in subclasses, but only its implementation does.

debug_toolbar/panels/__init__.py
```diff
@@ -130,6 +129,19 @@ def scripts(self):
         """
         return []
 
+    @property
+    def health_level(self):
+        """
+        Returns the health level of the panel as a `ToolbarHealthLevel` enum value.
```
codingjoe, line 135: I know we currently have a wild mix, but it's never too late to write docstrings in the present tense imperative mood.
```suggestion
        Return a calculated health level of the page as a `ToolbarHealthLevel` enum value.
```

debug_toolbar/panels/__init__.py
```diff
@@ -130,6 +129,19 @@ def scripts(self):
...
+
+        This property is used by the toolbar to determine the overall health status of each panel.
+        The default implementation returns `ToolbarHealthLevel.NONE`, indicating no issues.
```
codingjoe, line 138:
```suggestion
        Used by the toolbar to determine the overall health status of each panel.
        The default implementation returns `ToolbarHealthLevel.NONE`, indicating no issues.
```

debug_toolbar/panels/__init__.py
```diff
@@ -130,6 +129,19 @@ def scripts(self):
...
+
+        Subclasses should override this property to provide custom health logic, returning
+        `ToolbarHealthLevel.WARNING` or `ToolbarHealthLevel.ERROR` as appropriate based on panel-specific conditions.
```
codingjoe, line 141: The enum might get more entries later, you might not want to list them.

debug_toolbar/panels/alerts.py
```diff
@@ -92,6 +92,16 @@ def nav_subtitle(self):
...
 
+    @property
+    def health_level(self):
+        """
+        Return the health level of the panel based on the alerts.
+        """
+        if not self.get_stats().get("alerts"):
+            return HealthLevel.NONE
+
+        return HealthLevel.CRITICAL
```
codingjoe, line 103: Let's keep it positive ☮️ 
```suggestion
        if self.get_stats().get("alerts"):
            return HealthLevel.CRITICAL

        return HealthLevel.NONE
```

debug_toolbar/static/debug_toolbar/css/toolbar.css
codingjoe: This might also be a good application for the HSL color space. It also makes it much easier to review than RGB hex code.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -377,6 +391,29 @@
     animation: spin 2s linear infinite;
 }
 
+/* Panel and Toolbar hidden button health states */
+#djDebug .djdt-health-1 {
```
codingjoe, line 395: All browsers support nesting these days. So one `#djDebug` will do just fine.

debug_toolbar/static/debug_toolbar/css/toolbar.css
codingjoe: This will also add a dark mode, much easier in the future.

debug_toolbar/static/debug_toolbar/css/toolbar.css
codingjoe: Overwriting CSS statements is also slow. Ideally, you set the correct value in one shot. You could even calculate it with a 120deg hue range between green and red.

debug_toolbar/toolbar.py
```diff
@@ -84,15 +73,26 @@ def csp_nonce(self) -> str | None:
...
+    @property
+    def health_level(self):
+        """
+        Return the maximum health level across all panels.
+        This is used to color the toolbar hidden button.
+        """
+        if not self.panels:
+            return HealthLevel.NONE
+
+        return max(panel.health_level for panel in self.enabled_panels)
```
codingjoe, line 85: EAFP: Just try to return the max and catch the `ValueError` and return `.NONE`. Branching is expensive.

codingjoe (commented):
Hi @hunzlahmalik,

Thanks for the contribution! Since I am somewhat of a frontend nerd, I invested some time to leave you a thorough review. I hope it's helpful. Don't hesitate to ask questions should my comments have been too vague.

Cheers!
Joe

hunzlahmalik:
Thanks for the review @codingjoe. Will definitely go through these carefully in free time.

SebCorbin:
Just dropping some thoughts here:
- would it be smart to make health levels configurable in settings? (say I want my project to not have any duplicate/similar queries, and prevent page slower than 200ms)
- should we stick to levels or also surface some data (like the number of queries or page timings) like [django-devbar](https://pypi.org/project/django-devbar/)

PR #2184, opened 2025-08-15
Fix KeyError in DatabaseStore when dynamically adding panels and ensure from_store flag is set correctly

### LINKED ISSUE #2166: Gracefully handle panels that don't exist
It's more likely that a user will run into a 500 error where they request a panel id that doesn't exist now that we can persist the toolbar's store. I've run into this locally in my own testing a few times, but couldn't come up with a reproducible pattern.

Relevant code: https://github.com/django-commons/django-debug-toolbar/blob/08a110e2abaa63d22b55a5fb4b06532b3504e3bd/debug_toolbar/views.py#L24

#### Traceback
```
ERROR Internal Server Error: /__debug__/render_panel/
  File "/site-packages/django/core/handlers/exception.py", line 55, in inner
  File "/site-packages/django/core/handlers/base.py", line 197, in _get_response
  File "/site-packages/sentry_sdk/integrations/django/views.py", line 94, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
  File "/site-packages/debug_toolbar/decorators.py", line 34, in inner
    return view(request, *args, **kwargs)
  File "/site-packages/debug_toolbar/decorators.py", line 46, in inner
    return view(request, *args, **kwargs)
  File "/site-packages/debug_toolbar/views.py", line 24, in render_panel
    panel = toolbar.get_panel_by_id(request.GET["panel_id"])
  File "/site-packages/debug_toolbar/toolbar.py", line 79, in get_panel_by_id
    return self._panels[panel_id]
           ~~~~~~~~~~~~^^^^^^^^^^
KeyError: 'ProfilingPanel'
```

I suspect we may want to add a custom exception here when an invalid panel is requested to make it easier for an API integration to be built on top of `Toolbar`.

abdibaker:
#### Description

This commit addresses the KeyError issue that occurs when using DatabaseStore with dynamically added panels to DEBUG_TOOLBAR_PANELS. The problem occurred in `StoredDebugToolbar.from_store()` where the method would only load panels that had stored data, causing a KeyError when accessing panels that were added after the initial request.

The fix modifies the panel loading logic to:
1. Always load all configured panels regardless of stored data existence
2. Properly set the `from_store` flag for panels that have stored data
3. Maintain backward compatibility while ensuring graceful handling of missing panel data

This ensures that the toolbar works correctly when panels are dynamically added to the configuration, particularly when using DatabaseStore for persistent storage.

Fixes #2166

#### Checklist:

### DIFF
--- a/debug_toolbar/toolbar.py
+++ b/debug_toolbar/toolbar.py
@@ -223,7 +223,9 @@ class StoredDebugToolbar(DebugToolbar):
             data = toolbar.store.panel(toolbar.request_id, panel.panel_id)
             if data:
                 panel.load_stats_from_store(data)
-                toolbar._panels[panel.panel_id] = panel
+            else:
+                panel.from_store = True
+            toolbar._panels[panel.panel_id] = panel
         return toolbar
 
 
--- a/tests/test_database_store_fix.py
+++ b/tests/test_database_store_fix.py
@@ -1,0 +1,246 @@
+"""
+Test for the DatabaseStore panel loading fix
+"""
+
+from django.conf import settings
+from django.test import TestCase, override_settings
+from django.test.signals import setting_changed
+
+from debug_toolbar.settings import get_panels
+from debug_toolbar.store import get_store
+from debug_toolbar.toolbar import DebugToolbar, StoredDebugToolbar
+
+
+class DatabaseStorePanelLoadingTestCase(TestCase):
+    """
+    Test that StoredDebugToolbar.from_store loads all configured panels,
+    even those that don't have stored data.
+
+    This fixes the KeyError issue when users dynamically add panels
+    to DEBUG_TOOLBAR_PANELS after requests have been made.
+    """
+
+    def setUp(self):
+        """Clear any cached data and ensure clean state"""
+        # Store original panels configuration
+        self.original_panels = getattr(settings, "DEBUG_TOOLBAR_PANELS", None)
+
+        # Clear store data
+        get_store().clear()
+
+        # Clear panel classes cache
+        DebugToolbar._panel_classes = None
+        StoredDebugToolbar._panel_classes = None
+        get_panels.cache_clear()
+
+        # Reset to default panels to ensure consistent starting state
+        if hasattr(settings, "DEBUG_TOOLBAR_PANELS"):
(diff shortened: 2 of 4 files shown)


debug_toolbar/toolbar.py
```diff
@@ -223,7 +223,9 @@ def from_store(cls, request_id, panel_id=None):
             data = toolbar.store.panel(toolbar.request_id, panel.panel_id)
             if data:
                 panel.load_stats_from_store(data)
-                toolbar._panels[panel.panel_id] = panel
+            else:
+                panel.from_store = True
```
matthiask, line 227: We are only setting `from_store` in `debug_toolbar/panels/__init__.py` right now.

I think it would be better to call `load_stats_from_store` unconditionally if we actually depend on this attribute.

tests/test_database_store_fix.py
```diff
@@ -0,0 +1,246 @@
...
+        DebugToolbar._panel_classes = None
+        StoredDebugToolbar._panel_classes = None
+        get_panels.cache_clear()
+
+        # Reset to default panels to ensure consistent starting state
+        if hasattr(settings, "DEBUG_TOOLBAR_PANELS"):
+            delattr(settings, "DEBUG_TOOLBAR_PANELS")
+        DebugToolbar._panel_classes = None
+        StoredDebugToolbar._panel_classes = None
+        get_panels.cache_clear()
```
matthiask, line 41: I wonder a bit if all that is really necessary?

If you're using the `@override_settings` decorator, a signal is emitted which also calls `cache_clear`.

I'd prefer this to use `@override_settings(DEBUG_TOOLBAR_PANELS=...)` or something similar instead.

matthiask (commented):
Thank you! Makes sense to me. I think the settings handling should be cleaned up if possible.

JohananOppongAmoateng:
@abdibaker  are you still working on this?

tim-schilling:
@JohananOppongAmoateng please feel free to pick this up where it was left off.

tim-schilling:
Thank you for pushing this forward, though I'm closing due to staleness. If you'd like to resume work on it, please open a new PR.

PR #2186, opened 2025-08-15
Fix PostgreSQL compatibility issues in CI and SQL profiling

### LINKED ISSUE #2185: PostgreSQL compatibility issues in CI and SQL profiling
## Problem Description

The CI pipeline has several PostgreSQL compatibility issues that cause test failures:

### 1. Database Connection Error
```
FATAL: role "root" does not exist
```
The PostgreSQL test jobs are trying to connect using the "root" user, but PostgreSQL uses "debug_toolbar" user as configured in the service definition.

### 2. SQL Profiling Compatibility Error  
```
unrecognized configuration parameter "profiling"
```
The SQL panel's profiling feature uses MySQL-specific commands (`SET PROFILING=1/0` and `information_schema.profiling`) that don't exist in PostgreSQL.

### 3. User Creation Constraint Violations
```
duplicate key value violates unique constraint "auth_user_username_key"
```
Test isolation issues causing duplicate user creation attempts.

## Environment
- **CI Environment**: GitHub Actions
- **Database**: PostgreSQL 14 with PostGIS
- **Affected Test Jobs**: All PostgreSQL matrix variants
- **Python Versions**: 3.9, 3.10, 3.11, 3.12, 3.13

## Root Causes

1. **Missing Environment Variables**: PostgreSQL CI jobs don't set `DB_NAME`, `DB_USER`, `DB_PASSWORD` environment variables, unlike MySQL jobs
2. **Database-Specific Code**: SQL profiling assumes MySQL without vendor detection
3. **CI Configuration**: Lint job has malformed strategy configuration

### 1. Fix CI Configuration
- Add missing PostgreSQL environment variables to `.github/workflows/test.yml`
- Fix lint job strategy configuration

### 2. Fix SQL Profiling  
- Add vendor detection to `debug_toolbar.panels.sql.forms.SQLSelectForm.profile()`
- Raise clear error message for non-MySQL databases
- Update error handling in views

### 3. Add Tests
- Add PostgreSQL compatibility tests to verify fixes
- Test that profiling correctly rejects non-MySQL databases

## Impact
- **High**: Blocks all PostgreSQL CI jobs
- **Scope**: Affects testing reliability and PostgreSQL users
- **Regression Risk**: Low - changes are defensive and add proper error handling

## Implementation Plan
1. ✅ Identify root causes
2. ✅ Fix SQL profiling vendor detection  
3. ✅ Update CI configuration
4. ✅ Add comprehensive tests
5. 🔄 Create PR with fixes

The fix maintains backward compatibility while providing proper error messages for unsupported database operations.

abdibaker:
#### Description

This PR fixes PostgreSQL compatibility issues that were causing CI pipeline failures. The main issue was that the SQL panel's profiling feature was attempting to use MySQL-specific commands on PostgreSQL databases, resulting in errors like "unrecognized configuration parameter profiling".

Additionally, the CI configuration had missing environment variables for PostgreSQL test jobs and a malformed lint job configuration.

**Key Changes:**
- Added vendor detection to SQL panel profiling to prevent MySQL-specific commands on other databases
- Fixed CI PostgreSQL job environment variables (DB_NAME, DB_USER, DB_PASSWORD)
- Fixed malformed lint job strategy configuration
- Added comprehensive tests for PostgreSQL compatibility

**Root Cause:** The `SQLSelectForm.profile()` method was executing MySQL-specific `SET PROFILING` commands and querying `information_schema.profiling` without checking the database vendor, causing PostgreSQL CI jobs to fail.

**Solution:** Added vendor detection that raises a clear `ValueError` for non-MySQL databases, preventing the MySQL-specific code from executing on PostgreSQL/SQLite.

Fixes #2185

#### Checklist:

### DIFF
--- a/tests/test_postgresql_compatibility.py
+++ b/tests/test_postgresql_compatibility.py
@@ -1,0 +1,74 @@
+"""
+Tests for PostgreSQL compatibility fixes
+"""
+
+from django.db import connection
+from django.test import TestCase
+
+from debug_toolbar.panels.sql.forms import SQLSelectForm
+
+
+class PostgreSQLCompatibilityTestCase(TestCase):
+    """
+    Test that PostgreSQL-specific compatibility issues are handled correctly.
+    """
+
+    def test_profiling_only_supported_on_mysql(self):
+        """
+        Test that profiling raises appropriate error for non-MySQL databases.
+        """
+        # Skip this test if we're actually using MySQL
+        if connection.vendor == "mysql":
+            self.skipTest("This test is for non-MySQL databases")
+
+        # Create a form and mock the cleaned_data to test profile method
+        form = SQLSelectForm()
+        form.cleaned_data = {
+            "query": {
+                "vendor": connection.vendor,
+                "alias": "default",
+                "raw_sql": "SELECT 1",
+                "params": "[]",
+            }
+        }
+
+        # Test that profiling raises ValueError for non-MySQL databases
+        with self.assertRaises(ValueError) as cm:
+            form.profile()
+
+        self.assertIn("Profiling is not supported", str(cm.exception))
+        self.assertIn(connection.vendor, str(cm.exception))
+
+    def test_profiling_mysql_check(self):
+        """
+        Test that profiling only allows MySQL vendor.
+        """
+        form = SQLSelectForm()
+
+        # Test with PostgreSQL
+        form.cleaned_data = {
+            "query": {
+                "vendor": "postgresql",
+                "alias": "default",
+                "raw_sql": "SELECT 1",
+                "params": "[]",
+            }
+        }
+
+        with self.assertRaises(ValueError) as cm:
+            form.profile()
+        self.assertIn("Profiling is not supported for postgresql", str(cm.exception))
+
+        # Test with SQLite
+        form.cleaned_data = {
+            "query": {
+                "vendor": "sqlite",
+                "alias": "default",
+                "raw_sql": "SELECT 1",
+                "params": "[]",
+            }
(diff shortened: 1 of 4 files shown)


tim-schilling (changes requested):
I think the idea of this change is a quality one, but I think we should move the logic up into the form layer. Try subclassing `SQLSelectForm` with some custom validation to do this check. That will protect this in the `sql_profile` and keep our instrumentation logic cleaner.

We should also prevent the profile button from appearing when using an incompatible vendor.

Lastly, we may want to check [Django's own backend features](https://github.com/django/django/blob/main/django/db/backends/base/features.py) to see if this is something we can base off Django's own checks. There's no sense in both repositories maintaining a boolean constant about whether or not this feature is supported in MySQL.

tests/test_postgresql_compatibility.py
```diff
@@ -0,0 +1,74 @@
...
+    Test that PostgreSQL-specific compatibility issues are handled correctly.
+    """
+
+    def test_profiling_only_supported_on_mysql(self):
+        """
+        Test that profiling raises appropriate error for non-MySQL databases.
+        """
+        # Skip this test if we're actually using MySQL
+        if connection.vendor == "mysql":
+            self.skipTest("This test is for non-MySQL databases")
```
tim-schilling, line 22: This should be moved up as a decorator. Search for other usages of `connection.vendor` and tests that are skipped.

tim-schilling:
Thank you for pushing this forward, though I'm closing due to staleness. If you'd like to resume work on it, please open a new PR.

PR #2193, opened 2025-09-11
Add Community Panel

alan-rosenberg:
#### Description

This PR introduces a Community Panel to the Django Debug Toolbar, providing contributors with easy access to key community resources. Builds off of changes and feedback on [this](https://github.com/django-commons/django-debug-toolbar/pull/2119) PR by @andoriyaprashant.

Changes Introduced:

- Added Discussions link for community engagement.
- Added Issues link for bug reporting and feature requests.
- Added Contribution Guide link to encourage contributions.
- Added Documentation link for quick access to official docs.
- Added Resources link for learning how to use the debug toolbar.

- Fixes #2033
- Fixes #2194

Images:
Screenshot 2025-09-11 at 2 41 17 PM
Screenshot 2025-09-11 at 2 41 08 PM

#### Checklist:

- [X] I have added the relevant tests for this change.
- [X] I have added an item to the Pending section of ``docs/changes.rst``.

PR #2196, opened 2025-09-12
Changed StoredDebugToolbar.from_store to always create panel keys.

### LINKED ISSUE #2195: Unable to create a panel without a call to `record_stats`
When creating an empty panel such as:

```
class MyPanel(Panel):
    template = "debug_toolbar/panels/my_panel.html"
```

It errors when trying to access the panel in the `render_panel` view because the key doesn't exist in the store. I'll have a fix up for this shortly.

tim-schilling:
#### Description

This fixes the issue where panels that don't call record_stats aren't fetchable in render_panel. This unblocks #2193

Fixes #2195

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,7 +89,7 @@ jobs:
             - python-version: '3.13'
               database: postgis3
             - python-version: '3.13'
-              database: postgresql
+              database: psycopg3
               env:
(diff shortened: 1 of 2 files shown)


tim-schilling:
@matthiask I found the bug in the build and added a commit to fix CI

matthiask:
Great, thank you! I was staring at this for a long time without finding the cause.

PR #2198, opened 2025-09-13
Add Open in Editor support for template paths

federicobond:
#### Description

This adds a setting to configure an Open in editor URL when you click on a template path. Makes it easier to jump directly to the template and edit. Can develop it further and/or add tests if there is interest in the feature.

Captura de pantalla 2025-09-13 a la(s) 23 04 38

#### Checklist:


docs/configuration.rst
```diff
@@ -67,6 +67,14 @@ Toolbar options
   This setting is a set of the full Python paths to each panel that you
   want disabled (but still displayed) by default.
 
+* ``EDITOR_URL_CALLBACK``
+
+  Default: ``lambda f: None``
+
+  This function is passed a file path and should return a URL to open the file in
+  the editor. For example, to open the file in VS Code, you can use:
+  ``lambda f: f"vscode://file/{f}"``
```
tim-schilling: Is this ever not going to be a static prefix with the filename after it? I think if we can remove the lambda, that would be better.

tim-schilling (commented):
I wonder if it's possible to detect that the project is running in vscode and we could avoid the setting altogether?

docs/configuration.rst
federicobond: Not in every case, but an alternative is to own the different formats ourselves like the PHP Debug Bar does.

It's a better UX for most people and lets us add line references later at the cost of a more restricted set of supported editors.

See https://github.com/php-debugbar/php-debugbar/blob/c9a88bbc89958ccdab95734ad396cc82f8c65e43/src/DebugBar/DataFormatter/HasXdebugLinks.php#L107-L129

tim-schilling:
Ah, that's a good reference. I like the idea of having a magic-ish value/constant to specify rather than asking them to write in a lambda for a setting.

federicobond:
@tim-schilling updated to use a constant to specify the editor URL format.

Test failures appear unrelated.

debug_toolbar/utils.py
```diff
@@ -401,3 +401,28 @@ def is_processable_html_response(response):
...
+def get_editor_url(file: str, line: int = 1) -> str | None:
+    formats = {
+        "cursor": "cursor://file/{file}:{line}",
+        "emacs": "emacs://open?url=file://{file}&line={line}",
+        "espresso": "x-espresso://open?filepath={file}&lines={line}",
+        "idea": "idea://open?file={file}&line={line}",
+        "idea-remote": "javascript:(()=>{let r=new XMLHttpRequest; r.open('get','http://localhost:63342/api/file/?file={file}&line={line}');r.send();})()",
+        "macvim": "mvim://open/?url=file://{file}&line={line}",
+        "nova": "nova://open?path={file}&line={line}",
+        "pycharm": "pycharm://open?file={file}&line={line}",
```
tim-schilling, line 430: :star_struck:

debug_toolbar/panels/templates/panel.py
```diff
@@ -195,13 +196,15 @@ def generate_stats(self, request, response):
...
                 template.origin_hash = signing.dumps(template.origin.name)
+                template.editor_url = get_editor_url(template.origin.name)
             else:
                 template.origin_name = _("No origin")
                 template.origin_hash = ""
             info["template"] = {
                 "name": template.name,
                 "origin_name": template.origin_name,
                 "origin_hash": template.origin_hash,
+                "editor_url": getattr(template, "editor_url", None),
```
tim-schilling: Let me know if you feel this is nitpicky. The toolbar now collects data/stats and writes it to the store to be rendered out at a later time. There's an opportunity to annotate the data coming out for an individual panel in `get_stats()` instead of doing it on the `generate_stats()` / `record_stats()` side. The benefit here would be if someone were to change their settings to use vscode to sublime or vice versa, that would be reflected. If we write the URL, we're committing that particular toolbar config to the store.

I think in some cases we would want to store the data with the current config, but in other cases, like this one, it's better to always use the latest config.

What do you think @federicobond? Nitpicky or an actual DX improvement that's worth making?

federicobond: Not at all, makes total sense. I did not think about that pattern because none of the other panels override `get_stats` but it's not difficult to make that change.

debug_toolbar/utils.py
federicobond: The line info will be useful for linking traceback lines directly to the correct part of the code.

tim-schilling: Ah, unfortunately pycharm isn't supported from what I can tell.

tim-schilling: This may be a linux only problem. I'm tracking down some help to confirm on windows and mac.

tim-schilling: Yeah, this was a linux only problem. Going to hide these comments.

docs/configuration.rst
```diff
@@ -67,6 +67,16 @@ Toolbar options
...
 
+* ``EDITOR``
+
+  Default: ``'vscode'``
+
+  The editor to use to open file paths from the toolbar.
+
+  Available editors: ``'vscode'``, ``'cursor'``, ``'emacs'``, ``'idea'``,
+  ``'pycharm'``, ``'sublime'``, ``'vscode-insiders'``, ``'vscode-remote'``,
+  ``'vscodium'``, ``'windsurf'``
```
tim-schilling:
```suggestion
  Available editors:

  * ``'cursor'``
  * ``'emacs'``
  * ``'idea'``
  * ``'pycharm'``
  * ``'sublime'``
  * ``'vscode'``
  * ``'vscode-insiders'``
  * ``'vscode-remote'``
  * ``'vscodium'``
  * ``'windsurf'``
```

Let's make it proper alphabetical order and use some bullets to make it easier to scan for the reader.

tim-schilling (changes requested):
This is looking good. Marking it as Request Changes to ask for the tests. I think tests that cover `get_editor_url` and the new `get_stats()` of `TemplatesPanel` would be great.

The other stretch thing would be to add is a reference to the new setting in the TemplatesPanel documentation.

federicobond:
I just added the requested tests and updated the docs format.

I would prefer to hold off with documenting this as something that's mostly useful for the requests panel because I believe it's going to be useful in other places too (like stacktraces).

tests/test_utils.py
```diff
@@ -171,3 +172,33 @@ def test_non_dict_input(self):
...
+            "nova": "nova://open?path=test.py&line=5",
+            "pycharm": "pycharm://open?file=test.py&line=5",
+            "pycharm-remote": "javascript:(()=>{let r=new XMLHttpRequest; r.open('get','http://localhost:63342/api/file/test.py:5');r.send();})()",
+            "sublime": "subl://open?url=file://test.py&line=5",
+            "vscode": "vscode://file/test.py:5",
+            "vscode-insiders": "vscode-insiders://file/test.py:5",
+            "vscode-remote": "vscode://vscode-remote/test.py:5",
+            "vscode-insiders-remote": "vscode-insiders://vscode-remote/test.py:5",
+            "vscodium": "vscodium://file/test.py:5",
+            "windsurf": "windsurf://file/test.py:5",
```
tim-schilling, line 196:
```suggestion
            "windsurf": "windsurf://file/test.py:5",
            "non-existent-editor": None,
```
Nitpick to cover the no editor test case.

tests/panels/test_template.py
```diff
@@ -153,6 +153,18 @@ def test_template_source(self):
...
+        response = self.panel.process_request(self.request)
+        Template("").render(Context({}))
+        Template("", origin=Origin("test.html")).render(Context({}))
+        self.panel.generate_stats(self.request, response)
+        stats = self.panel.get_stats()
+        self.assertNotIn("editor_url", stats["templates"][0]["template"])
+        self.assertEqual(
+            stats["templates"][1]["template"]["editor_url"],
+            "vscode://file/test.html:1",
+        )
```
tim-schilling:
```suggestion
    def test_get_stats_includes_editor_url(self):
        response = self.panel.process_request(self.request)
        Template("", origin=Origin("test.html")).render(Context({}))
        self.panel.generate_stats(self.request, response)
        stats = self.panel.get_stats()
        self.assertEqual(
            stats["templates"][0]["template"]["editor_url"],
            "vscode://file/test.html:1",
        )

    def test_get_stats_excludes_editor_url(self):
        response = self.panel.process_request(self.request)
        Template("").render(Context({}))
        self.panel.generate_stats(self.request, response)
        stats = self.panel.get_stats()
        self.assertNotIn("editor_url", stats["templates"][0]["template"])
```

I think it's slightly clearer what the test is for if we split it into two.

tim-schilling:
That's fair about mentioning the editor setting the templates panel documentation. I think we do want to have this mentioned in the changes.rst file so folks get alerted to the change in the next release.

federicobond:
Done!

tim-schilling (approved):
I'm good with these changes. I'd like @matthiask to agree (I don't want to merge a new feature unilaterally). He's out on holiday currently, so it may be a few weeks.

tim-schilling:
Thank you @federicobond for your patience and perseverance!

federicobond:
Rebased against latest `main`.

debug_toolbar/utils.py
```diff
@@ -415,3 +415,28 @@ def get_csp_nonce(request) -> str | None:
...
+        "vscode-insiders": "vscode-insiders://file/{file}:{line}",
+        "vscode-remote": "vscode://vscode-remote/{file}:{line}",
+        "vscode-insiders-remote": "vscode-insiders://vscode-remote/{file}:{line}",
+        "vscodium": "vscodium://file/{file}:{line}",
+        "windsurf": "windsurf://file/{file}:{line}",
+    }
+    template = formats.get(dt_settings.get_config()["EDITOR"])
+    if template is None:
+        return None
+    return template.format(file=file, line=line)
```
codingjoe, line 442:
```suggestion
    if template:
        return template.format(file=file, line=line)
```

debug_toolbar/utils.py
```diff
@@ -415,3 +415,28 @@ def get_csp_nonce(request) -> str | None:
         return csp_nonce
     # Django's built-in CSP support uses get_nonce(request)
     return compat.get_nonce(request)
+
+
+def get_editor_url(file: str, line: int = 1) -> str | None:
+    formats = {
```
codingjoe, line 421: This would be a great application for the all new template strings.

debug_toolbar/settings.py
```diff
@@ -22,6 +22,7 @@ def _is_running_tests():
         "debug_toolbar.panels.profiling.ProfilingPanel",
         "debug_toolbar.panels.redirects.RedirectsPanel",
     },
+    "EDITOR": "vscode",
```
codingjoe, line 25: I don't think there should be a default. I'd rather have the feature disabled by default.

codingjoe (commented):
@federicobond may I be so bold as to propose a slimmer solution?
Instead of hardcoding all the different URIs for all the different tool, why not just give the option to provide a URI as a setting?

If not, this should at least be the fallback. So if the URI is not in the dictionary, whatever string was provided is interpreted as a custom URI template.

tim-schilling:
@codingjoe I think we're going to have a difference of opinion here. https://github.com/django-commons/django-debug-toolbar/pull/2198#issuecomment-3365860113

codingjoe:
> @codingjoe I think we're going to have a difference of opinion here. [#2198 (comment)](https://github.com/django-commons/django-debug-toolbar/pull/2198#issuecomment-3365860113)

Yes, I saw that, which is why I proposed both at the end. Joining ease of use with flexibility for the obscure. It also doesn't require a lambda; the setting can remain a string.

debug_toolbar/settings.py
tim-schilling: I think defaulting to enabled with vscode is reasonable.

codingjoe: Isn't JetBrains a big sponsor? If we're giving special treatment, let's make it community-friendly.

However, I was going for default OFF. I don't have VSCode installed, so I would get an error.

elineda: I'm for the off by default, no spécial traitment. And if a special traitment there is it should be the most used ide in the django community. Dunno how to find this info.

tim-schilling: @codingjoe as far as I can tell, it doesn't error. It's a no-op in the browser. If we want, we're safe picking _something_ as the default. However, we should consider how to make this setting a bit more obvious. I'll push a commit that adds the note and a link to the setting in the docs.

Image

tim-schilling: Should be resolved in 1d6349f4b6d0381b9ff6c3c5d04ec1b6b0b67bef

tim-schilling (approved):

tim-schilling:
@federicobond do you want to get this rebased on main (or merge main into it, we can squash commit it)? We'll need the changes.rst file cleaned up though.

PR #2201, opened 2025-09-17
Add zizmor for GHA security analysis

tim-schilling:
Code sourced from: https://docs.zizmor.sh/integrations/#github-actions

#### Checklist:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,8 @@ name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
 
 on: push
 
+permissions: {}
+
 env:
   PYPI_URL: https://pypi.org/p/django-debug-toolbar
   PYPI_TEST_URL: https://test.pypi.org/p/django-debug-toolbar
@@ -14,6 +16,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v5
+      with:
+        persist-credentials: false
     - name: Set up Python
       uses: actions/setup-python@v6
       with:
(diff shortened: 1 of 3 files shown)


matthiask (approved):
Thanks. I think it's a good idea to try this out.

tim-schilling:
We were discussing this in https://github.com/django-commons/best-practices/pull/13 and decided on limiting it to only run based on `.github/` files. What do you think about that?

.github/workflows/release.yml
```diff
@@ -79,7 +83,7 @@ jobs:
         GITHUB_TOKEN: ${{ github.token }}
       run: >-
         gh release create
-        '${{ github.ref_name }}'
+        '${GITHUB_REF_NAME}'
```
woodruffw, line 86: JFYI, I think this will cause a failure in your release process -- it's still single quoted so the shell variable won't be expanded correctly 🙂 

(i.e. it needs to be double-quoted instead.)

.github/workflows/release.yml
```diff
@@ -90,7 +94,7 @@ jobs:
       # sigstore-produced signatures and certificates.
       run: >-
         gh release upload
-        '${{ github.ref_name }}' dist/**
+        '${GITHUB_REF_NAME}' dist/**
```
woodruffw, line 97: Same as above!

woodruffw:
Sorry for being nosey on this PR, but I was looking at downstream users of `zizmor-action` and saw two things that I figured might cause problems for you all once you do your next release.

(And thanks for trying out zizmor!)

tim-schilling:
Thank you for the spot check! I knew I needed to test the release, but didn't get around to it. I really need to automate that 😅

PR #2211, opened 2025-09-25
Add SQLPanel support for DDT model query tracking

### LINKED ISSUE #2165: SQLPanel is tracking the DatabaseStore queries
The SQLPanel is tracking toolbar queries. While probably helpful for understanding the impact of the tool, I'm not sure if we should be tracking them. It's not really related to the developer's app.

#### Steps to reproduce
Settings:
```python
DEBUG_TOOLBAR_CONFIG = {
    "TOOLBAR_STORE_CLASS": "debug_toolbar.store.DatabaseStore",
}
```
1. Access the admin
2. View the SQL Panel
3. The first two lines should be related to `debug_toolbar_historyentry`

theShinigami:
#### Description

Added a new context variable `allow_ddt_models_tracking` to control whether  `SQLPanel` tracks *DDT* queries.

- Added `allow_ddt_models_tracking` contextvar, defaulting to `False`
- `SQLPanel` now checks this variable to decide if *DDT* models queries should be tracked.

Fixes #2165 

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -8,6 +8,7 @@ import django.test.testcases
 from django.utils.encoding import force_str
 
 from debug_toolbar.utils import get_stack_trace, get_template_info
+from django.apps import apps
 
 try:
     import psycopg
@@ -26,6 +27,17 @@ except ImportError:
 # by the TemplatePanel to prevent the toolbar from issuing
 # additional queries.
 allow_sql = contextvars.ContextVar("debug-toolbar-allow-sql", default=True)
+
+# Prevents tracking of DDT models
+allow_ddt_models_tracking = contextvars.ContextVar(
+    "debug-toolbar-allow-ddt-models", default=False
+)
+
+DDT_MODELS = {
+    m._meta.db_table
+    for m in apps.get_app_config("debug_toolbar").get_models()
+    if m._meta.app_label == "debug_toolbar"
+}
 
 
 class SQLQueryTriggered(Exception):
@@ -223,6 +235,12 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
                         "iso_level": iso_level,
                     }
                 )
+
+            # Skip recording if query includes DDT models.
+            if not allow_ddt_models_tracking.get() and any(
+                table in sql for table in DDT_MODELS
+            ):
+                return
 
             # We keep `sql` to maintain backwards compatibility
             self.logger.record(**kwargs)
--- a/tests/panels/test_sql.py
+++ b/tests/panels/test_sql.py
@@ -16,6 +16,7 @@ from django.test.utils import override_settings
 
 import debug_toolbar.panels.sql.tracking as sql_tracking
 from debug_toolbar.panels.sql import SQLPanel
+from debug_toolbar.models import HistoryEntry
 
 try:
     import psycopg
@@ -31,6 +32,20 @@ def sql_call(*, use_iterator=False):
     if use_iterator:
         qs = qs.iterator()
     return list(qs)
+
+
+def sql_call_ddt(*, use_iterator=False):
+    qs = HistoryEntry.objects.all()
+    if use_iterator:
+        qs = qs.iterator()
+    return list(qs)
+
+
+async def async_sql_call_ddt(*, use_iterator=False):
+    qs = HistoryEntry.objects.all()
+    if use_iterator:
+        qs = qs.iterator()
+    return await sync_to_async(list)(qs)
 
 
 async def async_sql_call(*, use_iterator=False):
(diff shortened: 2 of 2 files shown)


debug_toolbar/panels/sql/tracking.py
```diff
@@ -224,6 +236,12 @@ def _record(self, method, sql, params):
                     }
                 )
 
+            # Skip recording if query includes DDT models.
+            if not allow_ddt_models_tracking.get() and any(
+                table in sql for table in DDT_MODELS
+            ):
+                return
```
matthiask: We definitely shouldn't be returning inside a `finally` block.

Starting in python 3.14 the compiler emits a `SyntaxWarning` when seeing this construct:
https://docs.python.org/3.14/reference/compound_stmts.html#finally

theShinigami: I've noticed it when `ruff` complained :laughing:, I'm making changes. Thank you for the reference. :slightly_smiling_face:

debug_toolbar/panels/sql/tracking.py
```diff
@@ -27,6 +28,17 @@
 # additional queries.
 allow_sql = contextvars.ContextVar("debug-toolbar-allow-sql", default=True)
 
+# Prevents tracking of DDT models
+allow_ddt_models_tracking = contextvars.ContextVar(
+    "debug-toolbar-allow-ddt-models", default=False
+)
```
tim-schilling: @theShinigami I don't think I've communicated this idea well. The purpose here is to have a value that enables and disables whether we're using the sql panel. So by default it should track all SQL. However, when we get to the toolbar's own processing, it should be disabled. That way the toolbar isn't monitoring itself. Does that make sense?

theShinigami: @tim-schilling I'm not quite getting the idea, could you elaborate a little more? :slightly_smiling_face:

tim-schilling: Given this context var:

```python
is_recording = contextvars.ContextVar(
    "debug-toolbar-is-recording", default=True
)
```

Think of the `is_recording` as a global variable that controls whether a panel actually records the activity of the django project. When it's `True`, the SQL panel should record SQL queries. When it's False, the SQL panel shouldn't record SQL queries.

This comes into play for this issue because we want to disable the SQL panel when the toolbar is processing the request, but not when everything else is.

So we need to identify when to set and when to unset it (somewhere in the middleware most likely, we may run into some weirdness with the async flow), then lastly update the SQL panel to avoid tracking when it's not supposed to.

The places that you'll likely want to update are:
- [Prevent the SQLPanel from actually recording queries here](https://github.com/django-commons/django-debug-toolbar/blob/main/debug_toolbar/panels/sql/panel.py#L149)
- [Set and unset `is_recording` here](https://github.com/django-commons/django-debug-toolbar/blob/main/debug_toolbar/middleware.py#L181-L183)

This relies on the idea that `record_stats` (what uses the HistoryEntry model) will only ever be called from `generate_stats`. I think that will work...

tim-schilling: I feel like I may be missing something here around `enable_instrumentation` and `disable_instrumentation`. This global variable shouldn't be necessary. Sorry, my thoughts aren't fully baked yet

tests/panels/test_sql.py
```diff
@@ -33,6 +34,20 @@ def sql_call(*, use_iterator=False):
     return list(qs)
 
 
+def sql_call_ddt(*, use_iterator=False):
+    qs = HistoryEntry.objects.all()
+    if use_iterator:
+        qs = qs.iterator()
```
tim-schilling: Seems like this isn't used, so we should be able to remove it and the one from `async_sql_call_ddt`

tests/panels/test_sql.py
```diff
@@ -33,6 +34,20 @@ def sql_call(*, use_iterator=False):
     return list(qs)
 
 
+def sql_call_ddt(*, use_iterator=False):
```
tim-schilling: Could we add a doc string here to explain what this function is for please? I think that'll help us in the future when reviewing code.

tests/panels/test_sql.py
```diff
@@ -104,6 +119,54 @@ async def test_recording_concurrent_async(self):
...
+
+        sql_call_ddt()
+
+        # ensure query was logged
+        self.assertEqual(len(self.panel._queries), 1)
+        query = self.panel._queries[0]
+        self.assertEqual(query["alias"], "default")
+        self.assertTrue("sql" in query)
+        self.assertTrue("duration" in query)
+        self.assertTrue("stacktrace" in query)
```
tim-schilling: I think we may be better off replacing this with a check on the models' table in the sql query rather than these assertions.

tests/panels/test_sql.py
```diff
@@ -104,6 +119,54 @@ async def test_recording_concurrent_async(self):
...
+
+        # ensure the stacktrace is populated
+        self.assertTrue(len(query["stacktrace"]) > 0)
+
+    def test_ddt_models_untracking(self):
```
tim-schilling:
```suggestion
    def test_ddt_models_not_tracked(self):
```
nit: Small naming improvement (should be used on the other similar test too)

tests/panels/test_sql.py
```diff
@@ -104,6 +119,54 @@ async def test_recording_concurrent_async(self):
         # ensure the stacktrace is populated
         self.assertTrue(len(query["stacktrace"]) > 0)
 
+    @override_settings(DEBUG_TOOLBAR_CONFIG={"TRACK_DDT_MODELS": True})
+    def test_ddt_models_tracking(self):
```
tim-schilling:
```suggestion
    def test_ddt_models_tracked(self):
```
nit: Small naming improvement (should be used on the other similar test too)

tests/panels/test_sql.py
tim-schilling: Ack, posted too soon. I think a doc string on these two helper functions and the four tests would be fantastic.

tim-schilling:
I think I can live with the SQL search. I don't like it because anything can match on it, but I'm not sure how to exactly improve it. @matthiask any thoughts?

tim-schilling:
To clarify, a query such as `select * from my_other_table WHERE my_column = 'debug_toolbar.historyentry'; would match. For any possible query that a Django application is running, we're looking for a specific string in it. There's a possibility of false positives (matches that we don't actually want to match on)

theShinigami:
Hey @matthiask, did you manage to review this PR?

theShinigami:
@matthiask `SKIP_TOOLBAR_QUERIES` makes sense, I'll make the changes. :slightly_smiling_face:

theShinigami:
@tim-schilling @matthiask I've updated the tests to account for the new changes that hides DDT models when `TOOLBAR_STORE_CLASS` is not set to `debug_toolbar.store.DatabaseStore`.

tim-schilling:
@theShinigami thank you for making those changes! I renamed some of the methods to use "toolbar" instead of "ddt" unless it was specifically referring to the `DDT_MODELS` variable. I tweaked some of the doc strings to be a bit shorter and enforced a line limit. The last thing I added was the documentation around `SKIP_TOOLBAR_QUERIES` and added a mention to the changelog. If all the tests pass, I'll be happy to merge this.

tim-schilling:
And of course I forget to rebase on a fresh main :melting_face: Let me try that again.

PR #2217, opened 2025-10-03
Fix django unicode error

jmgutu:
#### Description

The force serialize function in [django-debug-toolbar/debug_toolbar/panels/settings.py] was throwing errors for values that are not serializable. The proposed solution returns the serialized value if the value is serializable or returns 'Debug toolbar was unable to parse value'

Fixes # (issue)

#### Checklist:

- [ X] I have added the relevant tests for this change.
- [ X] I have added an item to the Pending section of ``docs/changes.rst``.

### DIFF
--- a/debug_toolbar/panels/settings.py
+++ b/debug_toolbar/panels/settings.py
@@ -1,6 +1,7 @@
 from django.utils.encoding import force_str
 from django.utils.translation import gettext_lazy as _
 from django.views.debug import get_default_exception_reporter_filter
+from django.utils.encoding import DjangoUnicodeDecodeError
 
 from debug_toolbar.panels import Panel
 
@@ -24,10 +25,17 @@ class SettingsPanel(Panel):
         )
 
     def generate_stats(self, request, response):
+
+        def catch_force_errors(force_function, value):
+            try:
+                return force_function(value)
+            except DjangoUnicodeDecodeError:
+                return 'Debug toolbar was unable to parse value'
+
         self.record_stats(
             {
                 "settings": {
-                    key: force_str(value)
+                    key: catch_force_errors(force_str, value)
                     for key, value in sorted(get_safe_settings().items())
                 }
             }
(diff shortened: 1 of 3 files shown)


debug_toolbar/panels/settings.py
```diff
@@ -24,10 +24,16 @@ def title(self):
         )
 
     def generate_stats(self, request, response):
+        def catch_force_errors(force_function, value):
+            try:
+                return force_function(value)
+            except DjangoUnicodeDecodeError:
+                return "Debug toolbar was unable to parse value"
```
tim-schilling: There are a few places we use `force_str` and I suspect we may want to swap all of them out with this new function. Perhaps we should create a `sanitize.py` file and put this in there with a doc string explaining the interface and what it does. Then use that everywhere we've been using `force_str`. I'd consider keeping the name as `force_str` too, but that's just my preference. We'd want to have an explicit test covering the functionality of this new function too.

There are some other sanitizing functions in `utils.py` that can be moved over too, though that should be a separate commit. Or I can do that.

tim-schilling:
```suggestion
                return "Django Debug Toolbar was unable to parse value."
```
nitpick: We use the full name in other messaging presented to users.

docs/changes.rst
```diff
@@ -13,6 +13,8 @@ Pending
   class instance, regardless if any data was generated.
 * Fixed selenium tests for CI by using psycopg for Python 3.13 runs.
 * Added ``CommunityPanel`` containing links to documentation and resources.
+* Fixed force_str to catch error and give out default string if value is not
+  serializable.
```
```suggestion
* Updated logic that forces values to strings to render "Django Debug Toolbar was
  unable to parse value." when there's a decoding error.
```

I think we're better off trying to communicate what is going to change from a developer experience here rather than what technically changed. A user may try to figure out why they are getting this message and our documentation may not explain it. Having it here in the change log will help them realize it's intentional and be able to track what the issue/PR is to see why it was changed.

tim-schilling (changes requested):
Thank you @jmgutu for putting this together quickly. I think there are a few changes we should make to improve the toolbar as a whole and to make the developer experience a bit better. Please let me know if you think differently.

PR #2219, opened 2025-10-08
Show cache alias instead of backend repr in calls table

federicobond:
#### Description

This changes the values of the backend columns for cache calls in the Cache panel from `<django.core.cache.backends.dummy.DummyCache object at 0x1063950a0>` to `default` or the corresponding alias. This makes it much easier to identify which call is made to which backend when several cache backends share the same engine.

Before:

Captura de pantalla 2025-10-08 a la(s) 9 48 20

After:

Captura de pantalla 2025-10-08 a la(s) 9 49 26

#### Checklist:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -8,6 +8,8 @@ Pending
   toolbar data from redirected requests.
 * Fixed support for generating code coverage comments in PRs.
 * Added Django 6.0 to the testing matrix.
+* Show the cache backend alias and cache backend class name instead of
+  the cache instance in the cache panel.
 
 6.1.0 (2025-10-30)
 ------------------
@@ -83,8 +85,6 @@ Pending
   conflicts
 * Added CSS for resetting the height of elements too to avoid problems with
   global CSS of a website where the toolbar is used.
-* Show the cache backend alias instead of the cache instance for each call in
-  the cache panel.
 
 5.1.0 (2025-03-20)
 ------------------


debug_toolbar/panels/cache.py
```diff
@@ -194,9 +194,10 @@ def enable_instrumentation(self):
         # requests.  The monkey patch of CacheHander.create_connection() installed in
         # the .ready() method will ensure that any new cache connections that get opened
         # during this request will also be monkey patched.
-        for cache in caches.all(initialized_only=True):
-            _monkey_patch_cache(cache)
-            cache._djdt_panel = self
+        for alias in caches:
+            if hasattr(caches._connections, alias):
```
tim-schilling: I feel like it's going to be hard for us to remember that this is effectively the old version of `initialized_only=True`. Do you have any ideas on making this more maintainable? [Specifically if Django changes its implementation upstream.]

The only idea I have is to use a `cache._djdt_panel.record = functools.partial(cache._djdt_panel.record, alias=alias)`

Though it makes me wonder if we need the panel on the cache at all. Seems like we could just store a `_djdt_record` function on the cache, where then the partial gets much cleaner.

federicobond: > Though it makes me wonder if we need the panel on the cache at all. Seems like we could just store a _djdt_record function on the cache, where then the partial gets much cleaner.

We do not need the panel on the cache in the current code, it's just used a monkeypatch marker.

> The only idea I have is to use a cache._djdt_panel.record = functools.partial(cache._djdt_panel.record, alias=alias)

I thought about something similar, but it needed more changes to the panel inner workings so I went with the most straightforward implementation to validate interest first. I could try a similar approach, storing a `_djdt_record` function.

federicobond: I came back to rework this but if I understand correctly there is no way to get the alias for the given cache without initializing a connection to the backend as a side effect. Checking for the `caches._connections` looks like the only way to work around this when monkeypatching existing connections.

tim-schilling: Alright thank you @federicobond. We need to document this somehow in the code. I'm between separating it out as it's own function or just commenting the loop here:

```python
def initialized_caches():
    """
    Return the initialized caches and aliases.

    This does the same as`caches.all(initialized_only=True)`, but keeps
    the alias with each cache instance.
    """
    for alias in caches:
        if hasattr(caches._connections, alias):
            yield caches[alias], alias
```

So the rest can be
```python
for cache, alias in initialized_caches():
    _monkey_patch_cache(caches[alias], alias)
    cache._djdt_panel = self
```

debug_toolbar/panels/cache.py
```diff
@@ -39,15 +39,15 @@ def wrapper(*args, **kwargs):
...
             return original_method(*args, **kwargs)
         else:
-            return panel._record_call(cache, name, original_method, args, kwargs)
+            return panel._record_call(cache, alias, name, original_method, args, kwargs)
 
     setattr(cache, name, wrapper)
 
 
-def _monkey_patch_cache(cache):
+def _monkey_patch_cache(cache, alias):
```
tim-schilling: We can push more of the monkey patching down into this function to clean up the panel logic if we accept panel as an argument.

```suggestion
def _monkey_patch_cache(cache, alias, panel):
```

debug_toolbar/panels/cache.py
```diff
@@ -39,15 +39,15 @@ def wrapper(*args, **kwargs):
...
     if not hasattr(cache, "_djdt_patched"):
         for name in WRAPPED_CACHE_METHODS:
-            _monkey_patch_method(cache, name)
+            _monkey_patch_method(cache, name, alias)
```
tim-schilling, line 50:
```suggestion
            _monkey_patch_method(cache, name, alias)
        cache._djdt_panel = panel
```

tim-schilling:
@federicobond I spotted a few more clean-ups, but this is on the edge of being good to go. Thank you for being patient!

debug_toolbar/panels/cache.py
federicobond: Done!

tim-schilling:
@federicobond I removed the `_djdt_patched` attribute since it was causing confusion in the tests of when the cache integration had been patched or not. I also added in a suffix of `(BackendClassName)` to the rendered text. I figure some people would benefit from knowing what backend class is being used. If the tests pass, I'm happy to merge this.

tim-schilling:
Squashed the commits and merged. Thank you @federicobond!

PR #2220, opened 2025-10-09
Docs/add docs generation guide

### LINKED ISSUE #2215: Add contributing documentation on how to generate docs locally
Going through the Djangonaut Space, we realized that we don't have instructions on how to generate and view the documentation locally. As best as I can tell, it's navigating to the `docs/` folder and running `make html && open _build/html/index.html`:

```bash
cd docs/
make html
open _build/html/index.html
```

I wonder if it's the same on Windows?

okotdaniel:
#### Description

This pull request adds a new section to the contributing documentation that explains how to build and preview the project documentation locally using Make and Tox.

Previously, contributors did not have clear instructions on generating the documentation before submitting changes.
This update provides step-by-step guidance for both Linux/macOS and Windows users, ensuring a smoother onboarding experience for new contributors.

Fixes #2215

#### Checklist:

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -240,7 +240,8 @@ This will generate the HTML files in ``docs/_build/html/``.
 
 You can then open the documentation in your browser:
 
-- **macOS & Linux:** ``open docs/_build/html/index.html``
+- **Linux:** ``xdg-open docs/_build/html/index.html``
+- **macOS:** ``open docs/_build/html/index.html``
 - **Windows:** ``start docs\_build\html\index.html``
 
 *Tox automatically installs the necessary dependencies, so you don’t need


docs/contributing.rst
```diff
@@ -207,3 +209,75 @@ The release itself requires the following steps:
...
+----------------
+
+- If you encounter an error such as ``make: sphinx-build: Command not found``, install Sphinx manually:
+
+  .. code-block:: bash
+
+      pip install sphinx
+
+- If HTML files are not generated, review the build output for warnings or errors.
+- On Windows, if ``make`` is unavailable, use the Tox approach instead.
```
tim-schilling: I'd like to know why you feel that this section should be included. I'm open to keeping it, but my first instinct is to suggest removing it because the person should have the dependencies installed.

docs/contributing.rst
```diff
@@ -207,3 +209,75 @@ The release itself requires the following steps:
...
+.. code-block:: bash
+
+    tox -e docs -- html
+
+This will generate the HTML files in ``docs/_build/html/``.
+
+You can then open the documentation in your browser:
+
+- **macOS:** ``open docs/_build/html/index.html``
+- **Linux:** ``xdg-open docs/_build/html/index.html``
```
tim-schilling:
```suggestion
- **macOS & Linux:** ``open docs/_build/html/index.html``
```
Linux also supports `open`.

docs/contributing.rst
```diff
@@ -207,3 +209,75 @@ The release itself requires the following steps:
...
+Prerequisites
+--------------
+
+Before building the documentation, ensure that all dependencies are installed as described in
+:ref:`the setup instructions <code>`.
+If you plan to use *Tox*, make sure it is installed:
+
+.. code-block:: bash
+
+    pip install tox
```
tim-schilling: I think this should be removed since `tox` is included in our `requirements_dev.txt` file.

docs/contributing.rst
```diff
@@ -207,3 +209,75 @@ The release itself requires the following steps:
...
+Building the documentation locally
+----------------------------------
+
+The project's documentation is built using `Sphinx <https://www.sphinx-doc.org>`_.
+You can generate it locally to preview your changes before submitting a pull request.
+
+Two main approaches are supported:
+
+- **Make** – recommended for Linux/macOS users.
+- **Tox** – recommended for Windows users or for isolated, consistent builds.
```
tim-schilling: What do you think of removing `Make` as a suggested option? If tox works for all environments, maybe that's what we should suggest and let more inquisitive contributors discover the `make` commands on their own?

docs/contributing.rst
okotdaniel: Thanks @tim-schilling  for the feedback. my thinking for keeping the troubleshooting section is that it helps first-time contributors who may try to run make html directly and hit sphinx-build errors without realizing it’s dependency-related. i just thought it would act as a lightweight safeguard for those skipping tox or working outside virtual environments. That said, I’m happy to remove or condense it if we’d prefer to assume contributors already have everything set up which again i expect them too since all the dependancies are all taken care of in the `requirements_dev.txt`

docs/contributing.rst
okotdaniel: Sure,let me take care of that too.

docs/contributing.rst
okotdaniel: Adding that too.

docs/contributing.rst
tim-schilling: Hmm that's a good point. I think we can keep this, but adjust it slightly to remind them about tox and their virtualenv / requirements_dev.txt

I think rewording it to indicate if there's an error about a missing dependency, such as Sphinx Build then they should either activate their virtualenv and install dependencies or use tox.

docs/contributing.rst
okotdaniel: This is fine. i agree on this.

docs/contributing.rst
okotdaniel: Alright sure.

docs/contributing.rst
matthiask: > Linux also supports open.

I don't think that's generally true? Maybe I'm missing something, but `open` definitely doesn't exist on my system (an up-to-date installation of Fedora right now). The question and answer here is a bit old, but it still seems correct to me:
https://unix.stackexchange.com/questions/253376/open-command-to-open-a-file-in-an-application

tim-schilling: Ah, interesting. I've only used Ubuntu and made an assumption. Thank you for pointing that out.

tim-schilling:
I've rebased on main, fixed the pre-commit errors due to `doc8` (the lines were too long), and reverted my suggestion of combining the open commands for OSes.

I'm going to squash these commits and merge.

PR #2223, opened 2025-10-12
Deprecate RedirectsPanel in favor of the HistoryPanel

### LINKED ISSUE #2216: Retire RedirectsPanel in favor of HistoryPanel?
I believe the main purpose of the redirects panel was to allow the user to view the toolbar's data before the redirect occurred. However, with the HistoryPanel, it's now possible to switch to that request directly. If that's the only usecase for it, it's possible we could retire the RedirectsPanel.

JohananOppongAmoateng:
#### Description

this information as well.

Fixes #2216 

#### Checklist:

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/redirect.html
+++ b/debug_toolbar/templates/debug_toolbar/redirect.html
@@ -4,8 +4,68 @@
   <head>
     <title>Django Debug Toolbar Redirects Panel: {{ status_line }}</title>
     <script{% if toolbar.csp_nonce %} nonce="{{ toolbar.csp_nonce }}"{% endif %} type="module" src="{% static 'debug_toolbar/js/redirect.js' %}" async></script>
+    <style>
+      .djdt-deprecation-warning {
+        --djdt-warning-bg: #fff3cd;
+        --djdt-warning-text: #856404;
+        --djdt-warning-accent: #f1c40f;
+        --djdt-warning-link: #0056b3;
+        --djdt-warning-link-hover: #003d7a;
+        --djdt-warning-strong: #000;
+        --djdt-warning-shadow: rgba(0, 0, 0, 0.1);
+
+        position: relative;
+        background-color: var(--djdt-warning-bg);
+        color: var(--djdt-warning-text);
+        border-left: 0.25em solid var(--djdt-warning-accent);
+        padding: 1em;
+        margin: 1.5em 0;
+        border-radius: 0.25em;
+        box-shadow:
+          0 0.0625em 0.1875em var(--djdt-warning-shadow),
+          0 0.0625em 0.125em var(--djdt-warning-shadow);
+        font-family: var(--djdt-font-family-primary);
+        font-size: 0.875em;
+        line-height: 1.5;
+      }
+
+      .djdt-deprecation-warning::before {
+        content: "⚠️";
+        font-size: 1.125em;
+        margin-right: 0.75em;
+        vertical-align: middle;
+      }
+
+      .djdt-deprecation-warning a {
+        color: var(--djdt-warning-link);
+        text-decoration: none;
+        font-weight: 500;
+        border-bottom: 0.0625em solid currentColor;
+        padding-bottom: 0.0625em;
+        transition: all 0.2s ease;
+      }
+
+      .djdt-deprecation-warning a:hover {
+        color: var(--djdt-warning-link-hover);
+        border-bottom-width: 0.125em;
+      }
+
+      .djdt-deprecation-warning strong {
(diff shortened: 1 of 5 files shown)


tests/panels/test_redirects.py
```diff
@@ -100,3 +107,14 @@ def test_original_response_preserved(self):
         self.assertEqual(
             response.original_response.get("Location"), "http://somewhere/else/"
         )
+
+    def test_deprecation_warning(self):
+        """Test that a deprecation warning is shown when RedirectsPanel is instantiated."""
+        from debug_toolbar.toolbar import DebugToolbar
```
tim-schilling: Does this need to be a local import?

JohananOppongAmoateng: No will change it

tim-schilling:
Originally I was thinking it would be a message on the template that gets rendered for the `RedirectsPanel`. I'm not sure it's a significant improvement over having it in the terminal.

JohananOppongAmoateng:
Should I include it in the template too?

tim-schilling:
@JohananOppongAmoateng I'm good with this as is. I wanted to raise it to see what you thought.

elineda:
Hello, 
It's ok for me but I'm agree with your first though @tim-schilling. 

I think a message in the template it's not a lot of work and another warning for people who don't read the warning in console.

JohananOppongAmoateng:
Mm that's true I will add a warning in the template. Should it be something like a toast?

elineda:
> Mm that's true I will add a warning in the template. Should it be something like a toast?

Yep only if the user is on the template ofc.

debug_toolbar/templates/debug_toolbar/redirect.html
```diff
@@ -4,8 +4,12 @@
   <head>
     <title>Django Debug Toolbar Redirects Panel: {{ status_line }}</title>
     <script{% if toolbar.csp_nonce %} nonce="{{ toolbar.csp_nonce }}"{% endif %} type="module" src="{% static 'debug_toolbar/js/redirect.js' %}" async></script>
+    <link{% if toolbar.csp_nonce %} nonce="{{ toolbar.csp_nonce }}"{% endif %} rel="stylesheet" href="{% static 'debug_toolbar/css/toolbar.css' %}">
   </head>
-  <body>
+  <body id="djDebug" {% if toolbar.store_id %}data-store-id="{{ toolbar.store_id }}"{% endif %}>
```
tim-schilling: The `id="djDebug"` seems to be causing downstream effects. I wonder if it'd be better to have the styles defined in the `<head>` here directly?

debug_toolbar/templates/debug_toolbar/redirect.html
```diff
@@ -4,8 +4,12 @@
...
+    <div class="djdt-deprecation-warning">
+      <strong>{% translate "WARNING:" %}</strong> {% translate "The RedirectsPanel is deprecated and will be removed in a future version. The HistoryPanel now provides the ability to view toolbar data for redirected requests. If you still have a use case for this panel, please comment on this <a href='https://github.com/django-commons/django-debug-toolbar/issues/2216'>issue</a>" %}
```
tim-schilling:
```suggestion
      <strong>{% translate "WARNING:" %}</strong> {% translate "The RedirectsPanel is deprecated and will be removed in a future version. The HistoryPanel now provides the ability to view toolbar data for redirected requests. If you still have a use case for this panel, please comment on this <a target='_blank' href='https://github.com/django-commons/django-debug-toolbar/issues/2216'>issue</a>" %}
```

I also think a `{% blocktranslate %}` may make this a bit more readable since you could add more linebreaks to the code.

tim-schilling (changes requested):
The latest changes have caused the toolbar's side panel to disappear. I suspect it's due to trying to use the `id="djDebug"` attribute. My suggestion of using that as a prefix was wrong. This redirect HTML isn't actually within the toolbar's container, though we don't want custom styles floating around. We would need to identify a new container, like `djdt-redirect`. Though that said, I'm guessing that path is going to be more effort than we want to spend since this is deprecated. I think putting the styles in the `<head>` may be best. Though I wonder how that's going to go over with a CSP :grimacing:

debug_toolbar/templates/debug_toolbar/redirect.html
JohananOppongAmoateng: I think that would be a better approach since this panel is deprecated

tim-schilling:
Thank you for making those changes @JohananOppongAmoateng!

I removed the toolbar.css from the redirects template, cleaned up the indentation on the styles, squashed the commits and adjusted the commit message.

PR #2227, opened 2025-10-13
Add type hints to the toolbar and middleware modules

### LINKED ISSUE #1705: Add Type hints to the codebase

JohananOppongAmoateng:
#### Description
This PR adds type hints to the middleware and toolbar modules. It's based on the work of @leandrodesouzadev in this pr #1848 

Fixes #1705 

#### Checklist:

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -5,6 +5,7 @@ Debug Toolbar middleware
 import re
 import socket
 from functools import cache
+from typing import Callable
 
 from asgiref.sync import (
     async_to_sync,
@@ -25,7 +26,7 @@ from debug_toolbar.utils import clear_stack_trace_caches, is_processable_html_re
 _HTML_TYPES = ("text/html", "application/xhtml+xml")
 
 
-def show_toolbar(request: HttpRequest):
+def show_toolbar(request: HttpRequest) -> bool:
     """
     Default function to determine whether to show the toolbar on a given page.
     """
@@ -40,7 +41,7 @@ def show_toolbar(request: HttpRequest):
     return False
 
 
-def show_toolbar_with_docker(request: HttpRequest):
+def show_toolbar_with_docker(request: HttpRequest) -> bool:
     """
     Default function to determine whether to show the toolbar on a given page.
     """
@@ -72,7 +73,7 @@ def show_toolbar_with_docker(request: HttpRequest):
 
 
 @cache
-def show_toolbar_func_or_path():
+def show_toolbar_func_or_path() -> Callable:
     """
     Fetch the show toolbar callback from settings
 
@@ -87,7 +88,7 @@ def show_toolbar_func_or_path():
         return func_or_path
 
 
-def get_show_toolbar(async_mode):
+def get_show_toolbar(async_mode) -> Callable:
     """
     Get the callback function to show the toolbar.
 


JohananOppongAmoateng:
@tim-schilling @matthiask  I suggest we start using https://github.com/typeddjango/django-stubs for the typing

debug_toolbar/toolbar.py
```diff
@@ -39,24 +49,31 @@ def __init__(self, request, get_response, request_id=None):
...
+        # Use OrderedDict for the _panels attribute so that items can be efficiently
+        # removed using FIFO order in the DebugToolbar.store() method.  The .popitem()
+        # method of Python's built-in dict only supports LIFO removal.
+        # type: ignore[var-annotated]
+        self._panels = OrderedDict()
+        while panels:
+            panel = panels.pop()
+            self._panels[panel.panel_id] = panel
+        self.stats: dict[str, Any] = {}
+        self.server_timing_stats: dict[str, Any] = {}
```
tim-schilling: This feels like there may have been a merge conflict issue.

debug_toolbar/toolbar.py
```diff
@@ -72,15 +89,15 @@ def csp_nonce(self):
         """
         return getattr(self.request, "csp_nonce", None)
 
-    def get_panel_by_id(self, panel_id):
+    def get_panel_by_id(self, panel_id: str) -> "Panel":
```
tim-schilling: Sorry, I'm likely missing something basic. Why can't we use `-> Panel` since we have the `TYPE_CHECKING` import check?

debug_toolbar/toolbar.py
JohananOppongAmoateng: Yes this was due to a merge conflict

JohananOppongAmoateng:
@tim-schilling have you had time to look at the new changes?

debug_toolbar/toolbar.py
```diff
@@ -210,7 +219,7 @@ def __init__(self, request, get_response, request_id=None):
         self.init_store()
 
     @classmethod
-    def from_store(cls, request_id, panel_id=None):
+    def from_store(cls, request_id, panel_id=None) -> "StoredDebugToolbar":
```
tim-schilling: Should the rest of this signature (and `debug_toolbar_urls`) have types? Or how are you cutting things off?

debug_toolbar/middleware.py
```diff
@@ -86,7 +94,7 @@ def get_show_toolbar(async_mode):
     """
     Get the callback function to show the toolbar.
 
-    Will wrap the function with sync_to_async or
+     Will wrap the function with sync_to_async or
```
tim-schilling:
```suggestion
    Will wrap the function with sync_to_async or
```

debug_toolbar/toolbar.py
```diff
@@ -125,13 +132,10 @@ def fetch(cls, request_id, panel_id=None):
         if get_store().exists(request_id):
             return StoredDebugToolbar.from_store(request_id, panel_id=panel_id)
 
-    # Manually implement class-level caching of panel classes and url patterns
-    # because it's more obvious than going through an abstraction.
-
-    _panel_classes = None
+    _panel_classes: Optional[list[Panel]] = None
```
tim-schilling:
```suggestion
    # Manually implement class-level caching of panel classes and url patterns
    # because it's more obvious than going through an abstraction.
    
    _panel_classes: Optional[list[Panel]] = None
```

debug_toolbar/toolbar.py
```diff
@@ -140,10 +144,10 @@ def get_panel_classes(cls):
             cls._panel_classes = panel_classes
         return cls._panel_classes
 
-    _urlpatterns = None
+    _urlpatterns: Optional[list[URLPattern]] = None
 
     @classmethod
-    def get_urls(cls):
+    def get_urls(cls) -> list[URLPattern]:
```
tim-schilling:
```suggestion
    def get_urls(cls) -> list[URLPattern | URLResolver]:
```
Since this is a feature of third-party panel APIs, we should cover all cases. If another panel defines a `get_urls()` that uses an `include()`, that should be accounted for in the types.

debug_toolbar/toolbar.py
tim-schilling: You can do the following if you include `from __future__ import annotations` at the top. As of Python 3.14, it'll be deprecated.

```suggestion
    def from_store(cls, request_id, panel_id=None) -> StoredDebugToolbar:
```

debug_toolbar/panels/__init__.py
```diff
@@ -1,10 +1,15 @@
...
+
 from django.core.handlers.asgi import ASGIRequest
 from django.template.loader import render_to_string
 from django.utils.functional import classproperty
 
 from debug_toolbar import settings as dt_settings
 from debug_toolbar.utils import get_name_from_obj
 
+if TYPE_CHECKING:
+    from debug_toolbar._stubs import GetResponse
```
tim-schilling:
```suggestion
from debug_toolbar._stubs import GetResponse
```

From my understanding, it isn't required to check the `TYPE_CHECKING` for the stubs. Let me know if I'm wrong please.

debug_toolbar/middleware.py
```diff
@@ -108,7 +116,7 @@ class DebugToolbarMiddleware:
     sync_capable = True
     async_capable = True
 
-    def __init__(self, get_response):
+    def __init__(self, get_response: "GetResponse"):
```
tim-schilling:
```suggestion
    def __init__(self, get_response: GetResponse):
```

The `GetResponse` class is imported, so we don't need to use the string version.

debug_toolbar/panels/__init__.py
```diff
@@ -13,7 +18,7 @@ class Panel:
 
     is_async = False
 
-    def __init__(self, toolbar, get_response):
+    def __init__(self, toolbar, get_response: "GetResponse"):
```
tim-schilling:
```suggestion
    def __init__(self, toolbar, get_response: GetResponse):
```

The `GetResponse` class is imported so we don't need to use the string version.

debug_toolbar/toolbar.py
```diff
@@ -200,7 +207,9 @@ def from_store_get_response(request):
 
 
 class StoredDebugToolbar(DebugToolbar):
-    def __init__(self, request, get_response, request_id=None):
+    def __init__(
+        self, request: HttpRequest, get_response: "GetResponse", request_id=None
```
tim-schilling:
```suggestion
        self, request: HttpRequest, get_response: GetResponse, request_id=None
```

I think this is also missing the type definition for `request_id`

debug_toolbar/_stubs.py
```diff
@@ -1,9 +1,12 @@
 from __future__ import annotations
 
-from typing import Any, NamedTuple, Optional
+from typing import TYPE_CHECKING, Any, NamedTuple, Optional, Protocol
 
 from django import template as dj_template
 
+if TYPE_CHECKING:
+    from django.http import HttpRequest, HttpResponse
```
tim-schilling:
```suggestion
from django.http import HttpRequest, HttpResponse
```

It's not necessary to limit how often these two classes are imported.

debug_toolbar/middleware.py
```diff
@@ -13,14 +14,21 @@
...
+from django.http import HttpRequest, HttpResponse
 from django.utils.module_loading import import_string
 
 from debug_toolbar import settings as dt_settings
+from debug_toolbar.panels import Panel
 from debug_toolbar.toolbar import DebugToolbar
 from debug_toolbar.utils import clear_stack_trace_caches, is_processable_html_response
 
+if TYPE_CHECKING:
+    from debug_toolbar._stubs import GetResponse
```
tim-schilling:
```suggestion
from debug_toolbar._stubs import GetResponse
```

We shouldn't need to limit when we can import the stubs if I'm understanding things properly.

tim-schilling (changes requested):
This is coming along nicely, thank you for continuing to work on it. I think there are a few arguments in the method signatures that are missing. I think there may be too many `TYPE_CHECKING` protections as well, so we should be able to simplify things some more.

### Not to be solved right now:
I'm also realizing we need a way to evaluate this programmatically. It's going to be near impossible to keep this inline without some type checking.

debug_toolbar/toolbar.py
JohananOppongAmoateng: I was not really sure what to use for the type of request id that why i didnt type it

debug_toolbar/toolbar.py
JohananOppongAmoateng: I think they should but i was trying to do things incrementally

tim-schilling:
Had to redo the commit because I forgot to credit @leandrodesouzadev. That's been done in cf93187c253bc0d6f22772b599caa0f91c334c38 now.

PR #2228, opened 2025-10-14
Feature/django channels chat app

### LINKED ISSUE #2202: Add Django Channels integration to example app
Django Channels is an async library for Django that has a tutorial about building out a chat application. It's a common websocket interface. We should add that to our example app for easier manual testing. It would also serve as a way to prove that Debug Toolbar works with Django Channels

https://channels.readthedocs.io/en/latest/tutorial/index.html

References: #1413

Chiemezuo:
#### Description

This PR adds the [Django Channels'](https://channels.readthedocs.io/en/latest/) demo Chat app into the Debug Toolbar's example app for easy testing. This is in line to support async and show the Debug Toolbar's compatibility with Django Channels

Fixes #2202 

#### Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,13 +34,13 @@ repos:
       - id: biome-check
         verbose: true
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: 'v0.13.3'
+    rev: 'v0.14.0'
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
 -   repo: https://github.com/tox-dev/pyproject-fmt
-    rev: v2.7.0
+    rev: v2.10.0
     hooks:
       - id: pyproject-fmt
 -   repo: https://github.com/abravalheri/validate-pyproject
--- a/debug_toolbar/forms.py
+++ b/debug_toolbar/forms.py
@@ -3,7 +3,8 @@ import json
 from django import forms
 from django.core import signing
 from django.core.exceptions import ValidationError
-from django.utils.encoding import force_str
+
+from debug_toolbar.sanitize import force_str
 
 
 class SignedDataForm(forms.Form):
--- a/debug_toolbar/locale/en/LC_MESSAGES/django.po
+++ b/debug_toolbar/locale/en/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Django Debug Toolbar\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-08-06 07:12-0500\n"
+"POT-Creation-Date: 2025-10-16 19:11-0500\n"
 "PO-Revision-Date: 2012-03-31 20:10+0000\n"
 "Last-Translator: \n"
 "Language-Team: \n"
@@ -18,6 +18,14 @@ msgstr ""
 
 #: apps.py:18
 msgid "Debug Toolbar"
+msgstr ""
+
+#: models.py:11
+msgid "history entry"
+msgstr ""
+
+#: models.py:12
+msgid "history entries"
 msgstr ""
 
 #: panels/alerts.py:67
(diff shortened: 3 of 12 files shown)


Chiemezuo:
@tim-schilling I wasn't entirely sure whether to include the tests for the example app itself, seeing as it is something for testing, rather than a change to the Debug Toolbar itself. I was also unsure of the location to add the test.

JohananOppongAmoateng:
@Chiemezuo  I think we should add it to give the users of the example app that the code in there works as expected

Chiemezuo:
@JohananOppongAmoateng I think that's an excellent idea. The test requires `selenium` though. Would it be too much overhead? Since it has to simulate a browser environment.

example/urls.py
```diff
@@ -20,6 +21,8 @@
     ),
     path("jinja/", jinja2_view, name="jinja"),
     path("async/", async_home, name="async_home"),
+    path("async/chat/", async_chat_index, name="async_chat"),
+    path("async/chat/<str:room_name>/", async_chat_room, name="room"),
```
tim-schilling: I think these need go to in `async_/urls.py`

example/asgi.py
```diff
@@ -9,8 +9,23 @@
...
+# is populated before importing code that may import ORM models.
+django_asgi_app = get_asgi_application()
 
-application = get_asgi_application()
+application = ProtocolTypeRouter(
+    {
+        "http": django_asgi_app,
+        "websocket": AllowedHostsOriginValidator(
+            AuthMiddlewareStack(URLRouter(websocket_urlpatterns))
+        ),
```
tim-schilling, line 29: This is why the toolbar doesn't work with the channels' consumers. When you send a chat message via the websocket, that message isn't visible in the toolbar anywhere. Only the HTTP requests are. It bypasses the middleware that the toolbar works with.

To start, we would need to define a another middleware that is a subclass of `channels.middleware.BaseMiddleware`. There may be other issues that arise after that. However, I think this is enough for us to declare the toolbar doesn't support Django Channels' consumers yet :cry:

example/settings.py
```diff
@@ -70,6 +70,14 @@
...
 ASGI_APPLICATION = "example.asgi.application"
 
+CHANNEL_LAYERS = {
+    "default": {
+        "BACKEND": "channels_redis.core.RedisChannelLayer",
+        "CONFIG": {
+            "hosts": [("127.0.0.1", 6379)],
+        },
+    }
+}
```
tim-schilling: This should probably go in `async_/settings.py`

example/templates/chat/index.html
```diff
@@ -0,0 +1,27 @@
+<!-- example/templates/chat/index.html -->
```
tim-schilling:
```suggestion
{% comment %}
This template is from the Django Channels "Web Socket" chat app
https://channels.readthedocs.io/en/latest/tutorial/index.html
{% endcomment %}
```

example/templates/chat/room.html
```diff
@@ -0,0 +1,49 @@
+<!-- example/templates/chat/room.html -->
```
tim-schilling:
```suggestion
{% comment %}
This template is from the Django Channels "Web Socket" chat app
https://channels.readthedocs.io/en/latest/tutorial/index.html
{% endcomment %}
```

example/async_/consumers.py
```diff
@@ -0,0 +1,36 @@
+# This is the consumer logic for the Django Channels "Web Socket" chat app
```
tim-schilling, line 1:
```suggestion
# This is the consumer logic for the Django Channels "Web Socket" chat app
# https://channels.readthedocs.io/en/latest/tutorial/index.html
```

example/async_/routing.py
```diff
@@ -0,0 +1,8 @@
+# This is the routing logic for the Django Channels "Web Socket" chat app
```
tim-schilling, line 1:
```suggestion
# This is the routing logic for the Django Channels "Web Socket" chat app
# https://channels.readthedocs.io/en/latest/tutorial/index.html
```

example/templates/index.html
```diff
@@ -16,6 +16,7 @@ <h1>Index of Tests</h1>
         <li><a href="{% url 'turbo' %}">Hotwire Turbo</a></li>
         <li><a href="{% url 'htmx' %}">htmx</a></li>
         <li><a href="{% url 'bad_form' %}">Bad form</a></li>
+        <li><a href="{% url 'async_chat' %}">Chat app</a></li>
```
tim-schilling: Since the URLs will only be used for the async version, this probably will need to move down into the commented out area near `async_db_view`. It's not great, but I don't have a better solution at the moment.

tim-schilling:
@Chiemezuo I think in our meeting last week, we indicated we were going to shelve these changes on an issue. Does that sound right?

I think closing the PR and referring to your branch as long as you don't delete it, would be good. The next step was to create a new PR off of main that updates the documentation in the `architecture.rst` file to mention that the toolbar doesn't work with Channels' consumers.

Chiemezuo:
Yes @tim-schilling I just wanted to wrap up this PR properly, even though it would end up shelved. To not leave out any bits incase work continues on it later on.

tim-schilling:
Closing since the docs changes were merged in #2232 and #2231.

PR #2232, opened 2025-10-20
make doc corrections for django channels

### LINKED ISSUE #2231: Django Channels incompatibility
## Introduction

As a result of the work done in #2228 , it is somewhat safe to say that the current state of the Debug toolbar does not currently support Django Channels. While further inspection needs to be done to ascertain the required changes to enable compatitbility, @tim-schilling and I suspect the problem is related to the middleware class(es) used by Django Channels.

## Next Steps

- The PR can be shelved for now, but might be useful if/when work is done to enable compatibility.
- The documentation in the `architecture.rst` should include a disclaimer in the [Problematic parts](https://django-debug-toolbar.readthedocs.io/en/latest/architecture.html#problematic-parts) section that Django Channels is not yet compatible.
- The documentation in the [Django Channels & Async section](https://django-debug-toolbar.readthedocs.io/en/latest/installation.html#django-channels-async) of `installation.rst` should be corrected to state that there is currently no support for Django Channels.

Chiemezuo:
#### Description

Following up on PR #2228 , the documentation needed some correction to state that the Toolbar currently doesn't support Django Channels.

Fixes #2231 

#### Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
 
     services:
       mariadb:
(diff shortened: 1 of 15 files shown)


docs/installation.rst
```diff
@@ -273,8 +273,9 @@ And for Apache:
 Django Channels & Async
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-The Debug Toolbar currently has experimental support for Django Channels and
-async projects. The Debug Toolbar is compatible with the following exceptions:
+The Debug Toolbar currently has experimental support for async projects,
+but not for Django Channels. The Debug Toolbar is compatible with the
+following exceptions:
```
tim-schilling: This reads a bit weird with two blocks of exceptions. Maybe we just move the channels incompatibility down into the list of exceptions so it's just one list of ways the toolbar doesn't work with async projects?

docs/architecture.rst
```diff
@@ -85,3 +85,5 @@ Problematic Parts
   fully compatible and currently being worked on. For now, these panels
   are disabled by default when running in async environment.
   follow the progress of this issue in `Async compatible toolbar project <https://github.com/orgs/jazzband/projects/9>`_.
+- Django Channels incompatibility: The toolbar does not currently support
+  Django Channels' consumers.
```
tim-schilling:
```suggestion
  Django Channels' consumers due to the difference in middleware design.
```

The little contextual note may be helpful for a future reader if they don't realize to check the issues first.

Chiemezuo:
How's it looking now? @tim-schilling

tim-schilling:
Thank you @Chiemezuo!

Chiemezuo:
Thanks for the merge! @tim-schilling

PR #2235, opened 2025-10-22
Feature/task panel

### LINKED ISSUE #2204: Create Tasks panel
With Django Tasks being added to Django, we should add a panel to expose some meta data about an application's tasks usages.

Minimum requirements:
- Show message to install tasks if using  Django <6
- Show tasks queued as a part of the request

Other ideas:
- show all event information about the tasks that are running (would require DatabaseStore or similar to store data)
- show all registered tasks
- allow a task to be queued from the api
- allow retrying of a task
- performance comparisons?!

I think we should start with the minimum. If possible we should explore how difficult the rest would be and create issues for each with the findings.

Chiemezuo:
#### Description

This PR creates a panel for Django Tasks to allow the Debug Toolbar to monitor queued tasks.

Fixes #2204 

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/tasks.py
+++ b/debug_toolbar/panels/tasks.py
@@ -1,3 +1,4 @@
+import django
 from django.utils.translation import gettext_lazy as _, ngettext
 
 from debug_toolbar.panels import Panel
@@ -34,27 +35,23 @@ class TasksPanel(Panel):
 
     def enable_instrumentation(self):
         """Hook into task system to collect queued tasks"""
-        try:
-            import django
+        if self._tasks_available is False:
+            # Django tasks not available means that we cannot instrument
+            return
+        from django.tasks import Task
 
-            if django.VERSION < (6, 0):
-                return
-            from django.tasks import Task
+        print("[TasksPanel] instrumentation enabled:", hasattr(Task, "enqueue"))
 
-            print("[TasksPanel] instrumentation enabled:", hasattr(Task, "enqueue"))
+        # Store original enqueue method
+        if hasattr(Task, "enqueue"):
+            self._original_enqueue = Task.enqueue
 
-            # Store original enqueue method
-            if hasattr(Task, "enqueue"):
-                self._original_enqueue = Task.enqueue
+            def wrapped_enqueue(task, *args, **kwargs):
+                result = self._original_enqueue(task, *args, **kwargs).return_value
+                self._record_task(task, args, kwargs, result)
+                return result
 
-                def wrapped_enqueue(task, *args, **kwargs):
-                    result = self._original_enqueue(task, *args, **kwargs).return_value
-                    self._record_task(task, args, kwargs, result)
-                    return result
-
-                Task.enqueue = wrapped_enqueue
-        except (ImportError, AttributeError):
-            pass
+            Task.enqueue = wrapped_enqueue
 
     def _record_task(self, task, args, kwargs, result):
         """Record a task that was queued"""
@@ -75,13 +72,9 @@ class TasksPanel(Panel):
         except (ImportError, AttributeError):
             pass
 
-    def _check_tasks_available(self):
+    @property
+    def _tasks_available(self) -> bool:
         """Check if Django tasks system is available"""
-        try:
-            import django
-
-            if django.VERSION < (6, 0):
-                return False
-            return True
-        except (ImportError, AttributeError):
+        if django.VERSION < (6, 0):
             return False
+        return True
--- a/debug_toolbar/panels/tasks.py
+++ b/debug_toolbar/panels/tasks.py
@@ -1,0 +1,15 @@
+from django.utils.translation import gettext_lazy as _
+
+from debug_toolbar.panels import Panel
+
+
+class TasksPanel(Panel):
+    """
+    Panel that displays Django tasks queued or executed during the
+    processing of the request.
+    """
+
+    title = _("Tasks")
+    template = "debug_toolbar/panels/tasks.html"
+
+    is_async = True
--- a/debug_toolbar/panels/tasks.py
+++ b/debug_toolbar/panels/tasks.py
@@ -1,3 +1,5 @@
+import functools
+
 import django
 from django.utils.translation import gettext_lazy as _, ngettext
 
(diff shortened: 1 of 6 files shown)


debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+    def _check_tasks_available(self):
+        """Check if Django tasks system is available"""
+        try:
+            import django
+
+            if django.VERSION < (6, 0):
+                return False
+            return True
+        except (ImportError, AttributeError):
+            return False
```
robhudson, line 79: Is this to be used in the `enable_instrumentation` method? Seems like it would fit there. The `import django` can go at the top with the other imports and you can remove the exceptions for that. This project has django as a dependency so it's safe to assume Django is available.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,87 @@
...
+            # Store original enqueue method
+            if hasattr(Task, "enqueue"):
+                self._original_enqueue = Task.enqueue
+
+                def wrapped_enqueue(task, *args, **kwargs):
+                    result = self._original_enqueue(task, *args, **kwargs).return_value
+                    self._record_task(task, args, kwargs, result)
+                    return result
+
+                Task.enqueue = wrapped_enqueue
```
robhudson: This task wrapping looks great. That seems to be the main bits needed to collect stats for the toolbar panel. Nice!

example/async_/tasks.py
```diff
@@ -0,0 +1,19 @@
...
+    # Define a fallback decorator
+    def task(func=None, **kwargs):
+        def decorator(f):
+            return f
+
+        return decorator if func is None else decorator(func)
+
+
+@task
+def send_welcome_message(message):
```
easherma, line 13: I would suggest setting up some tasks that follow the Django examples tasks closely: 

https://docs.djangoproject.com/en/dev/topics/tasks/

That will help it feel a bit more 'real-world' and may inspire further ideas for functionality in the panel itself :-)

example/views.py
```diff
@@ -20,6 +21,13 @@ def jinja2_view(request):
 
 
 async def async_home(request):
```
easherma, line 24: I'm not _super_ familiar with this codebase, but when does this view get used? 

I'd defer to @tim-schilling and others, but it may make sense to have a dedicated view for triggering off example tasks vs, say, triggering every time you visit an example home page. That could also be a bit more self-documenting and give a place to build off of for the future (like testing larger queues, reporting results on a page template, etc)

debug_toolbar/templates/debug_toolbar/panels/tasks.html
```diff
@@ -0,0 +1,14 @@
+<table>
+    <thead>
+        <tr>
+            <th>Task Info</th>
```
easherma: This will need to be wrapped for translation :-)

example/async_/tasks.py
```diff
@@ -0,0 +1,19 @@
+try:
+    from django.tasks import task
+except ImportError:
+    # Define a fallback decorator
```
easherma, line 4: What are the cases where this could happen? I'm guessing just in older django versions?

Is a fallback decorator a common pattern in this case? I think if its an example meant to work with tasks it shouldn't need a fallback, but I'd be interested to hear more of your thought process. :-)

debug_toolbar/settings.py
```diff
@@ -81,6 +81,7 @@ def get_config():
     "debug_toolbar.panels.cache.CachePanel",
     "debug_toolbar.panels.signals.SignalsPanel",
     "debug_toolbar.panels.community.CommunityPanel",
+    "debug_toolbar.panels.tasks.TasksPanel",
```
easherma, line 85: @tim-schilling I wonder if it would make sense to check for the Django version here, and only include it if its on a Django version that supports the tasks?

robhudson (commented):
Great progress!

debug_toolbar/settings.py
tim-schilling: It could, but it's possible someone installs `django-tasks` in <6.0 version.

example/async_/tasks.py
matthiask: We have generally started preferring `django.VERSION` comparisons to handling `ImportError` in these cases since old branches can automatically be removed by django-upgrade in the former case.

example/views.py
tim-schilling: Yeah, it makes sense to have a separate view to test the tasks integration.

debug_toolbar/panels/tasks.py
Chiemezuo: Yes, that was the intention. I skipped it while playing around with stuff, but I'm incorporating it.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,80 @@
...
+        """Hook into task system to collect queued tasks"""
+        if self._tasks_available is False:
+            # Django tasks not available means that we cannot instrument
+            return
+        from django.tasks import Task
+
+        print("[TasksPanel] instrumentation enabled:", hasattr(Task, "enqueue"))
+
+        # Store original enqueue method
+        if hasattr(Task, "enqueue"):
```
tim-schilling: We can make some assumptions based on the library's API to the point where we don't need to check this.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,80 @@
...
+            self._original_enqueue = Task.enqueue
+
+            def wrapped_enqueue(task, *args, **kwargs):
```
tim-schilling: We should use the `functools.wrapper()` function here to preserve the docstrings of the wrapped function. You can see how it gets used in a similar fashion in panels/cache.py's `_monkey_patch_method` function.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,80 @@
...
+        self.record_stats(stats)
+
+    def enable_instrumentation(self):
+        """Hook into task system to collect queued tasks"""
+        if self._tasks_available is False:
+            # Django tasks not available means that we cannot instrument
+            return
+        from django.tasks import Task
+
+        print("[TasksPanel] instrumentation enabled:", hasattr(Task, "enqueue"))
```
tim-schilling: Looks like leftover a debugging line :grin:

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def disable_instrumentation(self):
+        """Restore original methods"""
+        try:
+            from django.tasks import Task
+
+            if hasattr(self, "_original_enqueue"):
+                Task.enqueue = self._original_enqueue
+        except (ImportError, AttributeError):
+            pass
```
tim-schilling, line 73: I'm not sure we should try to unwrap the method here in case something else has also wrapped it because that would be unwrapped too. Instead we could have the wrapped function call `_record_task` and then set that function to `None`, sort like how the CachePanel wraps the cache methods.

Another option is to use a Django signal and connect/disconnect from the signal handler sort of like the TemplatesPanel. Though we'd still need to do the monkey patching.

Give those two panel's patching methods a look and let me know if you have any questions.

tim-schilling (changes requested):
Thank you for continuing to work on this! I'm liking how it's turning out.

Chiemezuo:
Hi @tim-schilling 
Thanks for the feedback! I'll look through these in a bit and have some more things we could both look at!

debug_toolbar/panels/tasks.py
Chiemezuo: Great idea!

tim-schilling:
Replaced by #2407

PR #2243, opened 2025-11-07
Add cache view to example app

federicobond:
I added a cache view to the example app to more easily check the display of cache data in the debug toolbar.

### DIFF
--- a/example/views.py
+++ b/example/views.py
@@ -1,8 +1,8 @@
 import asyncio
 
 from asgiref.sync import sync_to_async
-from django.core.cache import cache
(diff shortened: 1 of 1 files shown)


example/views.py
```diff
@@ -40,3 +41,10 @@ async def async_db_concurrent(request):
     return await sync_to_async(render)(
         request, "async_db.html", {"user_count": user_count}
     )
+
+
+def cache_view(request):
+    cache.set("foo", "bar")
```
tim-schilling, line 47: TIL that Django sets a default timeout of 5m.

PR #2247, opened 2025-11-11
Add type hints for History Panel

JohananOppongAmoateng:
#### Description

Added type hinting for history panel code

Fixes # (issue)

#### Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,5 @@ updates:
           - "*"  # Group all Actions updates into a single larger pull request
     schedule:
       interval: weekly
+    cooldown:
+      default-days: 7
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,6 @@
+---
+changelog:
+  exclude:
+    authors:
+      - dependabot[bot]
+      - pre-commit-ci[bot]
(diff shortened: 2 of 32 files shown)


debug_toolbar/panels/history/panel.py
```diff
@@ -21,42 +22,42 @@ class HistoryPanel(Panel):
...
 
-    def generate_stats(self, request, response):
+    def generate_stats(self, request: HttpRequest, response: HttpResponse) -> None:
         try:
             if request.method == "GET":
-                data = request.GET.copy()
+                data: QueryDict = request.GET.copy()
             else:
-                data = request.POST.copy()
+                data: QueryDict = request.POST.copy()
```
tim-schilling: There's a `data = None` later that violates this type. I think you may want to extra out the definition to the top of this method as `data: QueryDict | None = None` and then in the exception block, replace `data = None` with `pass`.

JohananOppongAmoateng: thanks for catching that didn't see that. I will make the neccessary changes and commit

tim-schilling (approved):
Thank you @JohananOppongAmoateng

tim-schilling:
I squashed the commits and merged them through the web interface. I think it have mangled the author a tad.

PR #2252, opened 2025-11-14
Persist the credentials because coverage_comment needs them.

tim-schilling:
#### Description

This is a follow-up to #2245

#### Checklist:


tim-schilling:
If this doesn't break, we should follow up on the action. Zizmor really likes to advocate for the `persist_credentials = false`

tim-schilling:
Woohoo! Fixed the build. I owe doughnuts to the team for breaking it.

matthiask:
Thank you!

PR #2254, opened 2025-11-14
Bye Python 39

matthiask:
#### Description

Python 3.9 has reached its EOL date at the end of last month.

I think it's a good idea to remove it.

#### Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,6 @@ isolated_build = true
 envlist =
     docs
     packaging
-    py{310,311,312}-dj{42}-{sqlite,postgresql,postgis,mysql}
     py{310,311}-dj{42,51,52}-{sqlite,postgresql,psycopg3,postgis,mysql}
     py{312}-dj{42,51,52,60}-{sqlite,postgresql,psycopg3,postgis,mysql}
     py{313}-dj{51,52,60,main}-{sqlite,psycopg3,postgis3,mysql}


tox.ini
```diff
@@ -3,7 +3,7 @@ isolated_build = true
 envlist =
     docs
     packaging
-    py{39,310,311,312}-dj{42}-{sqlite,postgresql,postgis,mysql}
+    py{310,311,312}-dj{42}-{sqlite,postgresql,postgis,mysql}
     py{310,311}-dj{42,51,52}-{sqlite,postgresql,psycopg3,postgis,mysql}
```
tim-schilling: I think we have a lot of overlap here

tox.ini
```diff
@@ -3,7 +3,7 @@ isolated_build = true
...
     py{312}-dj{42,51,52,60}-{sqlite,postgresql,psycopg3,postgis,mysql}
```
tim-schilling:
```suggestion
    py{310,311}-dj{42,51,52}-{sqlite,postgresql,psycopg3,postgis,mysql}
    py{312}-dj{42,51,52,60}-{sqlite,postgresql,psycopg3,postgis,mysql}
```

tox.ini
matthiask: Yes, that's true, now that 39 is removed.

That said, I'm always confused reading the `envlist`. Two variables is somewhat fine, three variables is completely confusing 😂

tim-schilling: Yeah agreed, I feel like we end up reworking this matrix once a year because it's confusing.

PR #2258, opened 2025-11-19
Treat some warnings as errors in CI

tim-schilling:
#### Description

Explicitly ignores the RedirectPanel's deprecation warning.

#### Checklist:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -21,6 +21,8 @@ Pending
   limits. When ``SQLParseError`` is raised, the SQL panel now automatically
   disables grouping and retries formatting, preventing crashes with large
   queries.
+* Updated tox configuration to treat ``DeprecationWarning``,
+  ``ResourceWarning``, and ``PendingDeprecationWarning`` as errors.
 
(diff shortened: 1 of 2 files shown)


matthiask (approved):
I have to admit I'm confused. `-W error` would make warnings errors. `-W always` just makes us show warnings each time they are encountered, not just once per call location. 

The `tox.ini` change makes us not show warnings all the time anymore, and additionally also ignores the redirects panel warning... I think that's fine, since using a warnings eating context manager is hard here. 

However, the changelog entry doesn't seem to match what we're doing here?

tim-schilling:
I forgot to update the changelog after I switched away from error. I'm going to change this though to follow what Adam does with [django-mysql](https://github.com/adamchainz/django-mysql/blob/main/tox.ini) then ignore our specific warning.

```
      -W error::ResourceWarning \
      -W error::DeprecationWarning \
      -W error::PendingDeprecationWarning \
```

matthiask (approved):
Makes sense! Obviously we still have some problems to fix or to ignore if we want to merge this.

PR #2259, opened 2025-11-19
Added test to confirm assertNumQueries

### LINKED ISSUE #1791: Missing connection.queries

tim-schilling:
#### Description

This was confirmed to be a valid test in e7575e8 when it was a known issue.

Fixes #1791

#### Checklist:

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -325,6 +325,10 @@
     margin-bottom: 3px;
 }
 
+#djDebug pre span {
+    font-family: var(--djdt-font-family-monospace);
+}
+
 #djDebug .djdt-panelContent {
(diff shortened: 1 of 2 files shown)


federicobond (approved):

matthiask:
Great, thank you!

PR #2263, opened 2025-11-21
Fix string representation of values in settings panel

federicobond:
#### Description

There was some weird double quoting going on in the settings panel values column. This fixes the formatting so it's more consistent and clearly distinguishes string from non-string values. Added some tests for the new behavior.

#### Before

Captura de pantalla 2025-11-21 a la(s) 9 07 48 p  m

#### After

Captura de pantalla 2025-11-21 a la(s) 8 43 21 p  m

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/settings.py
+++ b/debug_toolbar/panels/settings.py
@@ -1,10 +1,18 @@
+from pprint import pformat
+
 from django.utils.translation import gettext_lazy as _
 from django.views.debug import get_default_exception_reporter_filter
 
 from debug_toolbar.panels import Panel
-from debug_toolbar.sanitize import force_str
 
 get_safe_settings = get_default_exception_reporter_filter().get_safe_settings
+
+
+def safe_pformat(obj):
+    try:
+        return pformat(obj)
+    except Exception as e:
+        return f"<unformattable {type(obj).__name__}: {e!r}>"
 
 
 class SettingsPanel(Panel):
@@ -27,7 +35,7 @@ class SettingsPanel(Panel):
         self.record_stats(
             {
                 "settings": {
-                    key: force_str(value)
+                    key: safe_pformat(value)
                     for key, value in sorted(get_safe_settings().items())
                 }
             }
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -16,6 +16,7 @@ Pending
   ``debug_toolbar.store.DatabaseStore`` with ``SKIP_TOOLBAR_QUERIES``.
 * Fixed font family for code blocks and stack traces in the toolbar.
 * Added test to confirm Django's ``TestCase.assertNumQueries`` works.
+* Fixed string representation of values in settings panel.
 
 6.1.0 (2025-10-30)
 ------------------
--- a/debug_toolbar/templates/debug_toolbar/panels/settings.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/settings.html
@@ -10,7 +10,7 @@
     {% for name, value in settings.items %}
       <tr>
         <td>{{ name }}</td>
-        <td><code>{{ value|pprint }}</code></td>
+        <td><code>{{ value }}</code></td>
       </tr>
     {% endfor %}
   </tbody>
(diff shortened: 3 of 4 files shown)


debug_toolbar/panels/settings.py
```diff
@@ -27,7 +28,7 @@ def generate_stats(self, request, response):
         self.record_stats(
             {
                 "settings": {
-                    key: force_str(value)
+                    key: pformat(value)
```
matthiask: Are you sure that `force_str` can just be removed? I think we added that to not crash when `value` contains binary data or something. Maybe `pformat` handles that just fine.

docs/changes.rst
```diff
@@ -16,6 +16,7 @@ Pending
   ``debug_toolbar.store.DatabaseStore`` with ``SKIP_TOOLBAR_QUERIES``.
 * Fixed font family for code blocks and stack traces in the toolbar.
 * Added test to confirm Django's ``TestCase.assertNumQueries`` works.
+* Fixed display of values in settings panel.
```
matthiask: Maybe more specific? Not sure if that's an improvement, just a suggestion.

```suggestion
* Fixed double quoting of values in settings panel.
```

debug_toolbar/panels/settings.py
federicobond: Not sure. It should return a string for most well-behaved objects but I can wrap it so it tolerates exceptions.

federicobond:
@matthiask the attached screenshot shows how it looks with this change. I added one with the current behavior just below.

debug_toolbar/panels/settings.py
matthiask: It has just been added here: https://github.com/django-commons/django-debug-toolbar/pull/2217/files

The `bytes.fromhex` value would probaly be a good test for this change.

matthiask:
> @matthiask the attached screenshot shows how it looks with this change. I added one with the current behavior just below.

I'm sorry if I'm being dense but I only see one screenshot?

debug_toolbar/panels/settings.py
federicobond: Done!

docs/changes.rst
federicobond: Would this work better?

> Fixed string representation of values in settings panel

federicobond:
How about now?

debug_toolbar/panels/settings.py
federicobond: pprint.pformat would format that bytes value as:

```
(b'\xa3\xf2\xb8\xc1N\x97-Z\x8f\xb3\xc7)\x1ad\xe0\x85\x9cG+\xf6=\x18\xa0\x94'
 b'^s\xb2\xc8O\x91z\xe2')
```
instead of `"Django Debug Toolbar was unable to parse value."`

docs/changes.rst
federicobond: @matthiask I've updated the commit message to be more specific. Let me know what you think.

debug_toolbar/panels/settings.py
```diff
@@ -1,12 +1,20 @@
...
 
 from debug_toolbar.panels import Panel
-from debug_toolbar.sanitize import force_str
 
 get_safe_settings = get_default_exception_reporter_filter().get_safe_settings
 
 
+def safe_pformat(obj):
+    try:
+        return pformat(obj)
```
tim-schilling, line 13: I'm wondering if maybe we should roll this into `force_str` as `force_str(pretty=True)`?

federicobond: I don't think they have the same semantics. In this case, for example, I would very much like to distinguish `False` (bool) from `"False"` (str), and `b"Foo"` (bytes) from `"Foo"` (str).

`force_str` map the first two into `False` and the other two into `Foo`, thus losing that important information.

tim-schilling: Thank you for clarifying that. I think in most cases where `force_str` is currently used we want to capture that distinction. Only in `SignedDataForm`'s usage of `force_str` would we not want that. We can do that in a follow-up change though.

tim-schilling (approved):

PR #2266, opened 2025-11-26
Use shadow dom to render the toolbar

federicobond:
#### Description

Following the discussion on https://github.com/django-commons/django-debug-toolbar/issues/2007, I started experimenting with shadow DOM and it turns out it's pretty easy to make it work with the debug toolbar.

Most of the changes are related to getting the correct root for querying elements in JS, and those could be extracted to a separate pull request to make it easy to review the changes specific to shadow DOM.

The only change required for custom panels is that they should access the root debug element via the `getDebugElement` function in `utils.js` instead of directly querying `#djDebug`, and they should perform all element queries from that root.

This change effectively isolates the Debug Toolbar styles from the rest of the page, which would make it easier to implement all sorts of utility classes for more advanced panel styling.

#### Checklist:

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -1,5 +1,5 @@
 /* Variable definitions */
-:host {
+#djDebug {
     /* Font families are the same as in Django admin/css/base.css */
     --djdt-font-family-primary:
         "Segoe UI", system-ui, Roboto, "Helvetica Neue", Arial, sans-serif,
@@ -13,7 +13,7 @@
         "Noto Color Emoji";
 }
 
-:host,
+#djDebug,
 #djDebug[data-theme="light"] {
     --djdt-font-color: black;
     --djdt-background-color: white;
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -1,5 +1,5 @@
 /* Variable definitions */
-:root {
+:host {
     /* Font families are the same as in Django admin/css/base.css */
     --djdt-font-family-primary:
         "Segoe UI", system-ui, Roboto, "Helvetica Neue", Arial, sans-serif,
@@ -13,7 +13,7 @@
         "Noto Color Emoji";
 }
 
-:root,
+:host,
 #djDebug[data-theme="light"] {
     --djdt-font-color: black;
     --djdt-background-color: white;
--- a/debug_toolbar/static/debug_toolbar/js/utils.js
+++ b/debug_toolbar/static/debug_toolbar/js/utils.js
@@ -76,9 +76,10 @@ function getDebugElement() {
     // everywhere the element is being selected. A fixed reference
     // to the element should be avoided because the entire DOM could
     // be reloaded such as via HTMX boosting.
-    const root = document.getElementById(
-        "djDebugShadowRootContainer"
-    ).shadowRoot;
+    let root = document.getElementById("djDebugRoot");
+    if (root.shadowRoot) {
+        root = root.shadowRoot;
+    }
     return root.querySelector("#djDebug");
 }
 
--- a/debug_toolbar/static/debug_toolbar/js/utils.js
+++ b/debug_toolbar/static/debug_toolbar/js/utils.js
@@ -76,7 +76,9 @@ function getDebugElement() {
     // everywhere the element is being selected. A fixed reference
     // to the element should be avoided because the entire DOM could
     // be reloaded such as via HTMX boosting.
-    const root = document.getElementById("djDebugRoot").shadowRoot;
+    const root = document.getElementById(
+        "djDebugShadowRootContainer"
+    ).shadowRoot;
     return root.querySelector("#djDebug");
 }
 
--- a/debug_toolbar/static/debug_toolbar/js/utils.js
+++ b/debug_toolbar/static/debug_toolbar/js/utils.js
@@ -70,12 +70,14 @@ const $$ = {
     },
 };
 
+/**
+ * Fetch the debug element from the DOM.
+ *
+ * This is used to avoid writing the element's id everywhere the element
+ * is being selected. A fixed reference to the element should be avoided
+ * because the entire DOM could be reloaded such as via HTMX boosting.
+ */
 function getDebugElement() {
-    // Fetch the debug element from the DOM.
-    // This is used to avoid writing the element's id
-    // everywhere the element is being selected. A fixed reference
-    // to the element should be avoided because the entire DOM could
-    // be reloaded such as via HTMX boosting.
     let root = document.getElementById("djDebugRoot");
     if (root.shadowRoot) {
         root = root.shadowRoot;
--- a/debug_toolbar/static/debug_toolbar/js/utils.js
+++ b/debug_toolbar/static/debug_toolbar/js/utils.js
@@ -70,6 +70,16 @@ const $$ = {
     },
 };
 
+function getDebugElement() {
+    // Fetch the debug element from the DOM.
+    // This is used to avoid writing the element's id
+    // everywhere the element is being selected. A fixed reference
+    // to the element should be avoided because the entire DOM could
+    // be reloaded such as via HTMX boosting.
+    const root = document.getElementById("djDebugRoot").shadowRoot;
+    return root.querySelector("#djDebug");
+}
+
 function ajax(url, init) {
     return fetch(url, Object.assign({ credentials: "same-origin" }, init))
         .then((response) => {
@@ -89,7 +99,8 @@ function ajax(url, init) {
             );
         })
         .catch((error) => {
-            const win = document.getElementById("djDebugWindow");
+            const djDebug = getDebugElement();
+            const win = djDebug.querySelector("#djDebugWindow");
             win.innerHTML = `<div class="djDebugPanelTitle"><h3>${error.message}</h3><button type="button" class="djDebugClose">»</button></div>`;
             $$.show(win);
             throw error;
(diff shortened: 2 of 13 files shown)


debug_toolbar/templates/debug_toolbar/base.html
```diff
@@ -1,43 +1,48 @@
...
-     {{ toolbar.config.ROOT_TAG_EXTRA_ATTRS|safe }}  data-update-on-fetch="{{ toolbar.config.UPDATE_ON_FETCH }}">
-  <div class="djdt-hidden" id="djDebugToolbar">
-    <ul id="djDebugPanelList">
-      <li><a id="djHideToolBarButton" href="#" title="{% translate 'Hide toolbar' %}">{% translate "Hide" %} »</a></li>
-      <li>
-        <a id="djToggleThemeButton" href="#" title="{% translate 'Toggle Theme' %}">
-          {% translate "Toggle Theme" %} {% include "debug_toolbar/includes/theme_selector.html" %}
-        </a>
-      </li>
+<div id="djDebugRoot" {{ toolbar.config.ROOT_TAG_EXTRA_ATTRS|safe }}>
```
tim-schilling, line 2:
```suggestion
<div id="djDebugShadowRootContainer" {{ toolbar.config.ROOT_TAG_EXTRA_ATTRS|safe }}>
```

If you don't mind a bit of bikeshedding, I think we can be pretty explicit about what this element is for. If I'm interpreting things wrong, please let me know.

debug_toolbar/templates/debug_toolbar/base.html
```diff
@@ -1,43 +1,48 @@
...
+        data-request-id="{{ toolbar.request_id }}"
+        data-render-panel-url="{% url 'djdt:render_panel' %}"
+        {% endif %}
+        {% url 'djdt:history_sidebar' as history_url %}
+        {% if history_url %}
+        data-sidebar-url="{{ history_url }}"
+        {% endif %}
+        data-default-show="{% if toolbar.config.SHOW_COLLAPSED %}false{% else %}true{% endif %}"
+        data-update-on-fetch="{{ toolbar.config.UPDATE_ON_FETCH }}">
+      <link{% if toolbar.csp_nonce %} nonce="{{ toolbar.csp_nonce }}"{% endif %} rel="stylesheet" href="{% static 'debug_toolbar/css/toolbar.css' %}">
```
tim-schilling: Is there a reason this is being added back in? I think this may be a mistake, but it's certainly possible I'm not understanding things properly.

tim-schilling (changes requested):
Oh dang, this was a much simpler implementation than I expected. Thank you @federicobond!

I think we may want to make a mention of how to fetch JS elements in the JavaScript documentation somewhere. I'm going to assume, this will also impact any third-party panels that have some JS that looks for elements in the DOM? That may make things a bit tricky to support backwards compatibility for a version or two 🤔 What do you think?

debug_toolbar/templates/debug_toolbar/base.html
federicobond: It's a mistake! Good catch.

debug_toolbar/templates/debug_toolbar/base.html
federicobond: Sure, don't mind.

tim-schilling: Cool, thank you. I pushed a follow-up commit to adjust the other spot. I missed that. I'll squash that out on merge later if you don't clean it that.

federicobond:
@tim-schilling I think we could take this approach:

**Phase 1:** Update the built-in JS in this version to support both approaches and add a note for custom panel developers to query elements from the root using the new `getDebugElement` function from `utils.js`. While we are here, we could add an option for early adopters to opt-in to shadow DOM.

**Phase 2:** Make Shadow DOM the default but allow users to opt-out if they experience broken custom panels.

This would give panel developers plenty of time to make their panels compatible with a shadow DOM toolbar and allow us to get feedback from early adopters.

tim-schilling:
I like it. That sounds like it'll work well. I think the last area of concern is CSP. We'll need to do some investigation into that too.

federicobond:
Cool! I went ahead and wrote some more changes to support opt-in shadow DOM via the `USE_SHADOW_DOM` setting.

As for CSP, I tested a CSP config with `"default-src": [CSP.SELF, CSP.NONCE]` which worked correctly with shadow DOM. The way the scripts and styles are loaded does not change whether we are using shadow DOM or not so it makes sense, but I'm not an expert on CSP.

Btw, we might want to get that CSP config into the example app to ease compliance but not sure how we would like to handle older Django versions in CI.

matthiask (approved):
LGTM. I wonder if we shouldn't flip the default and allow people to opt out? Bugs would probably be the only reason why anyone would want to **not** use shadow DOM when we support it?

tim-schilling:
That's a good point @matthiask. I'm good with that.

federicobond:
It will break any custom panels that use JS, probably, but it's not hard to fix. Should I go ahead and make it the default?

federicobond:
Rebased and made shadow DOM opt-out instead of opt-in. Should be ready to merge now.

docs/changes.rst
```diff
@@ -36,6 +36,8 @@ Pending
 * Added test to confirm Django's ``TestCase.assertNumQueries`` works.
 * Fixed string representation of values in settings panel.
 * Declared support for Django 6.0.
+* Added opt-in support for rendering the toolbar in a shadow DOM for better
+  isolation from the rest of the page.
```
matthiask: This says "opt-in support" even though the default has been changed.

matthiask (approved):
I think this is good to go except for the note in the changelog and the merge conflicts we now have.

@tim-schilling You submitted a "changes requested" review for an earlier version of this, do you have the time to look at it again?

docs/changes.rst
```suggestion
* Updated to render the toolbar in a shadow DOM for better isolation
  from the rest of the page. This can be disabled with the setting
  ``USE_SHADOW_DOM``.
```

docs/configuration.rst
```diff
@@ -439,7 +439,7 @@ could add a **debug_toolbar/base.html** template override to your project:
 
     {% block css %}{{ block.super }}
     <style>
-        :root {
+        :host {
```
tim-schilling: This doesn't work because the `#djDebug` selector takes priority. Regardless though, we should call this out in the release notes. The original `:root` selector won't work after this change.

debug_toolbar/static/debug_toolbar/css/toolbar.css
```diff
@@ -1,5 +1,5 @@
 /* Variable definitions */
-:root {
+#djDebug {
```
tim-schilling, line 2: Is this intended to be this value or `:host`? I saw in the theme documentation example you used `:host`, but that's incompatible with this. I quickly changed this and the next `#djDebug` to `:host` and things _seemed_ to work. I think using `:host` over the id selector is preferred since it's less of a change in terms of priority for css.

tim-schilling: Ohhh. I'm guessing if we aren't using the shadow DOM that `:host` won't work.

tim-schilling (commented):
Right now I'd like to see a few more changes, but I'm happy for someone to push back and say I'm nitpicking. I'll see if I can dismiss the requested changes without approving.

In general, I'm trying to be better about maintaining the changelog for changes. I added a more concrete suggestion for the shadow DOM feature. I think we also want to mention how themeing support will change. It's fine to just say, "Hey if you're using themes, it likely will break. Go read Theme Support to see how to do it now", but phrased better :grin:

debug_toolbar/static/debug_toolbar/js/utils.js
```diff
@@ -70,6 +70,19 @@ const $$ = {
     },
 };
 
+function getDebugElement() {
+    // Fetch the debug element from the DOM.
```
codingjoe: JSDocs is a nice format for Python like docstrings.

codingjoe:
@tim-schilling no let's take the extra minute. 


I'd love to deprecate the old solution. Maintaining both will be a nightmare long-term and complexity we don't need. If we can't switch switch confidence, deprecation is the next best thing.

A simple deprecation warnings when the setting is false should do the trick, right?

debug_toolbar/static/debug_toolbar/css/toolbar.css
federicobond: Right, it must work with and without shadow DOM. The docs likely need updating.

federicobond:
Agree we likely don't want to support both long-term. The decision would be between deprecating non-shadow DOM support with the next release or after the dust has settled.

*Edit:* I can get on board with deprecating it this release.

debug_toolbar/static/debug_toolbar/css/toolbar.css
tim-schilling: Sounds good. Yeah, then we need to update the docs and the change log. I'll resolve this thread.

tim-schilling:
> The decision would be between deprecating non-shadow DOM support with the next release or after the dust has settled.

I'd lean towards not deprecating this release (let's get this out the door), but it's a slight preference. If someone wants to spend the time doing the work to deprecate it now, that's fine.

I think we should use a major version number change for this PR though because we're breaking theme support and making a significant change to the tool.

docs/changes.rst
```diff
@@ -8,6 +8,10 @@ Pending
   fetch requests.
 * Added a note to the prerequisites section of the installation docs
   about requiring an up-to-date browser.
+* Updated to render the toolbar in a shadow DOM for better isolation
+  from the rest of the page. This can be disabled with the setting
+  ``USE_SHADOW_DOM``. Note that this may break custom themes and panes
+  that have not yet been updated for shadow DOM support.
```
tim-schilling: I'm not sure this is accurate. From my understanding, any custom theme that overrides something in the new `#djDebug` style defintion will need an update, regardless if the shadow DOM is being used. Does that match your understanding?

Edit: "The new `#djDebug` and old `:root`*"

federicobond: That's correct, I was probably too quick to trade accuracy for brevity. What about this?

```suggestion
  ``USE_SHADOW_DOM``. Note that custom themes overriding CSS variables on :root must
  move those overrides to #djDebug, and custom panels that rely on external styles or DOM
  lookups reaching into the toolbar will need updates to work with the shadow DOM.
```

tim-schilling: I think that's good, but let's move it to it's own bullet point. That way it's less likely a person will skip over it.

codingjoe:
IMHO, if we know now that we want to remove it. Telling the community early seems preferable to me. We can still decide to deprecate targeting the next-next major release but still announce it now, right?

tim-schilling (approved):
Thank you for your work on this @federicobond!

PR #2267, opened 2025-11-27
Add Django built-in CSP nonce support

ahumeau:
#### Description

Django 6.0 added built-in support for Content Security Policies.
Part of this support is associating a randomly generated nonce to each request that can then be
attached to `<script>` and `<style>` tags.

If users of django-debug-toolbar want to have a nonce-based CSP, they need django-debug-toolbar to add the nonce to the `<script>` and `<style>` tags that it generates.

django-debug-toolbar already does that for CSP nonces generated by the django-csp third-party lib but the nonces generated by django-csp and Django are accessed differently.

This PR adds support for the CSP nonces generated by the built-in Django implementation.

### DIFF
--- a/debug_toolbar/_compat.py
+++ b/debug_toolbar/_compat.py
@@ -1,3 +1,5 @@
+import django
+
 try:
     from django.contrib.auth.decorators import login_not_required
 except ImportError:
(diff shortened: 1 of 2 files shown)


matthiask (changes requested):
Thanks! This looks very good.

Please change the CSP nonce import to use a `django.VERSION` comparison block so that we can automatically remove it in the future using django-upgrade (https://django-upgrade.readthedocs.io/en/latest/fixers.html#versioned-blocks)

PR #2277, opened 2025-12-11
remove requirement_dev.txt from project

p-r-a-v-i-n:
since we are using pyproject.toml depending on extra file for dependency is redundant. we should use the power of pyproject.toml to take care of it.

according [PEP 735](https://peps.python.org/pep-0735/) we can use dependency groups which is more convenient.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -4,7 +4,7 @@ Change log
 Pending
 -------
 
-* Replaced requirements_dev.txt file for pyproject.toml support with
+* Replaced ``requirements_dev.txt`` file for ``pyproject.toml`` support with
   dependency groups.
 
 6.2.0 (2026-01-20)
--- a/tox.ini
+++ b/tox.ini
@@ -44,7 +44,6 @@ setenv =
 allowlist_externals = make
 pip_pre = True
 dependency_groups =
-    docs
     dev
 commands = python -b -W always -m coverage run -m django test -v2 {posargs:tests}
 
@@ -79,10 +78,8 @@ setenv =
 
 [testenv:docs]
 commands = make -C {toxinidir}/docs {posargs:spelling}
-deps =
-    Sphinx
-    sphinxcontrib-spelling
-    sphinx-rtd-theme
+dependency_groups =
+    docs
 
 [testenv:packaging]
 commands =


.readthedocs.yaml
```diff
@@ -14,6 +14,7 @@ sphinx:
 
 python:
   install:
-  - requirements: requirements_dev.txt
   - method: pip
-    path: .
+  - path: .
+    extra_requirements:
+      - dev
```
p-r-a-v-i-n: well i'm not confident about this part. I would appreciate any help ?

pyproject.toml
```diff
@@ -48,6 +48,41 @@ urls.Homepage = "https://github.com/django-commons/django-debug-toolbar"
 urls.Issues = "https://github.com/django-commons/django-debug-toolbar/issues"
 urls.Source = "https://github.com/django-commons/django-debug-toolbar"
 
+[dependency-groups]
```
tim-schilling, line 51: Can you please rework the comments here to be more similar to what was used in the requirements file? It looks like things got jostled around.

.readthedocs.yaml
tim-schilling: You're correct to be wary about this. It is used to install extra requirements (optional dependencies in TOML), not dependency groups.

Instead, we'd need to either have a docs/requirements.txt file, install the dev requirements in the `post_install` [build job](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-jobs) of read the docs, or something else entirely.

p-r-a-v-i-n: [build.commands](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-commands)

can we use something similar 
for example:

```
version: 2

build:
  os: ubuntu-24.04
  tools:
    python: "3.10"
  commands:
    - pip install --group dev

sphinx:
  configuration: docs/conf.py
```

how could I test this locally whether it is picking it up or not ?

p-r-a-v-i-n: I don't know but  i'm missing something from docs, bcs i'm really bad reading docs, it takes me time to find the right knob. But currently i see that they haven't covered "dependency-group" yet.

p-r-a-v-i-n: > Instead, we'd need to either have a docs/requirements.txt file, install the dev requirements in the post_install [build job](https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-jobs) of read the docs

I like your suggestion. I think if  we succeed in `post_install` then it would be great, but either way also fine.

pyproject.toml
p-r-a-v-i-n: I will rework this. I think `pyproject-fmt` jostled the comments around as it sorted dependecies.

.readthedocs.yaml
tim-schilling: > how could I test this locally whether it is picking it up or not ?

One idea is to create a separate read the docs integration for testing using the code from the toolbar project.

.readthedocs.yaml
```diff
@@ -8,12 +8,8 @@ build:
...
+  commands:
+    - pip install --group docs
 
 sphinx:
   configuration: docs/conf.py
-
-python:
-  install:
-  - requirements: requirements_dev.txt
-  - method: pip
```
p-r-a-v-i-n, line 18: I think part was just to install the dependencies only.

.readthedocs.yaml
```diff
@@ -8,12 +8,8 @@ build:
   os: ubuntu-24.04
   tools:
     python: "3.10"
+  commands:
+    - pip install --group docs
```
p-r-a-v-i-n: i think now this should work

tim-schilling: Thanks for adjusting things. Did you happen to test this?

pyproject.toml
```diff
@@ -48,6 +48,41 @@ urls.Homepage = "https://github.com/django-commons/django-debug-toolbar"
...
+  "jinja2",
+  "sqlparse",
+]
+django_async = [
+  "daphne",     # Integration support
+  "whitenoise",
+]
+other = [
+  "pre-commit",
+]
```
tim-schilling: I'm guessing you created the extra groups to satisfy the ask for an organized list of dependencies? I'm  hesitant about changing a public facing interfacing to satisfy an internal need. Could the formatter be disabled for this section instead?

p-r-a-v-i-n: I could merged everything in single group `dev` instead of seperates groups. Except `docs`.

.readthedocs.yaml
p-r-a-v-i-n: No, I haven't test this. bcs i have no idea to test this locally.

p-r-a-v-i-n:
Hey @tim-schilling ,
I like working with you. I like your insights. and your feedback really helps me understand things even more better.

Currently I have some thoughts in my mind.

1. I have strong believe that for building docs it will pick `pip install --group docs`, but it just assumption no strong evidence ( bcs no experience).
2. I'm not sure whether this changes are absolutely requires at this movement. Honestly, I submitted this PR for my personal milestone , and I understood personal milestone should not be greater than the  project need. Project requirement should be the first priority.

I'm happy to iterate as per the feedback. and I'm happy answer/discuss questions. Thank you.

pyproject.toml
p-r-a-v-i-n: pyproject-fmt currently does not support ignoring specific section.

pyproject.toml
```diff
@@ -48,6 +48,31 @@ urls.Homepage = "https://github.com/django-commons/django-debug-toolbar"
...
+  "sqlparse",
+  "tox",
+  "whitenoise", # To avoid dealing with static files
+
+]
+docs = [
+  "sphinx",
+  "sphinx-rtd-theme>1",
+  "sphinxcontrib-spelling",
+]
```
p-r-a-v-i-n, line 75: I kept docs group bcs building docs we don't require other dependencies.

tim-schilling: Makes sense!

pyproject.toml
tim-schilling: Alright, I think we can live with that.

docs/contributing.rst
```diff
@@ -267,7 +267,7 @@ activated and all dependencies are installed:
 
 .. code-block:: bash
 
-    pip install -r requirements_dev.txt
+    pip install --group dev
```
tim-schilling:
```suggestion
    pip install --group dev --group docs
```

This is referring to the documentation dependencies. I suspect we need to change all the references? Or have the dev group also include the docs group.

tim-schilling (commented):
This is a different workflow than I'm used to, though I think it's close to a mergeable state. @matthiask or I will need to check on the readthedocs integration after this gets merged. It's probably a good idea regardless since it's been a while since I've looked at that.

I'd like to get opinions from @django-commons/django-debug-toolbar on whether this is +1, 0 or -1 type thing. For me, I'm very used to `pip install -r requirements_dev.txt`. I suspect I will get tripped up by this change for a while. That said, I'm a 0 on it. I like the docs dependency group for read the docs.

matthiask:
I think dependency groups is the way to go (except if everybody starts using something like uv projects).

I worry slightly that e.g. the system pip on Ubuntu 22.04 (22.0.2, shipping with Python 3.10) doesn't support dependency groups yet. When using virtualenvs the first thing I do is update pip anyway so it's not a big concern, but I wonder. 

That said, people are probably (and should be) using newer versions of the OS and Python in their development environments for current projects. We don't support outdated versions of Python and Django, and Django also does not. So we're probably fine moving ahead with this.

Re. readthedocs, I really can't say. I've mostly been using old techniques as well when I had to install requirements. We will see what happens.

.readthedocs.yaml
tim-schilling: I'm going to change this based on this part of the docs: https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-commands

tim-schilling:
Rebased on main. Will squash these commits down, then wrangle with ReadTheDocs.

tim-schilling:
Thank you @p-r-a-v-i-n!

PR #2283, opened 2025-12-16
Remove comment for clarity in `DebugToolbarMiddleware` call method

### LINKED ISSUE #2281: Superfluous comment inside `DebugToolbarMiddleware.__call__`
The `DebugToolbarMiddleware.__call__` method contains seemingly superfluous comments that might be simplified - tough may be trivial - for better readability.

https://github.com/django-commons/django-debug-toolbar/blob/0136ff960887edbba4420e235b3ac3ae9dd08a68/debug_toolbar/middleware.py#L128-L133

Considering that `DebugToolbarMiddleware.__acall__` (the async one just beneath) also includes the same comment:

https://github.com/django-commons/django-debug-toolbar/blob/0136ff960887edbba4420e235b3ac3ae9dd08a68/debug_toolbar/middleware.py#L153-L155

A change to something like this might be more concise:

```python
    def __call__(self, request: HttpRequest) -> HttpResponse:
        # Decide whether the toolbar is active for this request.
        if self.async_mode:
            return self.__acall__(request)
        show_toolbar = get_show_toolbar(async_mode=self.async_mode)
```

I'm keen to creating a PR, but I would appreciate what others think before trying to revert something that may be intended in the first place for some reason.

rnazali:
#### Description

Based on how the discussion in the issue unfolds, I decided to propose a trivial change to remove single line of comment inside `DebugToolbarMiddleware.__call__`.

I've also added a space on the branching between `__acall__` and `get_show_toolbar` call.

Fixes #2281 

#### Checklist:

_(I don't think this change need tests nor an extra item in the Pending listing)_

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -126,9 +126,9 @@ class DebugToolbarMiddleware:
             markcoroutinefunction(self)
 
     def __call__(self, request: HttpRequest) -> HttpResponse:
-        # Decide whether the toolbar is active for this request.
         if self.async_mode:
             return self.__acall__(request)
+
         # Decide whether the toolbar is active for this request.
         show_toolbar = get_show_toolbar(async_mode=self.async_mode)
 


matthiask:
Thank you!

PR #2286, opened 2025-12-17
Download prof files

### LINKED ISSUE #1798: Download profiling results

JohananOppongAmoateng:
#### Description

This pr is based on the work of @andoriyaprashant  which he made to fix #1798. I am continuing his work to get it over the finish line. His pr is this #2146 .

Fixes #1798 

#### Checklist:


JohananOppongAmoateng:
Hello everyone sorry for not finishing this. I have been a bit caught up working on other OSS projects . I promise to push new commits this weekend to fix the reviews co pilot left

debug_toolbar/panels/profiling.py
```diff
@@ -183,8 +188,25 @@ def generate_stats(self, request, response):
         self.stats = Stats(self.profiler)
         self.stats.calc_callees()
 
-        root_func = cProfile.label(super().process_request.__code__)
-
+        if (
+            profile_root := dt_settings.get_config()["PROFILER_PROFILE_ROOT"]
+        ) and os.path.exists(profile_root):
+            filename = f"{uuid.uuid4().hex}.prof"
```
tim-schilling: Would it make sense to use the request id here instead? And maybe something that's a little clearer that it's from the toolbar. for example: `f"djdt_profile_{toolbar.request_id}"`

debug_toolbar/panels/profiling.py
```diff
@@ -183,8 +188,25 @@ def generate_stats(self, request, response):
...
+            prof_file_path = os.path.join(profile_root, filename)
+            try:
+                self.profiler.dump_stats(prof_file_path)
+                self.prof_file_path = signing.dumps(filename)
+            except OSError:
+                logger.error(
```
tim-schilling: `logger.exception()` will do the same as `logger.error(exc_info=True)`

debug_toolbar/panels/profiling.py
```diff
@@ -183,8 +188,25 @@ def generate_stats(self, request, response):
...
+                    "Failed to dump profiling stats to %s",
+                    prof_file_path,
+                    exc_info=True,
+                )
+                # If writing to the file fails, we don't want to break the
+                # whole page.
+
+        code = super().process_request.__code__
+        root_func = (code.co_filename, code.co_firstlineno, code.co_name)
```
tim-schilling: Is there a reason why we're not using `cProfile.label()` anymore?

debug_toolbar/panels/sql/tracking.py
```diff
@@ -146,7 +146,20 @@ def _last_executed_query(self, sql, params):
         # process during the .last_executed_query() call.
         self.db._djdt_logger = None
         try:
-            return self.db.ops.last_executed_query(self.cursor, sql, params)
+            # Handle executemany: take the first set of parameters for formatting
```
tim-schilling: Are these changes related to profiling files? If not, they belong in a different PR. If they are, can you provide a little context please?

debug_toolbar/panels/profiling.py
```diff
@@ -183,8 +188,25 @@ def generate_stats(self, request, response):
...
 
-        root_func = cProfile.label(super().process_request.__code__)
-
+        if (
+            profile_root := dt_settings.get_config()["PROFILER_PROFILE_ROOT"]
+        ) and os.path.exists(profile_root):
+            filename = f"{uuid.uuid4().hex}.prof"
+            prof_file_path = os.path.join(profile_root, filename)
+            try:
+                self.profiler.dump_stats(prof_file_path)
```
tim-schilling: Would we be better off using Django's storages instead here?

debug_toolbar/views.py
```diff
@@ -28,3 +33,30 @@ def render_panel(request: HttpRequest) -> JsonResponse:
...
+    if not (root := dt_settings.get_config()["PROFILER_PROFILE_ROOT"]):
+        raise Http404()
+
+    if not (file_path := request.GET.get("path")):
+        raise Http404()
+
+    try:
+        filename = signing.loads(file_path)
+    except signing.BadSignature:
+        raise Http404() from None
```
tim-schilling: Why do we need to sign these?

debug_toolbar/views.py
```diff
@@ -28,3 +33,30 @@ def render_panel(request: HttpRequest) -> JsonResponse:
         content = panel.content
         scripts = panel.scripts
     return JsonResponse({"content": content, "scripts": scripts})
+
+
+@require_GET
+@login_not_required
+def download_prof_file(request):
+    if not (root := dt_settings.get_config()["PROFILER_PROFILE_ROOT"]):
+        raise Http404()
```
tim-schilling: I think there would be value in allowing a person to download a profile file even after they have disabled the setting. I don't think there's a downside to it. What do you think?

debug_toolbar/templates/debug_toolbar/panels/profiling.html
```diff
@@ -1,4 +1,13 @@
 {% load i18n %}
+
+{% if prof_file_path %}
+<div class="djdt-profiling-control">
+  <a href="{% url 'djdt:debug_toolbar_download_prof_file' %}?path={{ prof_file_path|urlencode }}" class="djDebugButton">
```
tim-schilling: Since there's a chance that this could result in a 404, we should use `target="_blank"` to force it to open in a separate window/tab. This prevents the current window from being shown an error state and having to go back (then re-click through the toolbar options to get back to the previous state).

docs/configuration.rst
```diff
@@ -351,6 +351,17 @@ Panel options
...
+* ``PROFILER_PROFILE_ROOT``
+
+  Default: ``None``
+
+  Panel: profiling
+
+  This setting controls the directory where profile files are saved. If set
+  to ``None`` (the default), the profile file is not saved and the download
+  link is not shown. This directory must exist and be writable by the
+  web server process.
```
tim-schilling: Could we remove the "must exist" part of this requirement? It's a better developer experience if they can just set this setting and everything just works.

tim-schilling (changes requested):
Thank you for continuing to work on this 🚀. I've got some questions about the direction of this. If you're confused, or think I may be wrong, please let me know. I'm happy to hop on a call and chat about this too!

debug_toolbar/panels/profiling.py
JohananOppongAmoateng: Okay i will correct it

debug_toolbar/panels/profiling.py
JohananOppongAmoateng: Oh yeah i think we could use the request id

debug_toolbar/templates/debug_toolbar/panels/profiling.html
JohananOppongAmoateng: ok

docs/configuration.rst
JohananOppongAmoateng: Yes i think that we can create the directory for them if it doesnt exist. It should be possible

debug_toolbar/views.py
JohananOppongAmoateng: I am signing them to ensure that users dont just request any file on the filesystem. Without signing, anyone could guess or manipulate the path query parameter (e.g. ../secret.txt or an absolute path) and attempt to download files outside the configured PROFILER_PROFILE_ROOT. But i am also thinking since the toolbar is only used in dev environments maybe we dont need that extra level of caution. What do you think?

tim-schilling: Ahh, that makes sense. Though is a good reason for us to use Django's build `storages` too.

debug_toolbar/views.py
JohananOppongAmoateng: On second thought i agree with you they should be allowed to download them

debug_toolbar/panels/profiling.py
JohananOppongAmoateng: i was using it initially but the tests were breaking for py 3.11 and 3.12 for some unknown reason. I will try reversing it and rerunning the tests to see

debug_toolbar/panels/sql/tracking.py
JohananOppongAmoateng: Oops it was part of the code @andoriyaprashant pushed i was supposed to remove them but was an oversight

debug_toolbar/panels/profiling.py
JohananOppongAmoateng: switching to storages

debug_toolbar/views.py
JohananOppongAmoateng: > I think there would be value in allowing a person to download a profile file even after they have disabled the setting. I don't think there's a downside to it. What do you think?

Wait but we storing the files in that profiler profile root so if they are not set should we store it in media folder or so if it not set we can just check the media folder ot see if it exists

JohananOppongAmoateng: I am thinking if that profile root is none we should store the files in media folder and check media folder when we try downloading them

debug_toolbar/panels/profiling.py
```diff
@@ -183,8 +189,27 @@ def generate_stats(self, request, response):
         self.stats = Stats(self.profiler)
         self.stats.calc_callees()
 
-        root_func = cProfile.label(super().process_request.__code__)
+        profile_root = dt_settings.get_config()["PROFILER_PROFILE_ROOT"]
+        storage = FileSystemStorage(location=profile_root)
```
tim-schilling: Could we use [`django.core.files.storage.default_storage`](https://docs.djangoproject.com/en/6.0/ref/files/storage/#django.core.files.storage.default_storage) instead of instantiating a new one?

tim-schilling: Are we going in the wrong direction here trying to write actual files to the server? Couldn't we just store the file's content with the panel, then render that out like a file in `download_prof_file`?

Sorry to shift our direction again, but something isn't sitting right with me. What do you think? I'd love to know if you feel the same.

JohananOppongAmoateng: I have a question i assume by storing the files content in the panel this would be ephemeral right? If we dont need persistence we could do that

tim-schilling: Yes, they would be. Unless of course they are using the DatabaseStore.

matthiask: The database store also prunes old records. We have no guarantees regarding the contents of debug logs for now, for example across version upgrades. At least that's my understanding

tim-schilling: I've changed the approach of this to utilize the existing `FunctionCall` instances removing this concern.

debug_toolbar/panels/profiling.py
```diff
@@ -131,22 +143,69 @@ def cumtime_per_call(self):
...
             "cumtime": self.cumtime(),
             "cumtime_per_call": self.cumtime_per_call(),
             "tottime": self.tottime(),
             "tottime_per_call": self.tottime_per_call(),
             "count": self.count(),
+            "primitive_count": self.primitive_count(),
         }
 
 
+def _print_stats_from_function_calls(func_list: list[FunctionCall]) -> str:
```
tim-schilling, line 178: I'm not sure about this. I could use a second opinion on whether we want to go this approach or not.

tim-schilling (approved):
So I'm in favor of this, but it's just enough to get it merged. I would appreciate another opinion on this approach. I worry it's overly brittle. I suspect we'll end up redoing a lot of this as a part of integrating Yappi and fixing the profiling panel in general.

I had help from Claude to generate the `Stats.print_stats` reproduction.

debug_toolbar/panels/profiling.py
codingjoe: `pstats.Stats` can be printed to any stream-like object. So you can get a `io.StringIO` and simply call `print_stats`. I don't believe the is a need to recreate the funcitonality, right?

debug_toolbar/templates/debug_toolbar/panels/profiling.html
```diff
@@ -1,4 +1,10 @@
 {% load i18n %}
+
+<form method="get" action="{% url 'djdt:profiling_download' %}" target="_blank">
+    {{ profiling_download_form.as_div }}
+    <button>Download</button>
+</form>
```
codingjoe, line 6: Why use a form and a button for a GET request? Why not use an a-tag with a `download` attribute.

debug_toolbar/urls.py
```diff
@@ -2,4 +2,5 @@
 from debug_toolbar.toolbar import DebugToolbar
 
 app_name = APP_NAME
+
```
codingjoe: Unrealted change.

debug_toolbar/utils.py
```diff
@@ -415,3 +418,34 @@ def get_csp_nonce(request) -> str | None:
         return csp_nonce
     # Django's built-in CSP support uses get_nonce(request)
     return compat.get_nonce(request)
+
+
+@dataclass
+class ViewPathMetadata:
```
codingjoe: Dataclasses are great class factories but not great datastructure primitives. NamedTuples are based on C-structs and much more suitable for immutable data structures.

```suggestion
class ViewPathMetadata(typing.NamedTuple):
```

debug_toolbar/utils.py
```diff
@@ -415,3 +418,34 @@ def get_csp_nonce(request) -> str | None:
...
+    view_func: str
+    view_args: tuple[Any]
+    view_kwargs: dict[str, Any]
+    url_name: str
+
+
+def get_view_path_metadata(request: HttpRequest) -> ViewPathMetadata:
```
codingjoe: This could be a `from_request` factory class method on the actual `ViewPathMetadata` class.

debug_toolbar/utils.py
```diff
@@ -415,3 +418,34 @@ def get_csp_nonce(request) -> str | None:
...
+
+    if getattr(match, "url_name", False):
+        url_name = match.url_name
+        if match.namespaces:
+            url_name = ":".join([*match.namespaces, url_name])
+    else:
+        url_name = _("<unavailable>")
+
+    return ViewPathMetadata(
+        view_func=get_name_from_obj(func),
```
codingjoe: If you want to stick with a dataclass, this should probably be part of post_init.

debug_toolbar/panels/profiling.py
```diff
@@ -197,4 +255,65 @@ def generate_stats(self, request, response):
...
+        request_id: str = form.cleaned_data["request_id"]
+        toolbar: DebugToolbar | None = DebugToolbar.fetch(request_id)
+        if toolbar is None:
+            return HttpResponseBadRequest("Request is no longer available.")
+        panel = toolbar.get_panel_by_id(ProfilingPanel.panel_id)
+        panel_stats = panel.get_stats()
+        if "func_list" not in panel_stats:
+            return HttpResponseBadRequest("No profiling data exists for this request.")
+        content = _print_stats_from_function_calls(panel_stats["func_list"])
+        response = HttpResponse(content, content_type="text/plain")
```
codingjoe: The `Content-Disposition` header would enable you to propose a filename to the browser.

debug_toolbar/panels/profiling.py
```diff
@@ -131,22 +143,69 @@ def cumtime_per_call(self):
...
 
+def _print_stats_from_function_calls(func_list: list[FunctionCall]) -> str:
+    """
+    Create a string that is matches what pstats.Stats.print_stats would
+    print out.
+    """
+    # Column layout matches pstats output: ncalls right-justified in 9
+    # chars, each time value formatted as f8 ("%8.3f"). The 3-space indent
+    # on the header and ordering lines is part of the pstats format.
+    # https://docs.python.org/3/library/profile.html#pstats.Stats.print_stats
```
codingjoe, line 186: I don't want to insinuate anything, but the comments read like AI code. Maybe we don't fan the fire and limit comments to only what can't be easily understood from the code itself.

codingjoe (changes requested):
Hi there,

The second opinion arrived; I hope you don't mind.

I like the idea. However, the current implementation adds a pretty high maintenance burden onto this package. If you can make this work with a buffer object, like I described below, the scales might tip for me.

Cheers!
Joe

debug_toolbar/templates/debug_toolbar/panels/profiling.html
tim-schilling: I did it for consistency with the history panel which was originally done because the form used to inherit from SignedDataForm. I do like that it enforces the data validation more strictly.

debug_toolbar/utils.py
tim-schilling: Good call

debug_toolbar/panels/profiling.py
tim-schilling: The challenge is the panel doesn't show the same content.

debug_toolbar/panels/profiling.py
tim-schilling: I called out that I used Claude in the PR.

I think in this case the comments help. There's syntax used in this function that isn't typical in my opinion, so having it explained helps.

debug_toolbar/panels/profiling.py
tim-schilling: The profiling panel as a whole needs a bit of a rework. The threshold concept just doesn't seem to work properly. It seems like we should lean more into the actual APIs of the standard library.

Getting this all resolved is on my todo list, but I'd really like to avoid it for this feature.

debug_toolbar/panels/profiling.py
codingjoe: Yes, sorry. I only read that comment later.

Hm… depends. I think tests for "how this should look" actually help me personally more than the comments. But better too much documentation that too little 👍 

At least my spidysense for AI still works. 🤠

debug_toolbar/templates/debug_toolbar/panels/profiling.html
codingjoe: Which currently is only an ID, right? I don't really care much, but without knowing the history, I just looked a little odd.

tim-schilling: Yup, yup. That's correct.

codingjoe (approved):
Code looks good now. I have my reservations about the complexity/helpfulness ratio, but I am also not a heavy user for the profiling with the debug toolbar.

tim-schilling:
I'm sorry @JohananOppongAmoateng. I led us in the wrong direction at some point. The point of this change wasn't to generate a file with the printed stats, but the `.prof` binary file. This is what other tools utilize and what the issue was about.

That said, I'm not sure on the best way to implement this feature. Trying to generate the `.prof` file on the fly seems impossible since it's the binary we need. So we'd need to store it somewhere. Putting it in the stats is one possibility, but it'll need to be serialized down to JSON. The other option is to write it to a file immediately using the storages API, then maintain a reference to that file in stats.

If we're to attempt one of those, I think we may want to do it in (yet another) PR because the discussion and changes in this one are in the wrong direction.

PR #2287, opened 2025-12-20
Optimize CI with concurrency

p-r-a-v-i-n:
I added a concurrency group to the `test.yml` workflow to make it more efficient.

-It ensures that only the latest commit on a branch is running. 
-If a new push is made while an old one is still going, the old run will automatically cancel. 
-This helps us save GitHub Actions minutes and keeps our PRs clean.


matthiask:
Thanks!

PR #2291, opened 2026-01-09
Add graceful degradation for large SQL queries

### LINKED ISSUE #2293: SQL panel freezes or crashes when queries contain large IN clauses (e.g., thousands of UUIDs)
### Description

When SQL queries contain large `IN` clauses with thousands of parameters (e.g., UUIDs), the SQL panel either:
- **Freezes for 10-18+ seconds** (sqlparse < 0.5.5)
- **Crashes with SQLParseError** (sqlparse >= 0.5.5)

This is a common pattern when fetching IDs from one query and filtering by them in another:

```python
job_ids = list(Campaign.objects.filter(...).values_list('job_id', flat=True))
# job_ids contains 5000+ UUIDs

applications = JobApplication.objects.filter(job_id__in=job_ids)
# Database query is fast, but debug toolbar freezes/crashes
```

### Behavior Matrix

| debug-toolbar | sqlparse | When | Behavior |
|---------------|----------|------|----------|
| <= 5.x | < 0.5.5 | Page load | Freezes 10-18+ seconds |
| <= 5.x | >= 0.5.5 | Page load | SQLParseError crash |
| >= 6.x | < 0.5.5 | SQL panel click | Freezes 10-18+ seconds |
| >= 6.x | >= 0.5.5 | SQL panel click | SQLParseError crash |

### Benchmark Results (sqlparse 0.5.3)

| IN clause size | Format time |
|----------------|-------------|
| 100 UUIDs | 0.01s |
| 500 UUIDs | 0.08s |
| 1,000 UUIDs | 0.26s |
| 3,000 UUIDs | 1.81s |
| 5,000 UUIDs | 4.87s |
| 10,000 UUIDs | **18.02s** |

### Root Cause

The bottleneck is in `debug_toolbar/panels/sql/utils.py`:

```python
@lru_cache(maxsize=128)
def parse_sql(sql, *, simplify=False):
    stack = get_filter_stack(simplify=simplify)
    return "".join(stack.run(sql))  # sqlparse tokenizes entire SQL string
```

For a query with 5000 UUIDs:
- SQL string length: ~170,000 characters
- Token count: ~15,000+ tokens
- Each token processed by filters (indent, bold keywords, HTML escape)

The LRU cache doesn't help because each query with different parameters is a cache miss.

### Steps to Reproduce

A minimal reproduction project is available:
https://github.com/kkm-horikawa/debug-toolbar-perf-issue

```bash
git clone https://github.com/kkm-horikawa/debug-toolbar-perf-issue.git
cd debug-toolbar-perf-issue
uv sync
uv run python manage.py migrate
uv run python manage.py runserver
# Visit http://127.0.0.1:8000/ and click on test links
```

### Current Workaround

```python
DEBUG_TOOLBAR_CONFIG = {
    'PRETTIFY_SQL': False,  # Disables ALL SQL formatting
}
```

This is suboptimal because it disables formatting for all queries, not just problematic ones.

### Proposed Solution

I have submitted PR #2291 that:

1. **Adds `SQL_PRETTIFY_MAX_LENGTH` setting** (default 50000) to skip formatting for long queries
2. **Catches SQLParseError** from sqlparse >= 0.5.5 gracefully
3. **Shows informative message** with SQL preview when formatting is skipped

This provides:
- Normal formatting for typical queries
- Graceful degradation for extreme cases
- Configurable threshold

### Related Issues

- #1402 (closed as stale - this issue provides the requested reproduction steps)

### Environment

- Python 3.11+
- Django 5.0+
- django-debug-toolbar 5.2.0 / 6.1.0
- sqlparse 0.5.3 / 0.5.5+

kkm-horikawa:
#### Description

Add graceful degradation for SQL queries that exceed sqlparse's token limits, preventing crashes in the SQL panel.

**Problem:**
When SQL queries contain thousands of parameters (e.g., UUIDs in IN clauses), the SQL panel crashes with `SQLParseError` (sqlparse >= 0.5.5).

**Solution:**
Catch `SQLParseError` and retry formatting with grouping disabled. This approach:
- Handles the crash automatically without requiring configuration
- Eliminates the need for a new setting like `SQL_PRETTIFY_MAX_LENGTH`

**Out of scope:** The freezing behavior with older sqlparse versions (< 0.5.5) is a pre-existing issue and not addressed in this PR.

**Before (debug-toolbar 5.2 + sqlparse 0.5.3) - page freezes (out of scope):**

before-page-freeze

**Before (debug-toolbar 6.1 + sqlparse 0.5.5) - SQLParseError:**

before-sqlparse-error

**After (this PR):**

after-fix

**Reproduction project:** https://github.com/kkm-horikawa/debug-toolbar-perf-issue

Fixes #2293

#### Checklist:

### DIFF
--- a/debug_toolbar/panels/sql/utils.py
+++ b/debug_toolbar/panels/sql/utils.py
@@ -9,6 +9,13 @@ from sqlparse import tokens as T
 
 from debug_toolbar import settings as dt_settings
 
+# Import SQLParseError for graceful exception handling
+try:
+    from sqlparse.exceptions import SQLParseError
+except ImportError:
+    # Older sqlparse versions don't have this exception
+    SQLParseError = None
+
 
 class ElideSelectListsFilter:
     """sqlparse filter to elide the select list from top-level SELECT ... FROM clauses,
@@ -92,11 +99,59 @@ def is_select_query(sql):
     return sql.lower().lstrip(" (").startswith("select")
 
 
+def _format_skipped_sql(sql, reason):
+    """
+    Format SQL that was skipped from prettification.
+
+    Shows a notice and the first portion of raw SQL for debugging.
+    """
+    preview_length = 500
+    preview = escape(sql[:preview_length])
+    total_length = len(sql)
+
+    if total_length > preview_length:
+        suffix = f"\n\n... ({total_length - preview_length:,} more characters)"
+    else:
+        suffix = ""
+
+    return (
+        f'<span class="djdt-sql-skipped">'
+        f"<em>SQL formatting skipped ({reason})</em>"
+        f"</span>"
+        f"<pre>{escape(preview)}{suffix}</pre>"
+    )
+
+
 def reformat_sql(sql, *, with_toggle=False):
-    formatted = parse_sql(sql)
-    if not with_toggle:
-        return formatted
-    simplified = parse_sql(sql, simplify=True)
+    # Check length threshold to prevent slow formatting of large queries
+    max_length = dt_settings.get_config()["SQL_PRETTIFY_MAX_LENGTH"]
+    if max_length and len(sql) > max_length:
+        skipped = _format_skipped_sql(
+            sql,
+            f"query length {len(sql):,} exceeds threshold {max_length:,}",
+        )
+        if with_toggle:
+            # Return as non-collapsible since both versions would be the same
+            return f'<span class="djDebugUncollapsed">{skipped}</span>'
+        return skipped
+
+    try:
+        formatted = parse_sql(sql)
+        if not with_toggle:
+            return formatted
+        simplified = parse_sql(sql, simplify=True)
+    except Exception as e:
+        # Handle sqlparse exceptions (e.g., MAX_GROUPING_TOKENS exceeded in >= 0.5.5)
+        if SQLParseError is not None and isinstance(e, SQLParseError):
+            reason = f"sqlparse error: {e}"
+        else:
+            # Re-raise unexpected exceptions
+            raise
+        skipped = _format_skipped_sql(sql, reason)
+        if with_toggle:
+            return f'<span class="djDebugUncollapsed">{skipped}</span>'
+        return skipped
+
     uncollapsed = f'<span class="djDebugUncollapsed">{simplified}</span>'
     collapsed = f'<span class="djDebugCollapsed djdt-hidden">{formatted}</span>'
     return collapsed + uncollapsed
--- a/debug_toolbar/panels/sql/utils.py
+++ b/debug_toolbar/panels/sql/utils.py
@@ -6,6 +6,7 @@ import sqlparse
 from django.dispatch import receiver
 from django.test.signals import setting_changed
 from sqlparse import tokens as T
+from sqlparse.exceptions import SQLParseError
 
 from debug_toolbar import settings as dt_settings
 
@@ -92,42 +93,7 @@ def is_select_query(sql):
     return sql.lower().lstrip(" (").startswith("select")
 
 
-def _format_skipped_sql(sql, reason):
-    """
-    Format SQL that was skipped from prettification.
-
-    Shows a notice and the first portion of raw SQL for debugging.
-    """
-    preview_length = 500
-    preview = escape(sql[:preview_length])
-    total_length = len(sql)
-
-    if total_length > preview_length:
-        suffix = f"\n\n... ({total_length - preview_length:,} more characters)"
-    else:
-        suffix = ""
-
-    return (
-        f'<span class="djdt-sql-skipped">'
-        f"<em>SQL formatting skipped ({reason})</em>"
-        f"</span>"
-        f"<pre>{preview}{suffix}</pre>"
-    )
-
-
 def reformat_sql(sql, *, with_toggle=False):
-    # Check length threshold to prevent slow formatting of large queries
-    max_length = dt_settings.get_config()["SQL_PRETTIFY_MAX_LENGTH"]
-    if max_length and len(sql) > max_length:
-        skipped = _format_skipped_sql(
-            sql,
-            f"query length {len(sql):,} exceeds threshold {max_length:,}",
-        )
-        if with_toggle:
-            # Return as non-collapsible since both versions would be the same
-            return f'<span class="djDebugUncollapsed">{skipped}</span>'
-        return skipped
-
     formatted = parse_sql(sql)
     if not with_toggle:
         return formatted
@@ -141,7 +107,15 @@ def reformat_sql(sql, *, with_toggle=False):
 @lru_cache(maxsize=128)
 def parse_sql(sql, *, simplify=False):
     stack = get_filter_stack(simplify=simplify)
-    return "".join(stack.run(sql))
+    try:
+        return "".join(stack.run(sql))
+    except SQLParseError:
+        # The query either exceeds the number of tokens or depth of tokens.
+        # Recreate the FilterStack and explicitly disable the grouping to avoid
+        # those errors.
+        stack = get_filter_stack(simplify=simplify)
+        stack._grouping = False
+        return "".join(stack.run(sql))
 
 
 @cache
--- a/debug_toolbar/panels/sql/utils.py
+++ b/debug_toolbar/panels/sql/utils.py
@@ -9,13 +9,6 @@ from sqlparse import tokens as T
 
 from debug_toolbar import settings as dt_settings
 
-# Import SQLParseError for graceful exception handling
-try:
-    from sqlparse.exceptions import SQLParseError
-except ImportError:
-    # Older sqlparse versions don't have this exception
-    SQLParseError = None
-
 
 class ElideSelectListsFilter:
     """sqlparse filter to elide the select list from top-level SELECT ... FROM clauses,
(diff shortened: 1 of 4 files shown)


tim-schilling:
I'm unsure we should be adding this much code for this exception flow. Is sqlparse working on a solution for the underlying issue? This may be better suited as an issue with a workaround people can implement on their own.

tim-schilling:
Let's split the error handling of `SQLParseError` out of this PR, into its own.

kkm-horikawa:
Thanks for the feedback! I've removed the `SQLParseError` handling from this PR.

This PR now focuses solely on the `SQL_PRETTIFY_MAX_LENGTH` setting (proactive length check).

I'll create a separate PR for `SQLParseError` handling after this one is merged. Does this approach work for you?

debug_toolbar/panels/sql/utils.py
```diff
@@ -91,11 +91,47 @@ def is_select_query(sql):
...
+        f"<em>SQL formatting skipped ({reason})</em>"
+        f"</span>"
+        f"<pre>{preview}{suffix}</pre>"
+    )
+
+
 def reformat_sql(sql, *, with_toggle=False):
+    # Check length threshold to prevent slow formatting of large queries
+    max_length = dt_settings.get_config()["SQL_PRETTIFY_MAX_LENGTH"]
+    if max_length and len(sql) > max_length:
```
tim-schilling: I feel like this should be integrated into the sqlparse logic in `get_filter_stack()`. Have you looked into that at all to see if we can do something with `FilterStack`?

tests/panels/test_sql.py
```diff
@@ -869,6 +869,50 @@ def test_explain_with_union(self):
...
         self.assertTrue(query["is_select"])
 
+    @override_settings(
+        DEBUG_TOOLBAR_CONFIG={"SQL_PRETTIFY_MAX_LENGTH": 100, "PRETTIFY_SQL": True}
+    )
+    def test_sql_prettify_max_length(self):
+        """
+        Test that SQL formatting is skipped for queries exceeding the max length threshold.
+        """
+        from debug_toolbar.panels.sql.utils import reformat_sql
```
tim-schilling: Please avoid importing things within functions.

debug_toolbar/panels/sql/utils.py
```diff
@@ -91,11 +91,47 @@ def is_select_query(sql):
...
+        skipped = _format_skipped_sql(
+            sql,
+            f"query length {len(sql):,} exceeds threshold {max_length:,}",
```
tim-schilling: We may want to convert this all to a template and use `render_to_string()` so that it's easier to use the [`intcomma` template filter](https://docs.djangoproject.com/en/6.0/ref/contrib/humanize/#intcomma). This would better handle internationalization of using commas versus periods.

debug_toolbar/settings.py
```diff
@@ -50,6 +50,7 @@ def _is_running_tests():
         "django.utils.functional",
     ),
     "PRETTIFY_SQL": True,
+    "SQL_PRETTIFY_MAX_LENGTH": 50000,  # Skip formatting for SQL longer than this
```
tim-schilling:
```suggestion
    "PRETTIFY_SQL_MAX_LENGTH": 10000,  # Skip formatting for SQL longer than this
```
Let's rename this to better match the other setting. Let's also use the default SQLParse max value from https://sqlparse.readthedocs.io/en/latest/api.html#security-and-performance-considerations.

tim-schilling (changes requested):
Thank you for suggesting this. However, I think we have too many utility functions for rendering the SQL at this point. A change will need to review more of the whole process and identify why it's appropriate compared to other options. For example, I think there may be something achievable by modifying `parse_sql` and/or using SQLParse's `FilterStack` API. Let me know what you think please.

kkm-horikawa:
Thank you for the feedback. You're right that adding a separate utility function is unnecessary.

I've refactored to integrate the length check directly into `parse_sql`. I've also renamed the setting to `PRETTIFY_SQL_MAX_LENGTH` with a default of 10000 to match sqlparse's `MAX_GROUPING_TOKENS`.

The current implementation doesn't show a message when formatting is skipped - it just returns the escaped SQL. I made this choice to avoid adding new files or functions for this specific case, keeping the change minimal. However, if you'd prefer to have a notice displayed to users, I can add that using a template with `render_to_string()` and `intcomma` as you suggested. What do you think?

debug_toolbar/panels/sql/utils.py
kkm-horikawa: Done. I've integrated the length check directly into `parse_sql` instead.

tests/panels/test_sql.py
kkm-horikawa: Fixed. Moved the import to the top of the file.

debug_toolbar/settings.py
kkm-horikawa: Done.

debug_toolbar/panels/sql/utils.py
kkm-horikawa: I've removed the message display for now to keep the change minimal. If a notice is desired, I can add it using a template as you suggested - see my comment above for details.

debug_toolbar/panels/sql/utils.py
```diff
@@ -103,6 +103,9 @@ def reformat_sql(sql, *, with_toggle=False):
 
 @lru_cache(maxsize=128)
 def parse_sql(sql, *, simplify=False):
+    max_length = dt_settings.get_config()["PRETTIFY_SQL_MAX_LENGTH"]
+    if max_length and len(sql) > max_length:
```
tim-schilling: I'm not sure this is right. This would limit sql queries to a string length of 10,000 characters. I was under the impression we were trying to improve performance when the query has more than 10,000 tokens. I believe that's not the same thing as a query with a string length of 10,000 characters.

And if we're removing the formatting in this case, it would make sense to avoid calling `stack.enable_grouping()` in that case.

tim-schilling: I believe catching the error explicitly does what you're after:

```python
@lru_cache(maxsize=128)
def parse_sql(sql, *, simplify=False):
    stack = get_filter_stack(simplify=simplify)
    try:
        return "".join(stack.run(sql))
    except SQLParseError:
        # The query either exceeds the number of tokens or depth of tokens.
        # Recreate the FilterStack and explicitly disable the grouping to avoid
        # those errors.
        stack = get_filter_stack(simplify=simplify)
        stack._grouping = False
        return "".join(stack.run(sql))
  ```

Comment (kkm-horikawa) [reply debug_toolbar/panels/sql/utils.py]: Thank you for the review!

  Your suggested approach of catching SQLParseError is much better. It handles the issue only when it actually occurs,
  and eliminates the need for a new configuration setting.

  I'll update the implementation in this direction. Thank you for your patience and continued guidance through multiple
  reviews!

Comment (kkm-horikawa) [reply debug_toolbar/panels/sql/utils.py]: Done. Thank you for the review!

  I've updated the implementation to catch SQLParseError as you suggested. This approach also eliminates the need for
  the PRETTIFY_SQL_MAX_LENGTH setting.

  This PR focuses on handling the SQLParseError crash introduced in sqlparse >= 0.5.5. The freezing behavior with older
  versions (< 0.5.5) is pre-existing and considered out of scope for this PR.

  I've also updated the PR description to reflect these changes.

  Thank you for your patience and continued guidance through multiple reviews!

File: tests/panels/test_sql.py [hunk 6]
```diff
@@ -869,6 +869,48 @@ def test_explain_with_union(self):
...
+    @override_settings(DEBUG_TOOLBAR_CONFIG={"PRETTIFY_SQL": True})
+    def test_sql_parse_error_graceful_degradation(self):
+        """
+        Test that SQLParseError is handled gracefully by disabling grouping.
+        """
+        from unittest.mock import patch
+
+        from sqlparse.exceptions import SQLParseError
+
+        from debug_toolbar.panels.sql.utils import parse_sql
```
tim-schilling: @kkm-horikawa please review the LLM output more thoroughly in the future. For some reason, LLMs love to generate inline import statements. We try to avoid that practice in this project.

tests/panels/test_sql.py
```diff
@@ -869,6 +869,48 @@ def test_explain_with_union(self):
...
+
+        # Clear the cache to ensure our mock is used
+        parse_sql.cache_clear()
+
+        # Mock the filter stack to raise SQLParseError on first call
+        call_count = 0
+        original_run = None
+
+        def mock_run(sql):
```
tim-schilling: Give the [`side_effect` attribute](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect) a look and then the [`call_count` attribute](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_count). We can reuse some standard library tools to avoid rebuilding the wheel 😁

tim-schilling (changes requested):
I'd like to see the test improved a bit more before we merge this. Thank you for continuing to iterate on this!

tests/panels/test_sql.py
tim-schilling: If you get stuck, let me know. I purposefully didn't write out the code, but I'm happy to include that if you're just wanting this PR to wrap up quickly.

tim-schilling:
@kkm-horikawa can you merge your changes in from your other PR when you have time please?

tim-schilling (approved):
Thank you @kkm-horikawa. I merged your other change into this branch. I'm going to squash these commits and merge to main.

Issue #2260, opened 2025-11-20
Update PR template with question about using an LLM?


tim-schilling:
I would like to add a checkbox to our PR template that indicates if the change was at least partially generated with an LLM/AI and points to a LLM/AI policy that communicates our stance.

From my perspective, LLMs are tools that can help. However, they tend to be verbose and generate lots of code. If I'm reviewing that code, I want to be able to respect the changes to the level the human put effort into them. If I know a person spent 10 hours crafting a PR to generate a change, I'll be more accepting of minor differences. However, if it's been generated with an LLM/AI with less effort, I think it's fair to be a little more aggressive on asking for changes. After all, the LLM/AI isn't going to care and will gladly make those changes in a few minutes.

The challenge is, it's hard to know when this is the case or not. I think editing the PR template may help new contributors know that a LLM disclosure is appropriate.

[Edit]: I wrote this as a statement, but it's definitely a question open to discussion.

matthiask:
As a person who grudgingly started using LLMs too I agree. 

I think this is a good post: https://xeiaso.net/notes/2025/assisted-by-footer/ I'm not sure requiring such a footer would be a good idea (because the tooling doesn't really make it easy).

JohananOppongAmoateng:
@tim-schilling i think we can close this

tim-schilling:
Closed in #2294

PR #2294, opened 2026-01-14
Add LLM usage checkbox to PR template

yeongbin05:
#### Description

This PR implements the suggestion discussed in #2260 to add an AI/LLM usage checkbox to the PR template.
This addition helps reviewers understand better the context of the contribution.

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,4 +12,4 @@ Fixes # (issue)
 - [ ] I have added an item to the Pending section of ``docs/changes.rst``.
 
 #### AI/LLM Usage
-- [ ] This PR includes code generated with the help of an AI/LLM 
+- [ ] This PR includes code generated with the help of an AI/LLM


matthiask:
Thank you! I think we could follow Django's lead and implement something like this:
 
https://github.com/django/django/commit/4f580c465928df97407d74798ad72f2bb60a3e1d

I know this isn't exactly the thing which was discussed in the linked issue. We're in an environment which is moving quickly and I have a more nuanced position now where I think it's important that the code is verified, the fact that an LLM was used per se is less interesting than I thought a few months ago. Others may disagree though.

tim-schilling:
@matthiask For me, the concern is more about trying to provide fair feedback. If someone used an AI to generate code, there's probably parts of it that were done that weren't thought about (using a docstring block versus a line for example). It would be easier to review code when I know an LLM was used because there's less attachment to the code. I don't know. This doesn't need to be a permanent thing either, but I know right now I've made a few comments about potentially LLM generated code. Maybe I should be guiding them on how to use it better? Or how to review their own code before submitting it?

The verified bit doesn't really mean much for me. If I don't know the person, I have to verify the code myself regardless of whether they check that box.

matthiask:
> It would be easier to review code when I know an LLM was used because there's less attachment to the code. I don't know. 

Ah yes, that's a good point. I was thinking more about the quality of the result, and not so much about the feelings of the submitter. 

I think the pull request is good as it is then; I think people aren't discouraged by the current wording from contributing to the project, whether they have been using LLMs or not.

PR #2303, opened 2026-01-31
bug: fix  not overrideing Server-Timing if it exists in header already

### LINKED ISSUE #2300: :bug: do not override Server-Timing if it exists
The middleware just overrides the Server-Timing header if it exists. It should simply amend measurements to be non-destructive.

JohananOppongAmoateng:
#### Description

Fix middleware overriding server timing header

Fixes #2300 

#### Checklist:

#### AI/LLM Usage


tim-schilling (approved):
Looks good, thank you!

tim-schilling (changes requested):
Ah, actually @JohananOppongAmoateng we should note this in the change log as per the PR template. I think this is valuable to mention.

JohananOppongAmoateng:
@tim-schilling added it

tim-schilling (changes requested):
Looking at this again, it looks like we're always going to append to the header with this structure. I'm not sure we want to do that in all cases. For example, in `HistoryPanel.get_headers()` it's trying to set the request id. If it's already been set, we wouldn't want to append. We'd want to either leave it alone or overwrite it entirely.

Do you have any ideas on how to make this more reusable?

JohananOppongAmoateng:
the problem is how do you determine what to override or what to leave

tim-schilling:
I can try to come up with something, but it may be a bit. If you'd like to propose something, that may be faster. We can iterate on things too, it doesn't need to be perfect in the first attempt.

tim-schilling:
@JohananOppongAmoateng so I think one way is to have two different ways for a panel to set a header in the response. An update collection (overrides the value) and an append collection (adds to the values). See https://github.com/django-commons/django-debug-toolbar/pull/2303/commits/3d40ff77c75ed1482b91dfba3da974a2c6c8aa28 for the actual code. This was a first attempt, but the tests pass. We probably need to clean it up and rework our documentation to use it, if we want to.

JohananOppongAmoateng:
> @JohananOppongAmoateng so I think one way is to have two different ways for a panel to set a header in the response. An update collection (overrides the value) and an append collection (adds to the values). See [3d40ff7](https://github.com/django-commons/django-debug-toolbar/commit/3d40ff77c75ed1482b91dfba3da974a2c6c8aa28) for the actual code. This was a first attempt, but the tests pass. We probably need to clean it up and rework our documentation to use it, if we want to.

i think that would work

matthiask:
I think the PR shows promise.

I immediately thought of https://jsonpatch.com/ because of the update/append problem, but I don't know if something like that also exists for dicts/headers.

debug_toolbar/panels/__init__.py
```diff
@@ -7,6 +9,17 @@
...
 
+@dataclass
+class PanelHeaders:
+    """
+    A data-structure to allow panels to update headers or append to existing
+    values.
+    """
+
+    append: dict[str, str] = field(default_factory=dict)
+    update: dict[str, str] = field(default_factory=dict)
```
codingjoe, line 20: You can use a named tuple if you actually want to use the type, or a protocol, if you just want to define an interface.

I also left a small suggestion on how to condense the docstring.
```suggestion
class PanelHeaders(typing.NamedTuple):
    """
    Allow panels to update headers or append to existing values.
    """

    append: dict[str, str] = field(default_factory=dict)
    update: dict[str, str] = field(default_factory=dict)
```

debug_toolbar/middleware.py
```diff
@@ -214,12 +216,19 @@ def _postprocess(
...
+        update, append = {}, {}
         for panel in panels:
-            for header, value in panel.get_headers(request).items():
-                if header in headers:
-                    headers[header] += f", {value}"
-                else:
-                    headers[header] = value
-        return headers
+            panel_headers = panel.get_headers(request)
+            update.update(panel_headers.update)
```
codingjoe, line 225:
```suggestion
            update |= panel_headers.update
```

debug_toolbar/middleware.py
```diff
@@ -214,12 +216,19 @@ def _postprocess(
...
+            for key, value in panel_headers.append.items():
+                append.setdefault(key, response.headers.get(key, ""))
+                append[key] += f", {value}"
+        # Only include the append headers when no panel updated the value
+        # entirely.
+        for key, value in append.items():
+            if key not in update:
+                update[key] = value
+        return update
```
codingjoe, line 234: This can be shortened to:
```suggestion
        return append | update
```

codingjoe (commented):
Hm… it's a complex solution to a small problem.

I left a few comments, but since Django (wrongly) treats HTTP headers as a dictionary, a few dict operations would be sufficient, right?

debug_toolbar/panels/__init__.py
tim-schilling: The NamedTuple approach will result in us having to duplicate the default_factory logic. I think we're okay using a dataclass here.

tim-schilling: I'll push up the doc string change.

tim-schilling (approved):
I'm on board for this now. Though I can't help but wonder if `PanelHeaders.update` should be `PanelHeaders.replace` 😮‍💨

PR #2304, opened 2026-01-31
Add `CacheStore`, a store that uses Django's cache framework

robhudson:
Summary

- Adds a new `CacheStore` that uses Django's cache framework to persist toolbar data
- Adds `CACHE_BACKEND` and `CACHE_KEY_PREFIX` settings to configure which cache and key prefix to use
- Toolbar's own cache operations are excluded from the cache panel via an `_UntrackedCache` proxy

Motivation

By using Django's cache framework, users can choose any configured cache backend for toolbar storage. The interesting thing about this is that the `LocMemCache` backend can effectively act like the `MemoryStore`, and the `DatabaseCache` backend can effectively act like the `DatabaseStore`.

Notes

* I haven't tested all the backends (e.g. `FileBasedCache`)
* I haven't tested this on a real project with real database usage patterns

#### Checklist:

#### AI/LLM Usage
- [x] I have reviewed / edited this code thoroughly

### DIFF
--- a/debug_toolbar/store.py
+++ b/debug_toolbar/store.py
@@ -1,9 +1,11 @@
 import contextlib
+import functools
 import json
 from collections import defaultdict, deque
 from collections.abc import Iterable
 from typing import Any
 
+from django.core.cache import caches
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import transaction
 from django.utils.module_loading import import_string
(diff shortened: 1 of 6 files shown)


debug_toolbar/store.py
```diff
@@ -219,5 +220,144 @@ def panels(cls, request_id: str) -> Any:
...
+
+    def __init__(self, cache):
+        self._cache = cache
+
+    def __getattr__(self, name):
+        attr = getattr(self._cache, name)
+        if not callable(attr):
+            return attr
+
+        def untracked(*args, **kwargs):
```
tim-schilling, line 242: Should we be using `functools.wraps()` here?

debug_toolbar/store.py
```diff
@@ -219,5 +220,144 @@ def panels(cls, request_id: str) -> Any:
...
+    @classmethod
+    def clear(cls):
+        """Remove all requests from the request store."""
+        cache = cls._get_cache()
+        ids_key = cls._request_ids_key()
+        request_ids = cache.get(ids_key, [])
+
+        # Delete all request data
+        for request_id in request_ids:
+            cache.delete(cls._request_key(request_id))
```
tim-schilling: Did you consider using [`delete_many()`](https://docs.djangoproject.com/en/6.0/topics/cache/#django.core.cache.cache.delete_many)?

debug_toolbar/store.py
```diff
@@ -219,5 +220,144 @@ def panels(cls, request_id: str) -> Any:
...
+
+    @classmethod
+    def _request_key(cls, request_id: str) -> str:
+        """Return the cache key for a specific request's data."""
+        return f"{cls._key_prefix()}req:{request_id}"
+
+    @classmethod
+    def request_ids(cls) -> Iterable:
+        """The stored request ids."""
+        return cls._get_cache().get(cls._request_ids_key(), [])
```
tim-schilling, line 311: Should this read into a deque so that we get [`O(1)` on popleft rather than O(N) with `pop(0)` of a list](https://stackoverflow.com/a/32543932)?

tim-schilling (commented):
Thank you for adding this. We'll need to add some documentation around this as well. I suspect you're just looking for feedback on the approach right now though. I have a few questions.

For me, I think I would have gone with a `ContextVar` to control the tracking of the cache panel similar to `allow_sql`, but this should be fine.

tests/test_store.py
```diff
@@ -251,3 +251,100 @@ def test_cleanup_old_entries(self):
...
+
+@override_settings(
+    DEBUG_TOOLBAR_CONFIG={
+        "CACHES": {
+            "default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
+        },
+        "TOOLBAR_STORE_CLASS": "debug_toolbar.store.CacheStore",
+    }
+)
+class CacheStoreTestCase(TestCase):
```
tim-schilling: Maybe I'm missing it, but do we have a test that confirms that using the cache store doesn't cause the records to show up in the CachePanel's data?

tim-schilling:
Another thought, I wonder if we can run the full test suite using `DatabaseStore` and `CacheStore` now? We could expose that as part of the tox.ini or something similar. Let me know what you think.

robhudson:
Thanks for the review!

A couple notes:
- I did try with a context var for not recording internal cache operations but it caused a lot of boilerplate code. And you have to remember to use that boilerplate code. So I pivoted to the `_UntrackedCache` to make the code simpler.
- I had the thought of tweaking the test suite also but thought I'd just start off with the basic idea. I'd also be curious about testing with the various cache backends also.

I'll look into your other comments today. Good stuff.

robhudson:
An interesting discovery:

The `_UntrackedCache` wrapper only prevents CachePanel tracking by setting `cache._djdt_panel = None`. It doesn't prevent SQL tracking. When `DatabaseCache` executes SQL queries (to tables like `django_cache`), those queries aren't filtered by `SKIP_TOOLBAR_QUERIES` because they're not in `DDT_MODELS` (which only contains `debug_toolbar_historyentry`).

Is this acceptable behavior, or should we also prevent these SQL queries from being tracked?

It should be possible to dynamically add the `DatabaseCache` table name to `DDT_MODELS`, but only for the cache backend configured in the toolbar's `CACHE_BACKEND`. This way:
- The toolbar's own cache storage queries are filtered (similar to `DatabaseStore`)
- Other `DatabaseCache` backends used by the application remain visible

tim-schilling:
> Is this acceptable behavior, or should we also prevent these SQL queries from being tracked?

It seems like we should have the ability to disable all tracking from within the toolbar temporarily, but ideally also allow the user to view those events. I suspect this means we need to shift away from `SKIP_TOOLBAR_QUERIES` to a more generic setting.

However, reducing our scope to just this change, I think ideally we'd prevent these SQL queries from being tracked. If it's not going to result in something horrible, I think we should do that.

robhudson:
I refactored the tests a bit to test the various cache backends and added support for not tracking the self-generated SQL from the cache store using a database.

tests/test_store.py
```diff
@@ -50,59 +54,102 @@ def test_methods_are_not_implemented(self):
             store.BaseStore.panel("", "")
 
 
-class MemoryStoreTestCase(TestCase):
-    @classmethod
-    def setUpTestData(cls) -> None:
-        cls.store = store.MemoryStore
+class CommonStoreTestsMixin:
```
tim-schilling, line 57: I love this, thank you.

robhudson:
Sorry for the delay. I finally got around to adding some docs and updated the changes file. Let me know if there's anything else you'd like to see here. :)

PR #2314, opened 2026-02-19
Add a `py.typed` file, to make types available downstream

brianhelba:
#### Description
This will not change any behavior of this package, but it is required for consuming projects to read any existing type information here.

This file is required to exist and should be empty. There's no alternative mechanism. See: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

This follows #2227.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,7 @@ Pending
 * Standardize use of time/duration units and labels across panels.
 * Added translations for Lithuanian, Turkish and Uzbek.
 * Update the translations.
+* Expose a ``py.typed`` marker file.
 
 6.2.0 (2026-01-20)
 ------------------


tim-schilling:
If we don't have everything using type annotations, is that going to cause problems for someone using mypy?

brianhelba:
> If we don't have everything using type annotations, is that going to cause problems for someone using mypy?

No, not at all, it'll be the same as someone with local code / functions that aren't fully typed. In short, MyPy will infer untyped things as [`typing.Any`](https://docs.python.org/3/library/typing.html#typing.Any). Indeed, this is _already_ how MyPy downstream is perceiving everything in `django-debug-toolbar`, since we lack `py.typed`. The addition of this file will be strictly additive, allowing types to be used from this package where they do exist.

Personally, I think it's a nuisance that [PEP 561](https://peps.python.org/pep-0561/) makes us include this, but it's just [become a routine practice](https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/).

Also, note that `hatchling.build` will include `py.typed` in sdists without any additional configuration.

tim-schilling:
Thank you for that explanation @brianhelba! I'll merge this once the docs/linter passes.

PR #2321, opened 2026-03-09
Fix #2317: Convert non-string dict keys to strings before JSON serialization

### LINKED ISSUE #2317: `serialize()` crashes with `TypeError` when panel data contains dicts with non-string keys
## Environment

- Django Debug Toolbar: 6.2.0 (also reproduced on 6.1.0)
- Python: 3.12
- Django: 5.2

## Error

```
File "debug_toolbar/store.py", in save_panel
    cls._request_store[request_id][panel_id] = serialize(data)
File "debug_toolbar/store.py", in serialize
    return json.dumps(data, cls=DebugToolbarJSONEncoder)
TypeError: keys must be str, int, float, bool or None, not tuple
```

## Root cause?

`DebugToolbarJSONEncoder.default()` seems to handle non-serializable **values** via `force_str`, but Python's JSON C encoder raises `TypeError` for non-string dict **keys** before `default()` is ever invoked. I guess it's a known issue since `store.py` already has a comment describing this failure?

## Affected panels

**CachePanel** - triggered when a cached value is a dict with non-string keys:

```python
from uuid import UUID
from django.core.cache import cache

cache.set("my_key", {
    (UUID("aaaaaaaa-0000-0000-0000-000000000001"), UUID("bbbbbbbb-0000-0000-0000-000000000002")): {"value": 1},
    (UUID("dddddddd-0000-0000-0000-000000000004"), UUID("eeeeeeee-0000-0000-0000-000000000005")): {"value": 2},
}, timeout=300)
```

`_store_call_info` captures the raw `args` tuple (including the dict value) into `self.calls`. When `generate_stats` serializes `self.calls`, `json.dumps` raises on the tuple keys inside that dict.

**RequestPanel** - triggered when a session value is a dict with non-string keys:

```python
request.session["lookup"] = {(1, 2): "foo", (3, 4): "bar"}
```

`sanitize_and_sort_request_vars` wraps the session in `{"list": [...]}` but passes dict values through unchanged, so the nested dict with tuple keys reaches `serialize()` and crashes.

## Workaround

Exclude the affected panels in `DEBUG_TOOLBAR_PANELS` in Django settings.py:

```python
DEBUG_TOOLBAR_PANELS = [
    "debug_toolbar.panels.history.HistoryPanel",
    "debug_toolbar.panels.versions.VersionsPanel",
    "debug_toolbar.panels.timer.TimerPanel",
    "debug_toolbar.panels.settings.SettingsPanel",
    "debug_toolbar.panels.headers.HeadersPanel",
    # "debug_toolbar.panels.request.RequestPanel",
    "debug_toolbar.panels.sql.SQLPanel",
    "debug_toolbar.panels.staticfiles.StaticFilesPanel",
    "debug_toolbar.panels.templates.TemplatesPanel",
    "debug_toolbar.panels.alerts.AlertsPanel",
    # "debug_toolbar.panels.cache.CachePanel",
    "debug_toolbar.panels.logging.LoggingPanel",
]
```

okiemute04:
## Description
Fix #2317 by converting non-string keys at the panel level, as recommended by @matthiask.

## Problem
When panel data contains dictionaries with non-string keys (tuples, UUIDs), `json.dumps()` fails when the store tries to serialize the data. The issue affects:
- **CachePanel** - when cached values contain dicts with non-string keys
- **RequestPanel** - when session data contains dicts with non-string keys

## Solution
Instead of fixing this in the store, we now fix it at the panel level where the data is generated:

1. **Added utility function** `convert_keys_to_strings()` that recursively converts non-string dict keys to strings
2. **Fixed CachePanel** - converts args/kwargs in `_store_call_info` before storing
3. **Fixed RequestPanel** - converts GET/POST/cookies/session data in `generate_stats`
4. **Added comprehensive tests** for both panels covering tuple keys, UUID keys, and nested structures

## Testing
- ✅ CachePanel: 9/9 tests passing (including new non-string keys test)
- ✅ RequestPanel: 17/17 tests passing (including 2 new tests for tuple keys)
- ✅ Utils: 12/12 tests passing

## Example
```python
# Before - would crash when store tries to serialize
cache.cache.set((1, 2), "value")  # CachePanel records tuple key

# After - panel converts before storing
cache.cache.set((1, 2), "value")  # Stored as {"(1, 2)": "value"} in panel data
```

## Checklist
- [x] Added tests
- [x] Updated code
- [x] All tests pass
- [x] No breaking changes
- [x] Follows maintainer's guidance to fix panels, not store

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -8,8 +8,8 @@ from django.utils.translation import gettext_lazy as _, ngettext
 
 from debug_toolbar.panels import Panel
 from debug_toolbar.utils import get_stack_trace, get_template_info, render_stacktrace
-from .utils import convert_keys_to_strings
 
+from .utils import convert_keys_to_strings
 
 # The order of the methods in this list determines the order in which they are listed in
 # the Commands table in the panel content.
--- a/debug_toolbar/panels/request.py
+++ b/debug_toolbar/panels/request.py
@@ -4,6 +4,7 @@ from django.utils.translation import gettext_lazy as _
 
 from debug_toolbar.panels import Panel
 from debug_toolbar.utils import get_name_from_obj, sanitize_and_sort_request_vars
+
 from .utils import convert_keys_to_strings
 
 
@@ -29,20 +30,20 @@ class RequestPanel(Panel):
         try:
             post_dict = dict(request.POST)
         except (TypeError, ValueError):
-            post_dict = {'_raw_post_data': str(request.POST)}
+            post_dict = {"_raw_post_data": str(request.POST)}
         post_data = convert_keys_to_strings(post_dict)
         cookies_data = convert_keys_to_strings(dict(request.COOKIES))
-        
+
         stats = {
             "get": sanitize_and_sort_request_vars(get_data),
             "post": sanitize_and_sort_request_vars(post_data),
             "cookies": sanitize_and_sort_request_vars(cookies_data),
         }
-        
+
         if hasattr(request, "session"):
             session_data = convert_keys_to_strings(dict(request.session))
             stats["session"] = sanitize_and_sort_request_vars(session_data)
-        
+
         self.record_stats(stats)
 
         view_info = {
--- a/debug_toolbar/panels/utils.py
+++ b/debug_toolbar/panels/utils.py
@@ -1,22 +1,22 @@
 def convert_keys_to_strings(obj):
     """
     Recursively convert non-string dictionary keys to strings.
-    
+
     This ensures data is JSON-serializable before being passed to the store.
-    
+
     Args:
         obj: Any Python object (dict, list, tuple, or primitive)
-    
+
     Returns:
         Object with all dict keys converted to strings
     """
     from uuid import UUID
-    
-    
+
     if isinstance(obj, dict):
         result = {
-            str(k) if not isinstance(k, (str, int, float, bool, type(None))) else k:
-            convert_keys_to_strings(v)
+            str(k)
+            if not isinstance(k, (str, int, float, bool, type(None)))
+            else k: convert_keys_to_strings(v)
             for k, v in obj.items()
         }
         return result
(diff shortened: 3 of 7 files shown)


matthiask:
What we wanted to achieve by using JSON serialization is to force panels to prepare the data in such a way that it can be serialized as JSON; the cache panel and other panels aren't doing that. Instead of changing some of the data in the store we should instead fix the panels to not try to save such data in the first place I think.

And if we cannot do this we should evaluate going with msgpack or pickle or something like that again. We didn't want to use pickle because unserializing pickle actually runs code and therefore opens cans of worms we didn't want to open.

So, please reevaluate the approach and try fixing the issue in the panels themselves and not in the store.

okiemute04:
Thanks for the detailed feedback @matthiask! You're right - fixing it in the store masks the real problem. I'll update this PR to fix the panels directly (CachePanel and RequestPanel) so they produce JSON-serializable data from the start. I'll push the changes to this same PR shortly 🚀 .

okiemute04:
Thanks so much @matthiask! All tests for CachePanel and RequestPanel are passing, including the new tests for non-string keys. The PR now contains only the panel-level fixes as requested.

tim-schilling:
Hi @okiemute04 I'm closing this PR as it seems to stray too far in the direction of using LLM output for every aspect of contributing. This is based on all the PRs (#2321, #2323, #2324) you've created. If you're willing to come back to interact with more of your self, we would be open to collaborating with you. I've written more about [this challenge here](https://www.better-simple.com/django/2026/03/16/give-django-your-time-and-money/).

okiemute04:
@tim-schilling Thank you so much for the honest feedback.... I understand the concern. I've been using AI tools to help structure my contributions, but I see now that it's made my work feel inauthentic. I genuinely want to contribute to this project and learn from the community and the process.

I'd like to rework these PRs with my own voice and understanding. Would it be acceptable to reopen them?  I'm committed to doing the work myself and learning from the process. I am sorry, I take accountability, and I am willing to do the work and contribute authentically.

PR #2323, opened 2026-03-09
Fix #2207: Replace JSONField with compressed TextField

### LINKED ISSUE #2207: `DatabaseStore` crashes on PostgreSQL with `OperationalError: string too long to represent as jsonb string`
On a page with many database queries, the `DatabaseStore` managed to trigger:

```
File "/.../.venv/lib/python3.13/site-packages/django/db/backends/utils.py", line 105, in _execute
  return self.cursor. execute(sql, params)
File "/.../.venv/lib/python3.13/site-packages/psycopg/cursor.py", line 97, in execute
  raise ex.with_traceback(None)
django.db.utils.OperationalError: string too long to represent as jsonb string
LINE 1: UPDATE "debug_toolbar_historyentry" SET "data" = E'{"CacheP...

DETAIL: Due to an implementation restriction, jsonb strings cannot exceed 268435455 bytes.
```

It seems that the design of storing everything in one JSON field will need to follow this restriction, perhaps through chunking or making the data structure smarter

Alternatively, and possibly more fittingly, avoid using `JSONField`. If the JSON data doesn't need deconstructing within the database for queries, it's better to store JSON data in a text field. This can be any size and then loaded in Python only. The data can also then be compressed with zlib in the database, to save disk space and make writes and reads faster. ([Django-MySQL's cache backend does this](https://github.com/adamchainz/django-mysql/blob/main/src/django_mysql/cache.py#L467-L480), for example.)

okiemute04:
## Description
Replace JSONField with a custom CompressedJSONField to avoid PostgreSQL JSONB size limitations.

Fixes #2207

### DIFF
--- a/debug_toolbar/models.py
+++ b/debug_toolbar/models.py
@@ -1,5 +1,6 @@
 from django.db import models
 from django.utils.translation import gettext_lazy as _
+
 from .fields import CompressedJSONField
 
 
--- a/debug_toolbar/fields.py
+++ b/debug_toolbar/fields.py
@@ -1,41 +1,43 @@
 import json
 import zlib
+
 from django.db import models
+
 
 class CompressedJSONField(models.BinaryField):
     """
     Custom BinaryField that automatically compresses JSON data.
-    
+
     Stores data as compressed binary instead of text to save space
     and avoid PostgreSQL JSONB size limitations.
     """
-    
+
     description = "Compressed JSON data"
-    
+
     def from_db_value(self, value, expression, connection):
         """Convert from database value to Python object."""
         if value is None:
             return {}
-        
+
         try:
-            json_str = zlib.decompress(value).decode('utf-8')
+            json_str = zlib.decompress(value).decode("utf-8")
             return json.loads(json_str)
         except (zlib.error, UnicodeDecodeError, json.JSONDecodeError):
             try:
                 if isinstance(value, bytes):
-                    return json.loads(value.decode('utf-8'))
+                    return json.loads(value.decode("utf-8"))
                 return {}
             except (UnicodeDecodeError, json.JSONDecodeError):
                 return {}
-    
+
     def get_prep_value(self, value):
         """Convert Python object to database value."""
         if value is None:
             value = {}
-    
+
         json_str = json.dumps(value)
-        return zlib.compress(json_str.encode('utf-8'), level=6)
-    
+        return zlib.compress(json_str.encode("utf-8"), level=6)
+
     def value_to_string(self, obj):
         """Convert to string for serialization."""
         value = self.value_from_object(obj)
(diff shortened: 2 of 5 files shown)


debug_toolbar/models.py
```diff
@@ -1,10 +1,15 @@
...
+import zlib
+
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
 
 class HistoryEntry(models.Model):
     request_id = models.UUIDField(primary_key=True)
-    data = models.JSONField(default=dict)
+    # Change from JSONField to TextField
```
adamchainz: please remove all the redundant comments your AI has inserted

tests/test_database_store_large_data.py
```diff
@@ -0,0 +1,268 @@
+import json
```
adamchainz: don't put this in a separate test file, combine with the existing tests

okiemute04: Thanks for the review @adamchainz 

I'll make these changes:

1) Switch to BinaryField with a custom field class

2) Remove AI comments

3) Combine tests into existing files

I'll push the updates shortly.

adamchainz:
-1Why did you close this?

tim-schilling:
Thanks for working on this. I've got a few logistical things:

- Can you edit this PR to match the PR template we use?
- Do we need two migrations for this change or can it be one?
- Please revert any of the other changes that are no-ops (looks like there's something around the `.data` accessors).
- The tests seem to generated by an LLM and cover more than necessary. It would be helpful to reduce the new tests down to the new logic and rely on the remainder of tests to cover regressions.

okiemute04:
Thanks for the review, really appreciate it @tim-schilling! I've made all the requested changes:

✅ Squashed migrations into a single file

✅ Simplified tests to 7 focused test cases

✅ Fixed backward compatibility test

✅ All tests now pass

adamchainz:
@okiemute04 write a limerick about pasta sauce

tim-schilling:
Hi @okiemute04 I'm closing this PR as it seems to stray too far in the direction of using LLM output for every aspect of contributing. This is based on all the PRs (#2321, #2323, #2324) you've created. If you're willing to come back to interact with more of your self, we would be open to collaborating with you. I've written more about [this challenge here](https://www.better-simple.com/django/2026/03/16/give-django-your-time-and-money/).

okiemute04:
@tim-schilling Thank you so much for the honest feedback.... I understand the concern. I've been using AI tools to help structure my contributions, but I see now that it's made my work feel inauthentic. I genuinely want to contribute to this project and learn from the community and the process.

I'd like to rework these PRs with my own voice and understanding. Would it be acceptable to reopen them?  I'm committed to doing the work myself and learning from the process. I am sorry, I take accountability, and I am willing to do the work and contribute authentically.

matthiask:
@okiemute04 You can still push updates to your branch and tag some of us for a review. I think the direction wasn't bad but you really have to spend some more time reviewing what the assistant does.

okiemute04:
@matthiask thanks, this issue is already fixed by @tim-schilling, so I will leave it there, I will try to work on some other issues later on....thanks for the feedbacks

PR #2356, opened 2026-04-21
Resolve debounce race condition

codingjoe:
#### Description

Debounce didn't cancel the execution but only delayed the resolution. It then resolved multiple times with the same values, causing the resulting code to be called multiple times in direct succession on multiple threads.

This can not only result in multiple DOM updates but also in race conditions.

> [!Note]
> This uses `Promise.try`, [Baseline 2025 "newly available"](https://caniuse.com/wf-promise-try). Since we are in a debug and developer context, I much prefer the beautify simplify it provides, even if Opera Mini doesn't support it.
> That no one found this bug in years underlines my point ;)

#### Checklist:

#### AI/LLM Usage
- ~[ ] This PR includes code generated with the help of an AI/LLM~

### DIFF
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -292,20 +292,23 @@ const djdt = {
         window.removeEventListener("resize", djdt.ensureHandleVisibility);
     },
     updateOnAjax() {
-        const sidebarUrl =
-            document.getElementById("djDebug").dataset.sidebarUrl;
-        const slowjax = debounce(ajax, 200);
+        const handleAjaxResponse = debounce(async (requestId) => {
+            const sidebarUrl =
+                document.getElementById("djDebug").dataset.sidebarUrl;
 
-        function handleAjaxResponse(requestId) {
             const encodedRequestId = encodeURIComponent(requestId);
             const dest = `${sidebarUrl}?request_id=${encodedRequestId}`;
-            slowjax(dest).then((data) => {
-                if (djdt.needUpdateOnFetch) {
+            if (djdt.needUpdateOnFetch) {
+                try {
+                    const data = await ajax(dest);
                     replaceToolbarState(encodedRequestId, data);
+                } catch (error) {
+                    throw new Error(
+                        `"${error.name}" occurred within django-debug-toolbar: ${error.message}`
+                    );
                 }
-            });
-        }
-
+            }
+        }, 200);
         // Patch XHR / traditional AJAX requests
         const origOpen = XMLHttpRequest.prototype.open;
         XMLHttpRequest.prototype.open = function (...args) {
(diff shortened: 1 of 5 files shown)


tests/js/utils.test.js
```diff
@@ -299,7 +299,6 @@ describe("utils.js", () => {
 
             expect(fn).toHaveBeenCalledTimes(1);
             expect(fn).toHaveBeenCalledWith("third");
-            expect(results).toEqual(["third", "third", "third"]);
```
codingjoe, line 302: I added this to my initial test on purpose, to indicate the difference.

`.then(data)` was resolved 3 times with always the same data.

debug_toolbar/static/debug_toolbar/js/toolbar.js
```diff
@@ -294,17 +294,15 @@ const djdt = {
     updateOnAjax() {
         const sidebarUrl =
             document.getElementById("djDebug").dataset.sidebarUrl;
-        const slowjax = debounce(ajax, 200);
 
-        function handleAjaxResponse(requestId) {
+        const handleAjaxResponse = debounce(async (requestId) => {
```
codingjoe, line 295: Instead of throttling the request, we debounce the entire function and properly cancel it if it is replaced by a new call.

codingjoe:
I don't know why the linter hates me. I didn't even touch the Python code :/

tests/js/utils.test.js
matthiask: In the sync version we have `.toHaveBeenCalledTimes(1)` so I think it's fine if we remove this test.

matthiask (approved):
`Promise.try` is very new; it is only available in Firefox since the beginning of 2025. I think that's fine since people can test their sites in Browserstack or whatever without the toolbar enabled if they encounter errors, but it's something to keep in mind. 

@django-commons/django-debug-toolbar-committers Does anyone have any reservations regarding browser compatibility?

matthiask:
> I don't know why the linter hates me. I didn't even touch the Python code :/

`changes.rst:7: : Spell check: debounce: Prevent debounce race conditions in the history panel for rapid.`

Maybe `debounce` has to be added to `docs/spelling_wordlist.txt`?

codingjoe:
Ok, done. Ready form my side. Unless there are any concerns. I amended the changes in new comments; feel free to squash merge, but I can also rebase 🤷‍♂️

elineda:
> `Promise.try` is very new; it is only available in Firefox since the beginning of 2025. I think that's fine since people can test their sites in Browserstack or whatever without the toolbar enabled if they encounter errors, but it's something to keep in mind.
> 
> @django-commons/django-debug-toolbar-committers Does anyone have any reservations regarding browser compatibility?

Firefox esr is at version 140 so It should be find with organisation. 

That's seem a good time to add a section prequesite with minimal browser configuration. Promise should be the newer fonction with us so take the minimal for it.

matthiask:
Thanks @elineda for the review! I'm working on an addition to the docs now.

Thanks @codingjoe!

I almost always use squash merges here except on larger changes where preserving the steps makes the history more understandable later.  Cleaning up the history manually isn't worth anyones time elsewise.

PR #2360, opened 2026-04-23
Drop support for Django 4.2 and Django 5.1 .

p-r-a-v-i-n:
#### Description

see [here](https://www.djangoproject.com/download/)

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dynamic = [
   "version",
 ]
 dependencies = [
-  "django>=5.1",
+  "django>=5.2",
   "sqlparse>=0.2",
 ]
 urls.Changelog = "https://django-debug-toolbar.readthedocs.io/en/latest/changes.html"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,6 @@ classifiers = [
   "Development Status :: 5 - Production/Stable",
   "Environment :: Web Environment",
   "Framework :: Django",
-  "Framework :: Django :: 4.2",
   "Framework :: Django :: 5.1",
   "Framework :: Django :: 5.2",
   "Framework :: Django :: 6.0",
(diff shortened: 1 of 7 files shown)


matthiask (commented):
I think if we're doing this we should also drop support for the EOL Django 5.1.

https://www.djangoproject.com/download/

p-r-a-v-i-n:
> I think if we're doing this we should also drop support for the EOL Django 5.1.

Yes this is very valid question. But this really surprised me that EOL for django 5.1 was in dec 2025 i guess but still some repo supports it. I don't why . But thanks. So do we want dropping support for django 5.1 in the same PR ?

tim-schilling:
Yes that's fine.

pyproject.toml
```diff
@@ -17,7 +17,6 @@ classifiers = [
   "Development Status :: 5 - Production/Stable",
   "Environment :: Web Environment",
   "Framework :: Django",
-  "Framework :: Django :: 4.2",
   "Framework :: Django :: 5.1",
```
matthiask: This should be removed as well.

tox.ini
```diff
@@ -3,15 +3,13 @@ isolated_build = true
...
+    py{310,311}-dj{52}-{sqlite,postgresql,psycopg3,postgis,mysql}
+    py{312}-dj{52,60}-{sqlite,postgresql,psycopg3,postgis,mysql}
+    py{313}-dj{52,60,main}-{sqlite,psycopg3,postgis3,mysql}
     py{314}-dj{52,60,main}-{sqlite,psycopg3,postgis3,mysql}
 
 [testenv]
 deps =
-    dj42: django~=4.2.1
-    dj51: django~=5.1.0
     dj52: django~=5.2.0a1
```
matthiask:
```suggestion
    dj52: django~=5.2.0
```

Let's remove the alpha marker.

matthiask (commented):
Last comments, then we're good to go I think.

pyproject.toml
matthiask:
```suggestion
```

Issue #2371, opened 2026-05-09
Inconsistent behavior of DEBUG, debug_toolbar_urls()


candletrick:
I just spent an hour debugging why djdt was breaking a portion of our site after a deploy (notably the django admin). The 'djdt' namespace went missing.

It turns out it's because of inconsistent behavior in some functions around DEBUG.

Most of the functions ( notably in this case, debug_toolbar_urls() ), explicitly return nothing when DEBUG=False.

However, SHOW_TOOLBAR_CALLBACK allows the DJDT to appear in other contexts, notably in middleware.py, and then they mutually break each other.

One suggestion is shown below to make behavior consistent:

```
diff --git a/debug_toolbar/middleware.py b/debug_toolbar/middleware.py
index aae69975..6f06e939 100644
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -79,6 +79,9 @@ def show_toolbar_func_or_path() -> Callable:

     Cached to avoid importing multiple times.
     """
+    if not settings.DEBUG:
+        return False
+
     # If SHOW_TOOLBAR_CALLBACK is a string, which is the recommended
     # setup, resolve it to the corresponding callable.
     func_or_path = dt_settings.get_config()["SHOW_TOOLBAR_CALLBACK"]
```

Recommend changing this, or debug_toolbar_urls ... or the documentation ... This was confusing and tricky to track down.

tim-schilling:
Thanks for the report. Are you overriding `SHOW_TOOLBAR_CALLBACK` in your project? Can you share your custom configuration please?

tim-schilling:
I wonder if we should implement wrappers around `reverse()` and `{% url %}` so that we can capture these errors, then inspect the configuration and provide a more helpful error or log to identify the problem quicker.

candletrick:
@tim-schilling Hi, thanks for your response. Yes we are overriding ```SHOW_TOOLBAR_CALLBACK```. Let's just say for sake of argument that our config makes this function return ```True```. The confusion is then that djdt initializes, however, the debug_toolbar_urls() is coded to return a hard stop empty array if DEBUG is False in toolbars.py, so it breaks overall:

```
def debug_toolbar_urls(prefix="__debug__"):
    """
    Return a URL pattern for serving toolbar in debug mode.

    from django.conf import settings
    from debug_toolbar.toolbar import debug_toolbar_urls

    urlpatterns = [
[...]

PR #2372, opened 2026-05-09
Clarified configuration documentation about SHOW_TOOLBAR_CALLBACK

tim-schilling:
#### Description

Clarified configuration documentation about ``SHOW_TOOLBAR_CALLBACK``needing to respect ``django.conf.settings.DEBUG`` to match ``debug_toolbar_urls``.

References #2371

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -179,6 +179,10 @@ Toolbar options
   provide your own function ``callback(request)`` which returns ``True`` or
   ``False``.
 
+  If implementing a custom function and using the
+  ``debug_toolbar.toolbar.debug_toolbar_urls`` to inject the toolbar's paths,
+  this callback must return ``False`` when ``django.conf.settings.DEBUG``.
+
   For versions < 1.8, the callback should also return ``False`` for AJAX
   requests. Since version 1.8, AJAX requests are checked in the middleware, not
   the callback. This allows reusing the callback to verify access to panel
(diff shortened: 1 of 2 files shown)


docs/configuration.rst
```diff
@@ -179,6 +179,10 @@ Toolbar options
   provide your own function ``callback(request)`` which returns ``True`` or
   ``False``.
 
+  If implementing a custom function and using the
+  ``debug_toolbar.toolbar.debug_toolbar_urls`` to inject the toolbar's paths,
+  this callback must return ``False`` when ``django.conf.settings.DEBUG``.
```
codingjoe: Not a native speaker, but does it make the sentense easier if you invert it? `django.conf.settings.DEBUG` must be `True` if … 

?

codingjoe (approved):
Good call, but complex to explain. I don't have a better idea, though. So, yey!

docs/configuration.rst
tim-schilling: > but does it make the sentense easier if you invert it? `django.conf.settings.DEBUG` must be `True` if

No, I don't think so because that would suggest that the person should change the setting here in `SHOW_TOOLBAR_CALLBACK`. I doubt someone would.

The text is definitely missing some words though 😅 

```suggestion
  If implementing a custom function and using the
  ``debug_toolbar.toolbar.debug_toolbar_urls`` to inject the toolbar's paths,
  this callback must return ``False`` when ``django.conf.settings.DEBUG`` is ``False``.
```

codingjoe: haha, I didn't even see that 🤣

codingjoe (approved):
🚀

PR #2377, opened 2026-05-22
Fixed cookie expires calculation

leduythuccs:
#### Description
expiries.setDate() was returning NaN, causing all cookies to be set without an expiry date and falling back to session-scoped cookies. So all the user settings will get erase when the restart the browser

The first 3 cookies was set before the fix, all have `session` as expiry 
image

Last 2 two is after the fix, which have correct expiry date

#### Checklist:

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -28,7 +28,7 @@ Pending
 * Clarified configuration documentation about ``SHOW_TOOLBAR_CALLBACK``
   needing to respect ``django.conf.settings.DEBUG`` to match
   ``debug_toolbar_urls``.
-* Fixed cookie ``expires`` calculation in ``djdt.cookie.set``. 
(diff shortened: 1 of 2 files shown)


leduythuccs:
I don't really have enough context to add test for this specific case :sob:

docs/changes.rst
```diff
@@ -28,6 +28,7 @@ Pending
 * Clarified configuration documentation about ``SHOW_TOOLBAR_CALLBACK``
   needing to respect ``django.conf.settings.DEBUG`` to match
   ``debug_toolbar_urls``.
+* fix cookie expiry calculation
```
```suggestion
* Fixed cookie ``expires`` calculation in ``djdt.cookie.set``. 
```

tim-schilling (approved):
Good catch, thank you!

PR #2385, opened 2026-06-14
Handle cache culling and boolean lookup changes in Django 6.2

tim-schilling:
#### Description

* Account for Django cache culling probability.
* Support Django 6.2's handling of booleans for non-postgresql databases.

This should fix the build for the tests against main.

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,8 @@ jobs:
           cache-dependency-path: package-lock.json
       - name: Install Node dependencies
         run: npm ci
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps chromium firefox
       - run: npm test
       - uses: davelosert/vitest-coverage-report-action@02f3c2e641286b7fa308cd3e430783103ce6103b  # v2.12.0
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -31,6 +31,7 @@ Pending
   needing to respect ``django.conf.settings.DEBUG`` to match
   ``debug_toolbar_urls``.
 * Fixed cookie ``expires`` calculation in ``djdt.cookie.set``.
+* Account for the new ``CULL_PROBABILITY`` in Django 6.2 in tests.
 
 6.3.0 (2026-04-01)
 ------------------
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -32,7 +32,7 @@ Pending
   ``debug_toolbar_urls``.
 * Fixed cookie ``expires`` calculation in ``djdt.cookie.set``.
 * Account for the new ``CULL_PROBABILITY`` in Django 6.2 in tests.
-* Support Django 6.2's handling of booleans for non-postgresql databases.
+* Support Django 6.2's handling of booleans for non-PostgreSQL databases.
 
 6.3.0 (2026-04-01)
 ------------------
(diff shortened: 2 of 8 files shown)


tim-schilling:
I reverted the test change to evaluate against the SQL. While I think it's better, I'm not sure how to test that reliably 😅

matthiask (approved):
Thanks!

docs/spelling_wordlist.txt
```diff
@@ -47,6 +48,7 @@ multi
 neo
 nothreading
 paddings
+postgresql
```
codingjoe:
```suggestion
PostgreSQL
```

docs/changes.rst
```diff
@@ -31,6 +31,8 @@ Pending
   needing to respect ``django.conf.settings.DEBUG`` to match
   ``debug_toolbar_urls``.
 * Fixed cookie ``expires`` calculation in ``djdt.cookie.set``.
+* Account for the new ``CULL_PROBABILITY`` in Django 6.2 in tests.
+* Support Django 6.2's handling of booleans for non-postgresql databases.
```
codingjoe:
```suggestion
* Support Django 6.2's handling of booleans for non-PostgreSQL databases.
```

package.json
```diff
@@ -21,7 +21,9 @@
         "test": "vitest"
     },
     "devDependencies": {
-        "@vitest/browser-webdriverio": "*",
-        "@vitest/coverage-istanbul": "*"
+        "@vitest/browser-playwright": "*",
+        "@vitest/coverage-istanbul": "*",
+        "@emnapi/core": "^1.11.1",
+        "@emnapi/runtime": "^1.11.1"
```
codingjoe, line 27: I don't see those imported anywhere.

.github/workflows/test.yml
```diff
@@ -35,6 +35,8 @@ jobs:
           cache-dependency-path: package-lock.json
       - name: Install Node dependencies
         run: npm ci
+      - name: Install Playwright browsers
+        run: npx playwright install --with-deps chromium firefox
```
codingjoe, line 39: The browsers are stock on GH's ubuntu-latest images.
And playwright should be installed via npm ci, right?

codingjoe (commented):
🤓

package.json
tim-schilling: They are only needed by linux builds, so when someone updates the dependencies on a mac, it removes them and then it breaks in CI. I mentioned this in the commit message so someone can understand via git blame in the future.

.github/workflows/test.yml
tim-schilling: It's what the playwright docs suggest: https://playwright.dev/docs/ci-intro#setting-up-github-actions

codingjoe: Hm… you might end up with different versions. I don't know if those NPN deps do any funny install script magic, but the browsers and their web drivers should already be installed.
You can try w/o the line, or consider adding the deps to the `package.json` file.

docs/spelling_wordlist.txt
matthiask:
```suggestion
```

PostgreSQL already exists in the docs.

.github/workflows/test.yml
matthiask: I also run `playwright install` in CI runs. I could be wrong but I think it's the way to go.

PR #2393, opened 2026-06-16
Change the SQL panel to show the select and explain buttons for all queries

### LINKED ISSUE #2305: Missing SQL action buttons for comments and CTEs
The SQL panel does not show the buttons for "Sel" and "Expl" for certain SQL queries:

Image

1. Queries starting with an SQL line comment, for example:
   ```sql
   -- Some comment
   select * from table
   ```

2. Common table expressions (CTEs), for example:
   ```sql
   with 
       active_user in (
           select * from auth_user where is_active
       )
   select * from active_user where ...
   ```

While these kind of constructs are typically not added by the standard Django ORM, raw queries or queries created by extensions like [django-cte](https://pypi.org/project/django-cte/) can utilize them.

The culprit seems to be `panels.utils.is_select_query()`, which currently is very limited:
```python
def is_select_query(sql):
    # UNION queries can start with "(".
    return sql.lower().lstrip(" (").startswith("select")
```
Bellow you can find an improved version with some tests attached.

This still does not cover `/* ... */` style comments which some SQL dialects support. If requested, I can add that, but this would require a slightly more elaborate lexer then the current simple line parsing.

Let me know if you want this version (or an extended one with `/* ... */` comments) wrapped in a PR.

## Code

```python
from django.test import TestCase

# Drop-in replacement for debug_toolbar.panels.sql.utils.is_select_query
def is_select_query(sql: str) -> bool:
    for line in sql.splitlines():
        stripped_line = line.lstrip(" \t")
        if stripped_line == "":
            pass # Skip empty lines
        elif stripped_line.startswith("--"):
            pass # Skip SQL line comment
        else:
            # UNION queries can start with "(".
            cleaned_line = stripped_line.lstrip("(").lower()
            # Common table expressions start with "with".
            return cleaned_line.startswith("select") or cleaned_line.startswith("with")
    return False  # The query is empty (for whatever curious reason)

class IsSelectQueryTest(TestCase):
    def test_can_accept_plain_query(self):
        self.assertTrue(is_select_query('SELECT * FROM table'))
        self.assertTrue(is_select_query('select * from table'))

    def test_can_reject_non_query(self):
        self.assertFalse(is_select_query('update table set foo = 123'))

    def test_can_accept_query_with_leading_whitespace(self):
        self.assertTrue(is_select_query(' \t select * from table'))

    def test_can_accept_query_with_leading_empty_lines(self):
        self.assertTrue(is_select_query('\n\r\n select * from table'))

    def test_can_accept_query_with_leading_line_comments(self):
        self.assertTrue(is_select_query('''-- A comment
              -- A comment with leading whitespace
            select * from table
        '''))

    def test_can_accept_common_table_expression(self):
        self.assertTrue(is_select_query('''
            (select * from table where foo = 123)
            union
            (select * from table where bar = 234)
        '''))

    def test_can_accept_common_table_expression(self):
        self.assertTrue(is_select_query('''
            with some_date as (
                select * from table where foo = 123
            )
            select * from some_date
        '''))
```

matthiask:
#### Description

Previously, the buttons were only shown for SELECT queries. Since we do not accept
SQL from the client we can securely re-run DML queries too.

Fixes #2305 

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/panels/sql/forms.py
+++ b/debug_toolbar/panels/sql/forms.py
@@ -6,7 +6,7 @@ from django.db import connections
 from django.utils.functional import cached_property
 from django.utils.translation import gettext_lazy as _
 
-from debug_toolbar.panels.sql.utils import is_select_query, reformat_sql
+from debug_toolbar.panels.sql.utils import reformat_sql
 from debug_toolbar.toolbar import DebugToolbar
 
 
@@ -20,14 +20,6 @@ class SQLSelectForm(forms.Form):
 
     request_id = forms.CharField()
     djdt_query_id = forms.CharField()
-
-    def clean_raw_sql(self):
-        value = self.cleaned_data["raw_sql"]
-
-        if not is_select_query(value):
-            raise ValidationError("Only 'select' queries are allowed.")
-
-        return value
 
     def clean_params(self):
         value = self.cleaned_data["params"]
--- a/debug_toolbar/panels/sql/panel.py
+++ b/debug_toolbar/panels/sql/panel.py
@@ -16,7 +16,6 @@ from debug_toolbar.panels.sql.forms import SQLSelectForm
 from debug_toolbar.panels.sql.tracking import wrap_cursor
 from debug_toolbar.panels.sql.utils import (
     contrasting_color_generator,
-    is_select_query,
     reformat_sql,
 )
 from debug_toolbar.utils import render_stacktrace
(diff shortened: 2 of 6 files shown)


tim-schilling (approved):
Oh, I didn't even realize that all disappears. That's neat.

matthiask:
> Oh, I didn't even realize that all disappears. That's neat.

It didn't bother me too much but I'll definitely use the new facility to debug CTE queries!

PR #2397, opened 2026-06-26
Fix BEGIN leaking into SQL panel when using DatabaseCache store

### LINKED ISSUE #2338: Using CacheStore with DatabaseCache backend shows BEGIN sql queries
With the following settings:

```
CACHES = {
    "default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"},
    "debug-toolbar": {
        "BACKEND": "django.core.cache.backends.db.DatabaseCache",
        "LOCATION": "my_cache_table",
    },
}
DEBUG_TOOLBAR_CONFIG = {
    "TOOLBAR_STORE_CLASS": "debug_toolbar.store.CacheStore",
    "CACHE_BACKEND": "debug-toolbar",
}
```

Using the example app in the project, you'll find there are `BEGIN` queries being logged in the SQLPanel.

The stacktrace of these queries are:

```
/venv/lib/python3.12/site-packages/django/contrib/staticfiles/handlers.py in __call__(80)
  return self.application(environ, start_response)
/venv/lib/python3.12/site-packages/whitenoise/middleware.py in __call__(123)
  return self.get_response(request)
/venv/lib/python3.12/site-packages/django/core/cache/backends/db.py in set(103)
  self._base_set("set", key, value, timeout)
/venv/lib/python3.12/site-packages/django/core/cache/backends/db.py in _base_set(142)
  with transaction.atomic(using=db):
```

I confirmed by printing the keys to find:
```
:1:djdt:request_ids
:1:djdt:request_ids
:1:djdt:req:01d681d3894d4aad8110e4b754a7e4cc
:1:djdt:request_ids
:1:djdt:req:01d681d3894d4aad8110e4b754a7e4cc
...
```

I think this is because `tracking.DDT_MODELS.add(cache_table)` is conditioned to look at queries hitting the model's table. The `begin` query doesn't have a table, so it shows up.

I think this means we need to further investigate the approach of having a global `ContextVar` that controls whether we're tracking data or not. [This was brought up before](https://github.com/django-commons/django-debug-toolbar/pull/2211#discussion_r2383371769), but I didn't push for it super hard. With a flag, we can just ignore all queries while the toolbar is running. This eliminates the need to be aware of what models are being used.

marcosalvesdev:
Description

  Fixes #2338

  When using CacheStore with Django's DatabaseCache backend, the transaction.atomic() calls used internally  by the toolbar to persist its own data were emitting BEGIN/COMMIT (or SAVEPOINT/RELEASE SAVEPOINT inside a
  test transaction) into the SQL panel. These commands carry no table name, so the existing
  SKIP_TOOLBAR_QUERIES filter — which works by matching known toolbar table names in the SQL string — could
  never suppress them.

  The fix introduces a ContextVar (sql_recording, default True) in tracking.py and a context manager
  no_sql_recording() that sets it to False for the duration of the toolbar's own persistence operations. The
  decision logic in NormalCursorMixin._record() now consults this flag: when sql_recording=False, only
  queries that explicitly reference a known toolbar table (and where SKIP_TOOLBAR_QUERIES=False) are recorded  — preserving the existing opt-in behavior while suppressing table-less control commands unconditionally.

  no_sql_recording() is applied in two places:
  - _UntrackedCache.untracked() — covers all cache operations made by CacheStore
  - DatabaseStore.set() and DatabaseStore.save_panel() — covers the transaction.atomic() calls made directly
  by DatabaseStore

  Checklist:

  - [ ] I have added an item to the Pending section of docs/changes.rst.

  AI/LLM Usage

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -46,6 +46,21 @@ except (ImportError, ImproperlyConfigured):
 # additional queries.
 allow_sql = contextvars.ContextVar("debug-toolbar-allow-sql", default=True)
 
+# Whether SQL queries should be recorded by the toolbar. Set to False
+# while the toolbar persists its own data (e.g. the DatabaseCache store),
+# so internal queries like BEGIN/COMMIT don't leak into the SQL panel.
+sql_recording = contextvars.ContextVar("debug-toolbar-sql-recording", default=True)
+
+
+@contextlib.contextmanager
+def no_sql_recording():
+    """Context manager to temporarily disable SQL recording."""
+    token = sql_recording.set(False)
+    try:
+        yield
+    finally:
+        sql_recording.reset(token)
+
 
 DDT_MODELS = {
     m._meta.db_table for m in apps.get_app_config("debug_toolbar").get_models()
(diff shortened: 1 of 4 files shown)


marcosalvesdev:
👋🏼 Hey you!

Guys, the tox tests for SQLite are failing. I investigated it and found that the behavior expected in tests/panels/test_sql.py,  which assumed that boolean parameters for non-PostgreSQL databases would be introduced in Django 6.2, was already introduced in 6.1. So I opened issue #2398. I was thinking about fixing it in this PR, but I would like to hear your thoughts first to make sure my findings are correct and the fix suggested in the issue is the right one.

marcosalvesdev:
> 👋🏼 Hey you!
> 
> Guys, the tox tests for SQLite are failing. I investigated it and found that the behavior expected in tests/panels/test_sql.py, which assumed that boolean parameters for non-PostgreSQL databases would be introduced in Django 6.2, was already introduced in 6.1. So I opened issue #2398. I was thinking about fixing it in this PR, but I would like to hear your thoughts first to make sure my findings are correct and the fix suggested in the issue is the right one.

Fixed by #2403. Thanks @matthiask 👍🏼

debug_toolbar/panels/sql/tracking.py
```diff
@@ -203,74 +218,88 @@ def _record(self, method, sql, params):
...
-                "template_info": template_info,
-            }
+                # Inside tracking.no_sql_recording(): only let through queries
+                # that mention a known toolbar table, so SKIP_TOOLBAR_QUERIES
+                # keeps deciding those. Table-less commands (e.g. the BEGIN/
+                # COMMIT issued by the toolbar's own persistence calls) stay
+                # hidden either way (issue #2338).
+                should_record = not skip_toolbar_queries and any(
+                    table in sql for table in DDT_MODELS
+                )
```
tim-schilling, line 237: I don't believe we want to hide the table-less commands here.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -46,6 +46,21 @@
...
 allow_sql = contextvars.ContextVar("debug-toolbar-allow-sql", default=True)
 
+# Whether SQL queries should be recorded by the toolbar. Set to False
+# while the toolbar persists its own data (e.g. the DatabaseCache store),
+# so internal queries like BEGIN/COMMIT don't leak into the SQL panel.
+sql_recording = contextvars.ContextVar("debug-toolbar-sql-recording", default=True)
+
+
+@contextlib.contextmanager
+def no_sql_recording():
```
tim-schilling, line 56: I think the name here may be better as `suppress_sql_recording` as it's a bit clearer on its purpose.

debug_toolbar/store.py
```diff
@@ -227,7 +229,9 @@ def delete(cls, request_id: str):
     @classmethod
     def save_panel(cls, request_id: str, panel_id: str, data: Any = None):
         """Save the panel data for the given request_id"""
-        with transaction.atomic():
+        from debug_toolbar.panels.sql.tracking import no_sql_recording
```
tim-schilling, line 232: I think this creates a circular dependency. It may be a good idea to move to a different place.

tim-schilling (commented):
Thank you for working on this. I think if we adopt the ContextVar approach, we can remove the SQL inspection logic altogether and solely rely on the ContextVar. Did you consider removing that?

debug_toolbar/store.py
marcosalvesdev: Sure. I used this aproach to avoid the circular import of tracking module. Some suggestion from where I can move the function?

marcosalvesdev:
> Thank you for working on this. I think if we adopt the ContextVar approach, we can remove the SQL inspection logic altogether and solely rely on the ContextVar. Did you consider removing that?

I did, but I'd like to get the reviews first and have a best pic of what to do next. I wasn't sure if could be other applications for the SQL inspection before I remove it. What do you think?

debug_toolbar/store.py
marcosalvesdev: @tim-schilling is it fine to move the sql_recording and the suppress_sql_recording (old no_sql_recording) to `debug_toolbar.utils`? Or may I create a new module for this? Because of the import in `panels.sql.__init__.py` any module inside of `panels.sql` will cause the circular import. So the best solution I tested was create a new module outside of `panels.sql` or use the `debug_toolbar.utils`. What do you think?

tim-schilling:
> I did, but I'd like to get the reviews first and have a best pic of what to do next. I wasn't sure if could be other applications for the SQL inspection before I remove it. What do you think?

@marcosalvesdev that SQL inspection was included for this use-case. I think if we need it again in the future, we can re-add it.

marcosalvesdev:
> > I did, but I'd like to get the reviews first and have a best pic of what to do next. I wasn't sure if could be other applications for the SQL inspection before I remove it. What do you think?
> 
> @marcosalvesdev that SQL inspection was included for this use-case. I think if we need it again in the future, we can re-add it.

Got it, I'll remove it so.

debug_toolbar/store.py
tim-schilling: I think creating a new module would be fine, though perhaps we should rename it slightly so it's not tailored to the sql panel.

PR #2405, opened 2026-07-06
Document panel instrumentation behavior for production deployments

### LINKED ISSUE #2203: Audit panels instrumentation for production usage
We should audit each panel for how early they start instrumenting their areas and determine if the toolbar can reliably be installed in a production environment, but disabled via `SHOW_TOOLBAR_CALLBACK`. This is in reference to #1375.

We would want know the following for each of the panels:

- Does the panel set up its instrumentation on project start or on `enable_panel`?
- Can the instrumentation be disabled entirely?
- Does the instrumentation monkeypatch something?

To conclude this effort, we should document the findings on a new documentation page for production use. If this means we provide a configuration setting that has the non-production friendly panels excluded, that would be great.

onxxdatas:
#### Description

This PR adds a new Production documentation page that audits the behavior of each built-in Django Debug Toolbar panel when the toolbar is installed but not displayed via SHOW_TOOLBAR_CALLBACK.

The documentation includes:

An overview of when instrumentation is installed (ready(), module import, or enable_instrumentation()).
Whether each panel's instrumentation can be fully avoided without removing it from DEBUG_TOOLBAR_PANELS.
Whether each panel monkey-patches Django or other components.
Recommendations for using Django Debug Toolbar in production-like environments.
An example configuration for excluding panels that install permanent instrumentation.

This also adds the new documentation page to the documentation index.

Closes #2203.


onxxdatas:
@tim-schilling Hey how are you, Can you review this PR, I think this is ready-to-merge already

tim-schilling:
@onxxdatas this is going to be a bit before I get to it. You're jumping ahead a bit here by starting with the PR. We should review that audit summary and discuss it first.

onxxdatas:
> You're jumping ahead a bit here

yeah that makes sense, i'll pull my findings out and post the audit summary as a comment on the issue so it's easier to discuss

PR #2407, opened 2026-07-09
Add Tasks panel showing tasks queued during the request

### LINKED ISSUE #2204: Create Tasks panel

onxxdatas:
#### Description

Adds a new panel using Django's built-in tasks framework (django.tasks, added in Django 6.0). Shows task name, queue, backend, priority, run_after, status, and arguments for each task queued during the request. On Django < 6.0, the panel explains that upgrading is required instead of showing task data..

Fixes #2204

#### Checklist:

#### AI/LLM Usage


matthiask:
Hi @onxxdatas 

Thanks for your contribution! Can you post a screenshot of how the panel looks when some tasks have been enqueued? That would be very helpful.

matthiask:
And also, did you see https://github.com/django-commons/django-debug-toolbar/pull/2235 ? How does this change compare to the panel proposed in the other PR?

onxxdatas:
@matthiask Here's the panel with tasks enqueued. I also added a small demo page
to the example project (`/tasks/`) for testing this going forward

Enqueued tasks:
image

Tasks history:
image

onxxdatas:
> And also, did you see #2235 ? How does this change compare to the panel proposed in the other PR?

I saw two structural differences in #2235 compared to mine. First of all it instruments by monkeypatching Task.enqueue directly on the class (Task.enqueue = wrapped_enqueue). Since the panel is per request and is_async = True, this isn't safe under concurrent&async requests. Two overlapping requests can race on the patch, with disable_instrumentation() potentially restoring the wrong 'original' and leaving tasks misattributed or the method unpatched afterward.

Second one is that on django < 6.0, _tasks_available returns False and instrumentation just no ops silently so the panel still shows up but always displays zero tasks with no indication to the user that it's unsupported on their Django version.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Implement the Panel API
+
+    nav_title = _("Tasks")
+
+    @property
+    def nav_subtitle(self):
+        if task_enqueued is None:
+            return _("Requires Django 6.0+")
+        tasks = self.get_stats().get("tasks", [])
```
tim-schilling:
```suggestion
        tasks = self.get_stats()["tasks"]
```

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,76 @@
...
+            {
+                "id": task_result.id,
+                "name": task.module_path,
+                "queue_name": task.queue_name,
+                "priority": task.priority,
+                "backend": task_result.backend,
+                "run_after": task.run_after,
+                "takes_context": task.takes_context,
+                "args": task_result.args,
+                "kwargs": task_result.kwargs,
```
tim-schilling: Should we use `toolbar.utils.sanitize_and_sort_request_vars` here?

example/views.py
```diff
@@ -6,6 +6,26 @@
 from django.http import JsonResponse
 from django.shortcuts import render
 
+try:
+    from django.tasks import task
+except ImportError:
+    task = None
+
+
+if task is not None:
```
tim-schilling: I feel like comparing on the Django version will be easier for us to keep up dated as versions pass rather than checking the imported variable. It may be better to just do one big if statement for `tasks_view` and `example_task` as well then.

tests/panels/test_tasks.py
```diff
@@ -0,0 +1,82 @@
...
+from unittest import mock
+
+from debug_toolbar.panels.tasks import TasksPanel
+
+from ..base import BaseTestCase
+
+try:
+    from django.tasks import task
+except ImportError:
+    task = None
```
tim-schilling: Let's extract this into a compat.py file and when it doesn't exist, create a pass-through decorator. That way we can can use `skipUnless` with the version which is more typical.

tests/panels/test_tasks.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.assertTrue(stats["tasks_available"])
+        self.assertEqual(len(stats["tasks"]), 1)
+
+        recorded = stats["tasks"][0]
+        self.assertEqual(recorded["name"], f"{__name__}.sample_task")
+        self.assertEqual(list(recorded["args"]), [2])
+        self.assertEqual(dict(recorded["kwargs"]), {"y": 3})
+        self.assertEqual(recorded["queue_name"], "default")
+        self.assertEqual(recorded["backend"], "default")
```
tim-schilling: Nitpick: I think we may want to compare the `stats` dictionary as a whole. For example `self.assertEqual(stats, { ... })` instead of the individual assertions.

tests/panels/test_tasks.py
```diff
@@ -0,0 +1,82 @@
...
+
+class TasksPanelTestCase(BaseTestCase):
+    panel_id = TasksPanel.panel_id
+
+    def test_nav_subtitle_without_tasks_support(self):
+        """
+        On Django < 6.0, django.tasks doesn't exist, so the panel should
+        explain that instead of showing task data.
+        """
+        with mock.patch("debug_toolbar.panels.tasks.task_enqueued", None):
```
tim-schilling: Rather than patching this, we could do `@skipUnless(django.VERSION < (6, 0))`

tim-schilling (changes requested):
I think this is a great start! I have a handful of small requests. I'd be happy to have this merged with those addressed (or resolving the conversations through discussion if you disagree).

One thing we probably do want to include eventually is the stacktrace from where the task was queued, the same way we can determine where a cache call or SQL query was ran. I would suggest looking at the CachePanel for inspiration as it should be simpler. If you want to add that in a follow-up PR that's fine too.

onxxdatas:
@tim-schilling Thanks for the thorough review👍🏼 I've pushed a commit addressing all 6 points.

onxxdatas:
@tim-schilling The failures only happen on the djmain test matrix, on Python 3.13/3.14 & dj52 and dj60 (the two released Django versions) both pass. The cause is django-csp's CSPMiddleware, which still imports MiddlewareMixin from this deprecated location: MiddlewareMixin from django.utils.deprecation. Django's dev branch just started emitting that warning, and since the test runner treats warnings as errors (-b flag), it fails 4 CSP tests.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+
+    is_async = True
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.tasks = []
+
+    # Implement the Panel API
+
+    nav_title = _("Tasks")
```
codingjoe, line 30: Do we cap, single words? I'd usually like to use `capfirst` or `title` in templates and keep translation placeholders ambiguous.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @property
+    def nav_subtitle(self):
+        if task_enqueued is None:
+            return _("Requires Django 6.0+")
```
codingjoe: That's an interesting solution. A `django.__VERSION__` condition might be easier to find when we remove Django 5.2 support.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+                "name": task.module_path,
+                "queue_name": task.queue_name,
+                "priority": task.priority,
+                "backend": task_result.backend,
+                "run_after": task.run_after,
+                "takes_context": task.takes_context,
+                "args": sanitize_and_sort_request_vars(task_result.args),
+                "kwargs": sanitize_and_sort_request_vars(task_result.kwargs),
+                "status": task_result.status,
+            }
```
codingjoe: Why do we need to come up with our custom data structure here? `TaskResult` holds all the important information.

codingjoe: It can also fairly easily be JSON serialized, if that is needed.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _record_task(self, sender, task_result, **kwargs):
+        task = task_result.task
+        self.tasks.append(
+            {
+                "id": task_result.id,
+                "name": task.module_path,
+                "queue_name": task.queue_name,
+                "priority": task.priority,
+                "backend": task_result.backend,
```
codingjoe: The actual backend class is probably less interesting. The alias is usually probably the imporant information.

debug_toolbar/templates/debug_toolbar/panels/tasks.html
```diff
@@ -0,0 +1,48 @@
...
+    {% blocktranslate %}
+      The Tasks panel requires Django 6.0 or later. Upgrade Django to see
+      information about tasks queued during this request.
+    {% endblocktranslate %}
+  </p>
+{% elif tasks %}
+  <table>
+    <thead>
+      <tr>
+        <th>{% translate "Name" %}</th>
```
codingjoe: Just a general thought:
```suggestion
        <th>{% translate "name"|title %}</th>
```

docs/changes.rst
```diff
@@ -2,10 +2,15 @@ Change log
...
--------
 
+
+* Added a Tasks panel that displays tasks queued during the request,
+  using Django's task framework (``django.tasks``, requires Django 6.0+).
 * Fixed the Django version check in the SQL panel test suite for Django's
   boolean parameter handling.
+* Added a Tasks panel that shows tasks queued during the request via
+  Django's built-in tasks framework (``django.tasks``, Django 6.0+). On
+  older versions of Django, the panel explains that upgrading is required.
```
codingjoe, line 9: Maybe a merge error, but there seems to be duplication.

docs/panels.rst
```diff
@@ -108,6 +108,19 @@ Signals
...
+
+.. class:: debug_toolbar.panels.tasks.TasksPanel
+
+Shows the tasks queued during the request using Django's built-in tasks
+framework (``django.tasks``).
+
+.. note::
+   This panel requires Django 6.0 or later, since ``django.tasks`` was
+   added in that release. On earlier versions of Django, the panel
+   explains that upgrading is required instead of showing task data.
```
codingjoe: Maybe a little verbose. Most people will figure out that Django <6.0 doesn't have a task framework. The error message is good, but IMHO it doesn't need to be documented.

codingjoe (commented):
Nice, I am looking forward to this one 👍

debug_toolbar/panels/tasks.py
matthiask: Yep, using `django.VERSION` is preferrable since those blocks can automatically be dropped using django-upgrade later.

matthiask:
> @tim-schilling The failures only happen on the djmain test matrix, on Python 3.13/3.14 & dj52 and dj60 (the two released Django versions) both pass. The cause is django-csp's CSPMiddleware, which still imports MiddlewareMixin from this deprecated location: MiddlewareMixin from django.utils.deprecation. Django's dev branch just started emitting that warning, and since the test runner treats warnings as errors (-b flag), it fails 4 CSP tests.

Yep, that's fixed now.

debug_toolbar/panels/tasks.py
onxxdatas: I've updated the translation string to _("tasks"). I left the template unchanged since that would affect all panel titles.

debug_toolbar/panels/tasks.py
onxxdatas: updated to use django.VERSION for the compatibility check

debug_toolbar/panels/tasks.py
onxxdatas: removed the custom dictionary and now we rely on TaskResult directly.

debug_toolbar/panels/tasks.py
onxxdatas: Based on my decision, I updated the template to show backend alias from Tasks

debug_toolbar/templates/debug_toolbar/panels/tasks.html
onxxdatas: Done

onxxdatas: By the way, I lowercased the other translation strings too

debug_toolbar/panels/tasks.py
tim-schilling: We've had people want to know what the backend is for caches as well. Plus using this in different environments may mean using different backends (Immediate vs DB vs something else)

debug_toolbar/panels/tasks.py
tim-schilling:
```suggestion
        if django.VERSION < (6, 0):
            return _("Requires Django 6.0+")
```
@onxxdatas this is what they are suggesting.

debug_toolbar/panels/tasks.py
tim-schilling: @codingjoe are you suggesting we shouldn't be appending a custom dictionary to `tasks` here?

debug_toolbar/panels/tasks.py
onxxdatas: Yeah-yeah, I just pushed all my changes

debug_toolbar/panels/tasks.py
onxxdatas: Good point. I focused on the alias because it identifies the configured task backend, but I agree the actual backend implementation can be valuable when debugging differences between environments. What do you think about showing both alias and backend class at once?

docs/changes.rst
onxxdatas: Removed!

docs/panels.rst
onxxdatas: Now it just says "Requires Django 6.0+."

debug_toolbar/templates/debug_toolbar/panels/tasks.html
```diff
@@ -0,0 +1,40 @@
...
+      The Tasks panel requires Django 6.0 or later. Upgrade Django to see
+      information about tasks queued during this request.
+    {% endblocktranslate %}
+  </p>
+{% elif tasks %}
+  <table>
+    <thead>
+      <tr>
+        <th>{% translate "name"|title %}</th>
+        <th>{% translate "queue"|title %}</th>
```
codingjoe: ambiguous, could be verb to queue
```suggestion
        <th>{% translate "queue"|title context "noun" %}</th>
```

debug_toolbar/templates/debug_toolbar/panels/tasks.html
```diff
@@ -0,0 +1,40 @@
...
+    {% blocktranslate %}
+      The Tasks panel requires Django 6.0 or later. Upgrade Django to see
+      information about tasks queued during this request.
+    {% endblocktranslate %}
+  </p>
+{% elif tasks %}
+  <table>
+    <thead>
+      <tr>
+        <th>{% translate "name"|title %}</th>
```
codingjoe: ambiguous, could be verb to name

```suggestion
        <th>{% translate "name"|title context "noun" %}</th>
```

debug_toolbar/templates/debug_toolbar/panels/tasks.html
```diff
@@ -0,0 +1,40 @@
+{% load i18n %}
+{% if not tasks_available %}
```
codingjoe: I'd like to avoid negations. You can do if task .. elif task_available .. else.

That being said, the `task_available` could be replaced with a Django version check too. Again, easier to remove once Django 5.2 goes the way of the dodo.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+
+    template = "debug_toolbar/panels/tasks.html"
+
+    is_async = True
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.tasks = []
+
+    # Implement the Panel API
```
codingjoe: If it doesn't spark joy, it can go. — Marie Kondo
```suggestion
```

codingjoe (commented):
Almost! Sorry, I found just a few more small things. All nit picks, but when in rome 🤷

debug_toolbar/panels/tasks.py
onxxdatas: Removed. Didn't spark joy😄
Also fixed the "name"/"queue" ambiguity and the negation in the template, all pushed.

debug_toolbar/panels/tasks.py
codingjoe: I was suggesting to just pass the `TaskResult` instance to the template. I don't see why we need this mapping here. In other words:

```python
self.task: lists[TaskResult]
```

debug_toolbar/templates/debug_toolbar/panels/tasks.html
```diff
@@ -0,0 +1,48 @@
+{% load i18n %}
+{% if tasks %}
+  <table>
+    <thead>
+      <tr>
+        <th>{% translate "name"|title context "noun" %}</th>
```
codingjoe: Name may be ambiguous here. Path or Python-path might be more fitting.

debug_toolbar/panels/tasks.py
onxxdatas: do you have any other change request based on my last commit?

debug_toolbar/panels/tasks.py
codingjoe: I am just realizing they are one and the same. Poor naming on Django's site.

So task_result.backend is the alias.

debug_toolbar/templates/debug_toolbar/panels/tasks.html
onxxdatas: done

debug_toolbar/panels/tasks.py
codingjoe: Nope, that was the last one, promise 🤞

onxxdatas: Okay, that last one was fixed btw

codingjoe: Ok, but this one is still open, right? Or am I missing a commit somewhere? With GitHub these days you never know ;)

onxxdatas: github is githubbing :D

I forgot about this request
now, check that out, now the dict-flattening moved

codingjoe: @onxxdatas I think you misunderstood my comment. I don't understand why you don't just pass a list of `TaskResult`s to the template instead of the dictionary you are creating here.

onxxdatas: Alright, it's done as suggested now. I originally thought flattening the dictionary was a better idea because it would prevent us from having to modify any HTML templates. Thats why I was being a little too cautious. However, please check my last commit.

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+    def enable_instrumentation(self):
+        if task_enqueued is not None:
+            task_enqueued.connect(self._record_task)
+
+    def disable_instrumentation(self):
+        if task_enqueued is not None:
+            task_enqueued.disconnect(self._record_task)
+
+    def generate_stats(self, request, response):
+        # Pass TaskResult objects directly to the template
```
codingjoe, line 55:
```suggestion
```

debug_toolbar/panels/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+            return _("Requires Django 6.0+")
+        tasks = self.get_stats()["tasks"]
+        return ngettext("%(count)d task", "%(count)d tasks", len(tasks)) % {
+            "count": len(tasks)
+        }
+
+    title = _("Tasks")
+
+    def _record_task(self, sender, task_result, **kwargs):
+        # Store the TaskResult object directly
```
codingjoe:
```suggestion
```

codingjoe (approved):
Wonderful! Very lean and clean.

debug_toolbar/templates/debug_toolbar/panels/tasks.html
tim-schilling: @onxxdatas The `context "noun"` here is a helper for people doing translations, when they translate the word they know which version of "name" they should translate. In this case codingjoe is pointing out that name is the noun rather than the verb.

tim-schilling: Ugh, that's what I get for starting on the files changed rather than the issue first. Ignore me!

debug_toolbar/panels/tasks.py
tim-schilling: @onxxdatas @codingjoe I'm not sure about using `TaskResult` here, though we can make it work. Right now the tests are broken because the assertion on `stats["tasks"][0]` is comparing it to a dictionary. We'll need to change that to be a `TaskResult` instance. I'll push up a commit shortly that does that. With @onxxdatas original implementation of a dictionary, we were able to assume the template would render things out properly if the dict was realtively equal. By accessing the `TaskResult` attributes in the template, we're adding logic to the template that needs an explicit test.

I've fixed the test and added the template test in ddcccb979ade60328c68684bf172a7763d24d0d7. I'm good merging this as is, but wanted to raise this concern in case it causes people to think differently about the approach.

tim-schilling: Okay, now the tests are fixed :melting_face:

onxxdatas: > Okay, now the tests are fixed 🫠

I appreciate you taking the time to make this work.
Thanks for the review and patience on this PR

onxxdatas: > but wanted to raise this concern in case it causes people to think differently about the approach.

That's a good point about the template logic needing tests. I appreciate you fixing the tests and I'll keep this in mind for future contributions👍🏼

tim-schilling (approved):
I'm good with this as is. I'll give @codingjoe a few days to respond to the other thread if we want to change things, otherwise I'll merge it this weekend. Thank you for all your work on this @onxxdatas 🥳

codingjoe:
> I'm good with this as is. I'll give @codingjoe a few days to respond to the other thread if we want to change things, otherwise I'll merge it this weekend. Thank you for all your work on this @onxxdatas 🥳

I have no strong opinions about the context. Whatever gets this out of the door :)

tim-schilling:
Thank you @onxxdatas and @codingjoe for your efforts on this!

PR #2418, opened 2026-07-15
Improve cache hit/miss reporting for `cache.get()`

### LINKED ISSUE #2417: Cache panel incorrectly reports hits and misses for cache.get() when a default value is supplied or None is cached
The Cache panel currently determines whether a `cache.get()` operation is a hit or a miss solely by checking whether the returned value is `None`.

This leads to incorrect hit/miss reporting in the following cases:

### 1. Missing key with a default value

```python
cache.get("missing", "default")
```

The cache backend does not contain the key, so this is a **cache miss**. However, the Cache panel reports it as a **cache hit** because the returned value is the supplied default (`"default"`), which is not `None`.

### 2. Cached value is `None`

```python
cache.set("key", None)
cache.get("key")
```

The key exists in the cache, so this is a **cache hit**. However, the Cache panel reports it as a **cache miss** because the cached value is `None`.

## Expected behavior

The Cache panel should distinguish between:

* a missing cache key, and
* a cache key whose stored value is `None` (or any value equal to the caller's supplied default).

## Background

Django's cache framework recommends using a unique sentinel object when calling `cache.get()` to distinguish between a missing key and a stored value of `None`, see the highlighted guidance [here](https://docs.djangoproject.com/en/6.0/topics/cache/#:~:text=If%20you%20need%20to%20determine%20whether%20the%20object%20exists%20in%20the%20cache%20and%20you%20have%20stored%20a%20literal%20value%20None%2C%20use%20a%20sentinel%20object%20as%20the%20default%3A). The current implementation does not account for this behavior, which results in incorrect cache hit/miss statistics for the scenarios described above.

Harsh3006:
#### Description

This change fixes incorrect cache hit/miss reporting in the Cache panel for `cache.get()`.

Previously, the Cache panel determined whether a `cache.get()` call was a hit or a miss solely based on whether the returned value was `None`. This resulted in two incorrect scenarios:

A missing key with a supplied default (for example, `cache.get("missing", "default")`) was reported as a cache hit, even though the value was not retrieved from the cache.
A cached value of `None` was reported as a cache miss, even though the key existed in the cache.

The Cache panel now correctly distinguishes between a missing cache key and a cached value equal to the caller's supplied default while preserving the existing `cache.get()` behavior.

Fixes #2417 

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -65,6 +65,7 @@ class CachePanel(Panel):
     is_async = True
 
     _context_locals = Local()
+    _missing_key = object()
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -116,7 +117,12 @@ class CachePanel(Panel):
         template_info,
         backend,
     ):
-        if name == "get" or name == "get_or_set":
+        if name == "get":
+            if return_value is self._missing_key:
+                self.misses += 1
+            else:
+                self.hits += 1
+        elif name == "get_or_set":
             if return_value is None:
                 self.misses += 1
             else:
@@ -149,9 +155,22 @@ class CachePanel(Panel):
         # monkey-patched cache methods to skip recording additional calls made during
         # the course of this call, and then reset it back afterward.
         cache._djdt_panel = None
+        user_default = None
         try:
             start_time = perf_counter()
-            value = original_method(*args, **kwargs)
+            if name == "get":
+                user_default = kwargs.get("default", args[1] if len(args) > 1 else None)
+                # Replace the caller's default with an internal sentinel so a cache miss
+                # can be distinguished from a cached value equal to the supplied default.
+                if "default" in kwargs:
+                    call_args = args
+                    call_kwargs = {**kwargs, "default": self._missing_key}
+                else:
+                    call_args = (args[0], self._missing_key, *args[2:])
+                    call_kwargs = kwargs
+                value = original_method(*call_args, **call_kwargs)
+            else:
+                value = original_method(*args, **kwargs)
             t = perf_counter() - start_time
         finally:
             cache._djdt_panel = self
@@ -166,6 +185,10 @@ class CachePanel(Panel):
             template_info=get_template_info(),
             backend=f"{alias} ({type(cache).__name__})",
         )
+        # Preserve the original cache.get() behavior by returning the caller's
+        # default instead of the internal sentinel on a cache miss.
+        if name == "get" and value is self._missing_key:
+            return user_default
         return value
 
     # Implement the Panel API
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -149,7 +149,7 @@ class CachePanel(Panel):
 
     def _record_call(self, cache, alias, name, original_method, args, kwargs):
         # Some cache backends implement certain cache methods in terms of other cache
-        # methods (e.g. get_or_set() in terms of get() and add()).  In order to only
+        # methods (e.g. get_or_set() in terms of get() and add()). In order to only
         # record the calls made directly by the user code, set the cache's _djdt_panel
         # attribute to None before invoking the original method, which will cause the
         # monkey-patched cache methods to skip recording additional calls made during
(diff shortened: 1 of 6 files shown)


debug_toolbar/panels/cache.py
```diff
@@ -149,9 +155,22 @@ def _record_call(self, cache, alias, name, original_method, args, kwargs):
...
+            if name == "get":
+                user_default = kwargs.get("default", args[1] if len(args) > 1 else None)
+                # Replace the caller's default with an internal sentinel so a cache miss
+                # can be distinguished from a cached value equal to the supplied default.
+                if "default" in kwargs:
+                    call_args = args
+                    call_kwargs = {**kwargs, "default": self._missing_key}
+                else:
+                    call_args = (args[0], self._missing_key, *args[2:])
+                    call_kwargs = kwargs
```
mgstabrani: `args[0]` raises `IndexError` when the key is passed as a keyword, e.g. `cache.get(key="foo")` or `cache.get(key="foo", version=1)`. In those calls args is empty. Suggesting guarding the positional path, e.g.:

```suggestion
                if "default" in kwargs:
                    call_args = args
                    call_kwargs = {**kwargs, "default": self._missing_key}
                elif args:
                    call_args = (args[0], self._missing_key, *args[2:])
                    call_kwargs = kwargs
                else:  # key passed as keyword
                    call_args = args
                    call_kwargs = {**kwargs, "default": self._missing_key}
```

tests/panels/test_cache.py
```diff
@@ -45,6 +45,37 @@ def test_hits_and_misses(self):
...
+
+    def test_cached_value_equal_to_default_counts_as_hit(self):
+        cache.cache.clear()
+
+        cache.cache.set("foo", "bar")
+        self.assertEqual(cache.cache.get("foo", "bar"), "bar")
+
+        self.assertEqual(self.panel.hits, 1)
+        self.assertEqual(self.panel.misses, 0)
+
```
mgstabrani: CI flags 3 new uncovered statements in `cache.py` (lines 127, 166–167). The tests only cover the positional default. There's no case for the `default= `keyword path or for passing version. Adding these would close the coverage gap and also exercise the keyword-key edge case noted in `cache.py`


```suggestion
  def test_missing_key_with_default_kwarg_counts_as_miss(self):
      cache.cache.clear()
      self.assertEqual(cache.cache.get("foo", default="bar"), "bar")
      self.assertEqual(self.panel.hits, 0)
      self.assertEqual(self.panel.misses, 1)
```

debug_toolbar/panels/cache.py
Harsh3006: Good catch, thanks! I've updated the implementation to handle keyword-only key arguments by injecting the sentinel via kwargs when no positional arguments are present.

tests/panels/test_cache.py
Harsh3006: Thanks! I've added regression tests covering the keyword default path, keyword key with version, and keyword key with default, which exercise the new branches and should close the coverage gap.

Harsh3006:
@mgstabrani I noticed that CI was still flagging the `self.misses += 1` branch for `get_or_set()` as uncovered. I've added a test for the `None` return path in a follow-up commit to close the remaining coverage gap.

No `get_or_set()` logic was changed; this only adds the missing test coverage.

tim-schilling (changes requested):
I appreciate the time and effort @Harsh3006, but I think this implementation isn't something we can adopt. I'm open to alternative solutions or ideas, but this as is, I'm not onboard with right now.

This changes what the user is doing and could have downstream effects in people's applications. For example, if someone has another observability tool implemented underneath the toolbar that counts the unique default values passed in, we would break that functionality. And it would break silently while being extremely difficult to debug. They would see some random memory address being passed as the default which would be hard to search on. They'd have to audit all the packages that do something related to caching to track it down to us, and when they do, I suspect they'll be a little upset 😅 

Shorter version: The toolbar shouldn't be making changes to function calls without permission from the user.

Harsh3006:
Thanks @tim-schilling for the explanation. That makes sense. I hadn't considered the effect on downstream instrumentation that may observe the original call arguments.

I spent some more time looking for an alternative that could solve this without modifying the actual `cache.get()` call, but I couldn't find a fully reliable solution. It seems this is a limitation of the current API: once `cache.get()` returns, there isn't enough information to determine whether the returned value came from the cache or from the supplied default.

One alternative I considered was comparing the returned value with the supplied default instead of checking for `None`. This would correctly identify a miss when the key is missing and a non-None default is provided. However, it only shits the ambiguity rather than eliminating it: it would incorrectly report a miss when the cached value happens to be equal to the supplied default.

I also cross-checked this approach against the current test suite, and as expected, these two tests fail:
- `test_cached_none_counts_as_hit`
- `test_cached_value_equal_to_default_counts_as_hit`

At this point, the sentinel approach seems to be the only approach I've found that can distinguish these cases reliably, but I understand the concern about modifying the arguments passed to the wrapped function.

Harsh3006:
@tim-schilling Here are the screenshots showing the template and documentation changes, including a note on the remaining ambiguity and a link to the open discussion.

Updated Cache Panel:

image

image

Updated Documentation:
image

tim-schilling:
@Harsh3006 I consolidated the new test methods into three new methods that are specific to the new functionality added in. I felt like the new test methods were typical of AI testing and while adequate, didn't communicate why they needed to exist. This was pretty nitpicky so I made the changes on my own. If you would have preferred that I asked you to make them, let me know and I'll try to that differently in the future.

Harsh3006:
Thanks @tim-schilling for merging the PR. I appreciate you making the changes directly, and I’m completely fine with that. I’ve gone through the consolidated tests, and I agree that they communicate the intent much better. The feedback is helpful, and I’ll keep this approach in mind for future contributions.

PR #2422, opened 2026-07-25
Fix show_toolbar_with_docker on runtimes with unrelated host address

### LINKED ISSUE #2419: show_toolbar_with_docker doesn't work on orbstack
hi!
On [orbstack](https://orbstack.dev/), `docker.host.local` resolves to  0.250.250.254 (https://github.com/orgs/orbstack/discussions/1829)
This means that the check in [middleware.show_toolbar_with_docker](https://github.com/django-commons/django-debug-toolbar/blob/9e9ebf974765308de1550002192782a50e64e969/debug_toolbar/middleware.py#L44) doesn't successfully compute the host IP address and the toolbar doesn't show.

While the code does call out "This is not guaranteed to work.", it might be nice for it to try both approaches, especially given that this is a custom "only for use in docker" callback anyway.
I'm successfully using the following in my callback. Because we can't get a "real" IP for the host I just guess the gateway based on the container IP. In my testing with 20 different compose stacks on my computer every stack gets its own /24, so this assumption holds. 

```py
      container_ips = socket.gethostbyname_ex(socket.gethostname())[2]
      gateways = {ip.rsplit(".", 1)[0] + ".1" for ip in container_ips}
      return request.META.get("REMOTE_ADDR") in gateways
```

Sanjays2402:
#### Description

`show_toolbar_with_docker` derives the Docker host address by resolving
`host.docker.internal` and replacing the last segment with `.1`. On OrbStack
that name resolves to `0.250.250.254`, which is outside the container network,
so the derived address never matches `REMOTE_ADDR` and the toolbar stays
hidden. When that probe does not match, the host address is now derived from
the container's own addresses instead.

Fixes #2419

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -66,6 +66,10 @@ def show_toolbar_with_docker(request: HttpRequest) -> bool:
     # to an address that is unrelated to the container network, so the check
     # above cannot derive the host address from it.
     try:
+        # Make a reasonable guess at the gateway from the container's own
+        # interface. This assumes the process is running inside a container.
+        # If this were used outside a container this would grant access to
+        # anything that appears to come from the network's gateway (router).
         container_ips = socket.gethostbyname_ex(socket.gethostname())[2]
         gateways = {ip.rsplit(".", 1)[0] + ".1" for ip in container_ips}
         if request.META.get("REMOTE_ADDR") in gateways:
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -48,11 +48,9 @@ def show_toolbar_with_docker(request: HttpRequest) -> bool:
     if not settings.DEBUG:
         return False
 
-    # Test: settings
     if request.META.get("REMOTE_ADDR") in settings.INTERNAL_IPS:
         return True
 
-    # Test: Docker
     try:
         # This is a hack for docker installations. It attempts to look
         # up the IP address of the docker host.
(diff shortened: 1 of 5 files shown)


tim-schilling:
I tweaked the solution to be more amenable to follow-up docker tests. If this new check fails, it shouldn't automatically return false. It ideally should fall through to the last return false statement.

tim-schilling (approved):

tim-schilling (changes requested):
Can you round out the tests to have coverage of the missing lines identified here: https://github.com/django-commons/django-debug-toolbar/pull/2422#issuecomment-5142783309?

tim-schilling:
Let's also update the documentation warning around docker with a note to include something like:

> When the Docker host lookup fails, the `show_toolbar_with_docker` callback guesses the host's address from the container's own network. If run outside a container that guess resolve to another machine on your network, granting it access to the toolbar's data. The `show_toolbar_with_docker` callback should only be used inside a container.

Sanjays2402:
Added that note to the Docker warning in `docs/installation.rst` (64ad00a), wording close to yours, just softened "resolve" to "can resolve":

> When the Docker host lookup fails, the `show_toolbar_with_docker` callback guesses the host's address from the container's own network. If run outside a container that guess can resolve to another machine on your network, granting it access to the toolbar's data. The `show_toolbar_with_docker` callback should only be used inside a container.

Thanks for the fall-through rework, that reads better than the nested version I had.

tim-schilling (approved):

tim-schilling:
As I went to merge it I got cold-feet that someone could use this improperly. However, the check on `DEBUG` is solid, so not super worried about that, but I was a little worried someone may copy it without understand the implications. So I added an extra comment to explain what it's doing and where the risk would be if it were misused. Thank you @Sanjays2402 for this!

alastair:
Thank you @tim-schilling and @Sanjays2402!

PR #2424, opened 2026-07-27
Make show_toolbar_with_docker work when host.docker.internal is not the gateway

### LINKED ISSUE #2419: show_toolbar_with_docker doesn't work on orbstack

lowbyteguy:
#### Description

On [OrbStack](https://orbstack.dev/), `host.docker.internal` resolves to `0.250.250.254`, which is unrelated to the container network, so the address computed by `show_toolbar_with_docker` never matches `REMOTE_ADDR` and the toolbar is not shown.

This keeps the existing `host.docker.internal` check first and adds a fallback that derives the candidate gateways from the container's own addresses, as suggested in the issue.

Fixes #2419

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/middleware.py
+++ b/debug_toolbar/middleware.py
@@ -68,6 +68,20 @@ def show_toolbar_with_docker(request: HttpRequest) -> bool:
         # It's fine if the lookup errored since they may not be using docker
         pass
 
+    # Test: Docker, without relying on host.docker.internal
+    # Some runtimes (e.g. OrbStack) either do not provide host.docker.internal
+    # or resolve it to an address that is not the gateway. Fall back to
+    # guessing the gateway from the container's own addresses.
+    try:
+        container_ips = socket.gethostbyname_ex(socket.gethostname())[2]
+    except socket.gaierror:
+        # It's fine if the lookup errored since they may not be using docker
+        pass
+    else:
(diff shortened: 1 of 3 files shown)


tim-schilling:
Closing as a duplicate of #2422

PR #2425, opened 2026-07-28
Address new ruff 0.16 and biome 2.5.5 warnings

codingjoe:
I added explicitly expected exceptions for cases where I could figure it out. In some places I omitted the error messages.

### DIFF
--- a/debug_toolbar/panels/profiling.py
+++ b/debug_toolbar/panels/profiling.py
@@ -87,7 +87,7 @@ class FunctionCall:
             )
 
     def subfuncs(self):
-        h, s, v = self.hsv
+        h, s, _v = self.hsv
         count = len(self.statobj.all_callees[self.func])
         for i, (func, stats) in enumerate(self.statobj.all_callees[self.func].items()):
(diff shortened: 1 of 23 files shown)


debug_toolbar/panels/profiling.py
```diff
@@ -109,19 +109,19 @@ def tottime(self):
         return self.stats[2]
 
     def cumtime(self):
-        cc, nc, tt, ct = self.stats
+        _cc, _nc, _tt, _ct = self.stats
```
federicobond: This looks weird. Do we need this line at all, given that we access `self.stats[3]` below?

federicobond (approved):
Looks good overall!

debug_toolbar/panels/profiling.py
codingjoe: This was a ruff unsafe autofix. I will address this better.

codingjoe: Addressed it.

tim-schilling (approved):

tim-schilling:
Good stuff folks!

PR #2433, opened 2026-08-05
Defer the toolbar handle position calculation to an animation frame

novucs:
#### Description

`ensureHandleVisibility()` reads `handle.offsetWidth` during `init()`, directly after `$$.show(handle)` takes the handle from `display:none`. Layout is dirty at that point, so the read forces a synchronous document-wide style and layout flush after the page has already painted. On projects with a large stylesheet this is slow enough to be visible as a flash of unstyled content in Firefox.

It only affects a collapsed toolbar, since `showToolbar()` does not call `ensureHandleVisibility()`.

Deferring the body to `requestAnimationFrame` moves the read past the paint. Behaviour is otherwise unchanged, and the resize listener benefits too, as repeated events now coalesce to at most one reflow per frame.

~~No test. The only observable change is timing, and `ensureHandleVisibility()` isn't reachable without scaffolding the whole toolbar DOM. Happy to add one if you'd like.~~ Now has a selenium test for this.

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/tests/views.py
+++ b/tests/views.py
@@ -1,9 +1,10 @@
 import asyncio
+import time
 
 from asgiref.sync import sync_to_async
 from django.contrib.auth.models import User
 from django.core.cache import cache
-from django.http import HttpResponseRedirect, JsonResponse
+from django.http import HttpResponse, HttpResponseRedirect, JsonResponse
 from django.shortcuts import render
 from django.template.response import TemplateResponse
 from django.views.decorators.cache import cache_page
@@ -135,3 +136,14 @@ def redirect_view(request):
 
 def ajax_view(request):
     return render(request, "ajax/ajax.html")
+
+
+def flash_of_unstyled_content_view(request):
+    return render(request, "flash_of_unstyled_content.html")
+
+
+def delayed_stylesheet_view(request):
+    # Holds the sheet pending long enough for the toolbar's async script to run first.
+    time.sleep(0.1)
+    css = "html { --delayed-stylesheet: applied; }"
+    return HttpResponse(css, content_type="text/css")
--- a/debug_toolbar/static/debug_toolbar/js/toolbar.js
+++ b/debug_toolbar/static/debug_toolbar/js/toolbar.js
@@ -247,14 +247,16 @@ const djdt = {
         }
     },
     ensureHandleVisibility() {
-        const djDebug = getDebugElement();
-        const handle = djDebug.querySelector("#djDebugToolbarHandle");
-        // set handle position
-        const handleTop = Math.min(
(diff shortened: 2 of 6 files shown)


tim-schilling:
Thank you for the PR @novucs. Would you be able to reproduce this in a tool like selenium?

novucs:
> Thank you for the PR @novucs. Would you be able to reproduce this in a tool like selenium?

@tim-schilling great idea. Here's a minimal reproducible script, how-to for running, and attached are video comparisons of the results: https://gist.github.com/novucs/9f8c31fa5661aecb3b8a940408914865?permalink_comment_id=6299585#gistcomment-6299585

[patched.webm](https://github.com/user-attachments/assets/d8a3124b-2d32-46a8-8cf1-c0dc6757db8a)
[unpatched.webm](https://github.com/user-attachments/assets/b75cebdf-15ad-47f4-a830-27a1371d2bb1)

novucs:
Re-pushed to resolve lint errors in the changelog

tim-schilling:
Thanks for sharing that! Though I should have been clear that it was a yes/no question 😅. We have selenium tests for now so we'll need to translate your gist. I'd like to switch to playwright eventually.

If you have the time, do you mind pulling that example into a selenium test in the project? No worries if you're at your limit.

novucs:
Oops... 😂 I'll see what I can do when I next get a minute! Thanks for letting me know

novucs:
@tim-schilling I've now added & pushed a small regression test for this. While there, I've found a number of failing tests due to the shadow DOM change in #2266. The Selenium tests still find elements on the document, which can't reach inside the open shadow root. ~~Happy to open a fix for these in a separate PR~~ I have pushed fixes in a separate PR: https://github.com/django-commons/django-debug-toolbar/pull/2434.

tests/views.py
```diff
@@ -135,3 +136,14 @@ def redirect_view(request):
 
 def ajax_view(request):
     return render(request, "ajax/ajax.html")
+
+
+def fouc_view(request):
```
tim-schilling: Can we change this to be more descriptive please? This is a bit too condensed for me to know what it's for. The same applies for the other places `fouc` is used. The URL path can stay like that if you prefer, but the view and url names should be descriptive.

novucs: Ah, yep no problem, I had FOUC on the brain and I guess not everyone will be immediately familiar. Expanded the contraction everywhere. I've also stripped the Claude co-author like you wanted from the other PR to save you the job. Thanks again for the review!

tim-schilling (approved):
Looks good. I'd like to give this one a little time to let someone else chime in though.

PR #2391, opened 2026-06-16
Fix binary parameter handling in SQL panel

matthiask:
#### Description

This resurrects https://github.com/django-commons/django-debug-toolbar/pull/2169 and hopefully fixes the remaining issues.

Fixes #423 

#### Checklist:

- [x] I have added the relevant tests for this change.
- [x] I have added an item to the Pending section of ``docs/changes.rst``.

#### AI/LLM Usage
- [x] This PR includes code generated with the help of an AI/LLM

PR #2434, opened 2026-08-06
Fix integration tests

novucs:
#### Description

While working on selenium tests for https://github.com/django-commons/django-debug-toolbar/pull/2433, I noticed the ones on `main` were already failing. Here's a bundle of small patches to get integration tests back up & running.

Patches:
* **[Search the toolbar's shadow root in the Selenium tests](https://github.com/django-commons/django-debug-toolbar/commit/5bfc6426fe5edae316da2242f96306670df384d4)** - The tests look elements up on the document, which cannot see into the open shadow root the toolbar has rendered into since https://github.com/django-commons/django-debug-toolbar/pull/2266, so every lookup raised `NoSuchElementException`. Fall back to searching the shadow root when the document search finds nothing, which leaves the existing call sites unchanged.
* [Assert on table textContent in the Selenium tests](https://github.com/django-commons/django-debug-toolbar/commit/345888ae038762bb2c1112a55b2c842b308966d2) - The panel stylesheet uppercases table headers and `WebElement.text` returns rendered text, so the assertions no longer matched the source strings. Read textContent instead, which is untransformed.
* [Show the SQL select and explain buttons for queries without parameters](https://github.com/django-commons/django-debug-toolbar/commit/bc5df6c649f2496c696369398b03c21c3bb40893) - The buttons are wrapped in a check on query.params, which held a JSON string until that changed to the decoded value. An empty list of parameters then read as empty and the buttons stopped rendering, so a query with no parameters lost its actions.

  Compare against None instead, which is what the panel records when the parameters cannot be serialized. This restores the behaviour the check had while the value was a string.
* [Find the toolbar window inside the shadow root when a panel errors](https://github.com/django-commons/django-debug-toolbar/commit/c42aac42c951c23d5fa1a22dcddbb846b3affc19) - The failure branch of ajax() reached for the window with document.getElementById, which cannot see into the shadow root the toolbar renders into, so the lookup returned null and the handler threw instead of reporting the error. Every other reference already goes through the toolbar element.
* [Actually run the Selenium tests in CI](https://github.com/django-commons/django-debug-toolbar/commit/ce74113d75f5873d8e9a1702e930b70997c12fcd) - The matrix entries carry env: followed by a plain scalar, so it parses
as the string `"DJANGO_SELENIUM_TESTS=true"` rather than a mapping, and `matrix.env.DJANGO_SELENIUM_TESTS` resolves to nothing. The suite has been skipping since the job was added.

  Make it a mapping. The runner image already provides Firefox and geckodriver, and the tests run headless when CI is set, so no extra setup is needed.
* [Stop the history panel buttons from submitting their forms](https://github.com/django-commons/django-debug-toolbar/pull/2434/commits/6cc4e492786585a1342ecad759635c0309ddff83) - Both buttons sit in a form and carry no type, so they default to submit. They only avoid navigating because history.js calls preventDefault, and that script loads with the panel, so a click landing first sends the browser to the raw JSON of the sidebar or refresh view. The toolbar is then gone from the page, which is what the switch test hit on CI.

  The forms are read by ajaxForm rather than submitted, so nothing needs the submit behaviour. Clicking early is now a no-op instead, so the test retries the click until the switch takes effect.

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -133,13 +133,13 @@ jobs:
             - python-version: '3.13'
               database: psycopg3
               env:
-                DJANGO_SELENIUM_TESTS=true
+                DJANGO_SELENIUM_TESTS: "true"
             - python-version: '3.14'
               database: postgis3
             - python-version: '3.14'
               database: psycopg3
               env:
-                DJANGO_SELENIUM_TESTS=true
+                DJANGO_SELENIUM_TESTS: "true"
 
     services:
       postgres:
--- a/debug_toolbar/static/debug_toolbar/js/utils.js
+++ b/debug_toolbar/static/debug_toolbar/js/utils.js
@@ -102,7 +102,7 @@ export async function ajax(url, init) {
         }
         throw new Error(`${response.status}: ${response.statusText}`);
     } catch (error) {
-        const win = document.getElementById("djDebugWindow");
+        const win = getDebugElement().querySelector("#djDebugWindow");
         win.innerHTML = `<div class="djDebugPanelTitle"><h3>${error.message}</h3><button type="button" class="djDebugClose">»</button></div>`;
         $$.show(win);
         throw error;
--- a/debug_toolbar/templates/debug_toolbar/panels/history.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/history.html
@@ -1,7 +1,7 @@
 {% load i18n static %}
 <form method="get" action="{% url 'djdt:history_refresh' %}">
     {{ refresh_form.as_div }}
-    <button class="refreshHistory">Refresh</button>
+    <button type="button" class="refreshHistory">Refresh</button>
 </form>
 <table class="djdt-max-height-100">
     <thead>
--- a/debug_toolbar/templates/debug_toolbar/panels/history_tr.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/history_tr.html
@@ -44,7 +44,7 @@
     <td class="djdt-actions">
         <form method="get" action="{% url 'djdt:history_sidebar' %}">
             {{ history_context.form.as_div }}
-            <button data-request-id="{{ request_id }}" class="switchHistory">Switch</button>
+            <button type="button" data-request-id="{{ request_id }}" class="switchHistory">Switch</button>
         </form>
     </td>
 </tr>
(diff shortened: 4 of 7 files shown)


tim-schilling:
Hmm, weird. They were failing locally but not in CI?

novucs:
> Hmm, weird. They were failing locally but not in CI?

It looks like they may not have been running in CI to begin with, unless I’m missing a trick. But yes, they were failing locally until applying these changes

novucs:
Either way unfortunately it seems GitHub actions is down for the last run of tests on this PR 😂

novucs:
Latest run of tests in CI appeared to fail: https://github.com/django-commons/django-debug-toolbar/actions/runs/31118573493/job/92875980501?pr=2434

Looks like there was a race condition making one of the tests flaky. Latest commit should resolve.

.github/workflows/test.yml
```diff
@@ -133,13 +133,13 @@ jobs:
...
               env:
-                DJANGO_SELENIUM_TESTS=true
+                DJANGO_SELENIUM_TESTS: "true"
             - python-version: '3.14'
               database: postgis3
             - python-version: '3.14'
               database: psycopg3
               env:
-                DJANGO_SELENIUM_TESTS=true
+                DJANGO_SELENIUM_TESTS: "true"
```
tim-schilling, line 142: 😮‍💨 that one hurts. Thanks for fixing it.

novucs:
Gah. Same test is causing problems (`test_rerender_on_history_switch`). Unfortunately I've not been able to replicate this behaviour locally quite yet. I'll dig in...

novucs:
Ok so I think I've got it and pushed up: Race condition between history.js loading and the tests hitting the Switch button. Before history.js gets loaded, clicking the button submits the form and redirects the page to the raw JSON endpoint. It's now a no-op before the script loads, and the test retires the button. Hopefully the flakiness will now disappear! 🤞

Note: I still haven't been able to reproduce the race condition locally, but CI seems to error somewhat consistently on this

tim-schilling (approved):
The only thing that gives me a little pause is the change to the SQL panel's template. That original truthy check is over a decade old, so it seems weird we need it to fix CI.

novucs:
> The only thing that gives me a little pause is the change to the SQL panel's template. That original truthy check is over a decade old, so it seems weird we need it to fix CI.

Thanks for the review! What the check sees changed rather than the check itself: #2391 swapped `_params` from `json.dumps(self._decode(params))` to the decoded value, so a query with no parameters went from `"[]"` (truthy) to `[]` (falsy) and lost its Sel/Expl buttons.

tim-schilling:
I'm going to strip the claude co-author from the commits then merge this.

PR #2437, opened 2026-08-08
Fix executemany() raising from the SQL panel

### LINKED ISSUE #1069: "TypeError: not enough arguments for format string" on executemany INSERT in sqlite3
Hi Folks. Thanks very much for your work. I'm getting an error on django 1.11 and django-debug-toolbar 1.9.1. I get no errors with postgres on the back end, but sqlite3 gets:

    Internal Server Error: /forecast/34/upload/286
      File "/Users/cornell/.local/share/virtualenvs/forecast-repository-aIKRoinE/lib/python3.6/site-packages/django/core/handlers/exception.py", line 41, in inner
      File "/Users/cornell/.local/share/virtualenvs/forecast-repository-aIKRoinE/lib/python3.6/site-packages/django/core/handlers/base.py", line 187, in _get_response
	response = self.process_exception_by_middleware(e, request)
      File "/Users/cornell/.local/share/virtualenvs/forecast-repository-aIKRoinE/lib/python3.6/site-packages/django/core/handlers/base.py", line 185, in _get_response
      File "/Users/cornell/IdeaProjects/forecast-repository/forecast_app/views.py", line 904, in upload_forecast
	forecast_model.load_forecast(Path(tmp_data_file), time_zero, file_name)
      File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/contextlib.py", line 53, in inner
	return func(*args, **kwds)
      File "/Users/cornell/IdeaProjects/forecast-repository/forecast_app/models/forecast_model.py", line 85, in load_forecast
	new_forecast.load_csv_data(csv_file_path)
      File "/Users/cornell/IdeaProjects/forecast-repository/forecast_app/models/data.py", line 48, in load_csv_data
	cursor.executemany(sql, rows)
      File "/Users/cornell/.local/share/virtualenvs/forecast-repository-aIKRoinE/lib/python3.6/site-packages/debug_toolbar/panels/sql/tracking.py", line 167, in executemany
	return self._record(self.cursor.executemany, sql, param_list)
      File "/Users/cornell/.local/share/virtualenvs/forecast-repository-aIKRoinE/lib/python3.6/site-packages/debug_toolbar/panels/sql/tracking.py", line 130, in _record
	self.cursor, sql, self._quote_params(params)),
      File "/Users/cornell/.local/share/virtualenvs/forecast-repository-aIKRoinE/lib/python3.6/site-packages/django/db/backends/sqlite3/operations.py", line 142, in last_executed_query
	return sql % params
    TypeError: not enough arguments for format string

Here's is the minimal view function that recreates the problem:

    def load_csv_data(self, cdc_csv_file):
        with connection.cursor() as cursor:
            rows = [('US National', 34, 'point', 'week', None, None, 50, 4421)]
            sql = """
                INSERT INTO forecast_app_forecastdata (location, target_id, row_type, unit, bin_start_incl, bin_end_notincl, value, forecast_id)
                VALUES (%s, %s, %s, %s, %s, %s, %s, %s);
            """
            cursor.executemany(sql, rows)

The error is in:  https://github.com/jazzband/django-debug-toolbar/blob/ac52b23f9a05cafe0f7124a1f9d7b8236a3027d5/debug_toolbar/panels/sql/tracking.py#L132

The variables at that point are, via `print(self.cursor, repr(sql), repr(params), repr(self._quote_params(params)))`:

    <django.db.backends.utils.CursorDebugWrapper object at 0x107e6fa90>

     '\n                INSERT INTO forecast_app_forecastdata (location, target_id, row_type, unit, bin_start_incl, bin_end_notincl, value, forecast_id)\n                VALUES (%s, %s, %s, %s, %s, %s, %s, %s);\n            '

    [('US National', 34, 'point', 'week', None, None, 50, 4421)]

    ["('US National', 34, 'point', 'week', None, None, 50, 4421)"]

Finally, the failing `last_executed_query()` line in django/db/backends/sqlite3/operations.py is:

    return sql % params

The variables at that point are, via `print(repr(sql), repr(params))`:

    '\n                INSERT INTO forecast_app_forecastdata (location, target_id, row_type, unit, bin_start_incl, bin_end_notincl, value, forecast_id)\n                VALUES (%s, %s, %s, %s, %s, %s, %s, %s);\n            '

    ("'(''US National'', 34, ''point'', ''week'', None, None, 50, 4421)'",)

Sure enough, that fails.

Is there anything else I can share about this? In the meantime I've updated my django settings to disable the toolbar for sqlite3. Thank you.

alliasgher:
#### Description

With the SQL panel instrumented, `cursor.executemany(sql, param_list)` raises instead of being recorded.

`executemany()` hands the *list of param sequences* to the same `_record()` that `execute()` uses, and `_record()`'s `finally:` block calls `self._last_executed_query(sql, params)`. The backend's `last_executed_query()` expects a flat sequence of scalars — sqlite re-quotes them with `cursor.execute("SELECT " + ", ".join(["QUOTE(?)"] * len(params)), params)` — so binding a tuple fails:

```
sqlite3.ProgrammingError: Error binding parameter 1: type 'tuple' is not supported
```

#### Approach

`django/db/backends/utils.py::CursorDebugWrapper.executemany` passes `many=True`, and `debug_sql` then logs `"%s times: %s" % (times, sql)` instead of calling `last_executed_query`. The toolbar's wrapper had no `many` concept at all, so this threads one through `_record()` and does the same.

I also hid the Sel/Expl/Prof buttons for these queries. `SQLSelectForm.select`/`explain`/`profile` re-run `query["raw_sql"]` with `query["params"]` through a plain `cursor.execute`, which fails the same way — so clicking them on an executemany row would just reproduce the original error. **I realise #2393 deliberately made those buttons unconditional; this is a narrower exclusion for the one case where the params are structurally not re-executable, not a rollback of that.**

image

#### Scope

This affects code calling `cursor.executemany()` on raw SQL.

Fixes #1069

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -187,7 +187,7 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
         finally:
             self.db._djdt_logger = self.logger
 
-    def _record(self, method, sql, params):
+    def _record(self, method, sql, params, *, many=False):
         alias = self.db.alias
         vendor = self.db.vendor
 
@@ -219,13 +219,27 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
             else:
                 sql = str(sql)
 
+            if many:
+                # params is a sequence of param sequences, but
+                # last_executed_query() expects a flat sequence of scalars, so
+                # it cannot interpolate this. Report the statement and how many
+                # times it ran, matching Django's own CursorDebugWrapper.
+                try:
+                    times = len(params)
+                except TypeError:
+                    times = "?"
+                display_sql = f"{times} times: {sql}"
+            else:
+                display_sql = self._last_executed_query(sql, params)
+
             kwargs = {
                 "vendor": vendor,
                 "alias": alias,
-                "sql": self._last_executed_query(sql, params),
+                "sql": display_sql,
                 "duration": duration,
                 "raw_sql": sql,
                 "params": _params,
+                "many": many,
                 "stacktrace": get_stack_trace(skip=2),
                 "template_info": template_info,
             }
@@ -281,4 +295,4 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
         return self._record(super().execute, sql, params)
 
     def executemany(self, sql, param_list):
-        return self._record(super().executemany, sql, param_list)
+        return self._record(super().executemany, sql, param_list, many=True)
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -7,6 +7,7 @@ from time import perf_counter
 import django.test.testcases
 from django.apps import apps
 from django.core.exceptions import ImproperlyConfigured
+from django.utils.translation import gettext as _, ngettext
 
 from debug_toolbar import settings as dt_settings
 from debug_toolbar.sanitize import force_str
@@ -227,8 +228,13 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
                 try:
                     times = len(params)
                 except TypeError:
-                    times = "?"
-                display_sql = f"{times} times: {sql}"
+                    display_sql = _("? times: %(sql)s") % {"sql": sql}
+                else:
+                    display_sql = ngettext(
+                        "%(count)d time: %(sql)s",
+                        "%(count)d times: %(sql)s",
+                        times,
+                    ) % {"count": times, "sql": sql}
             else:
                 display_sql = self._last_executed_query(sql, params)
 
--- a/debug_toolbar/panels/sql/tracking.py
+++ b/debug_toolbar/panels/sql/tracking.py
@@ -7,7 +7,6 @@ from time import perf_counter
 import django.test.testcases
 from django.apps import apps
 from django.core.exceptions import ImproperlyConfigured
-from django.utils.translation import gettext as _, ngettext
 
 from debug_toolbar import settings as dt_settings
 from debug_toolbar.sanitize import force_str
@@ -223,20 +222,16 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
             if many:
                 # params is a sequence of param sequences, but
                 # last_executed_query() expects a flat sequence of scalars, so
-                # it cannot interpolate this. Report the statement and how many
-                # times it ran, matching Django's own CursorDebugWrapper.
+                # it cannot interpolate this. Record the statement as it was
+                # given, and how many times it ran.
+                display_sql = sql
                 try:
-                    times = len(params)
+                    count = len(params)
                 except TypeError:
-                    display_sql = _("? times: %(sql)s") % {"sql": sql}
-                else:
-                    display_sql = ngettext(
-                        "%(count)d time: %(sql)s",
-                        "%(count)d times: %(sql)s",
-                        times,
-                    ) % {"count": times, "sql": sql}
+                    count = None
             else:
                 display_sql = self._last_executed_query(sql, params)
+                count = None
 
             kwargs = {
                 "vendor": vendor,
@@ -245,7 +240,7 @@ class NormalCursorMixin(DjDTCursorWrapperMixin):
                 "duration": duration,
                 "raw_sql": sql,
                 "params": _params,
-                "many": many,
+                "count": count,
                 "stacktrace": get_stack_trace(skip=2),
                 "template_info": template_info,
             }
--- a/debug_toolbar/templates/debug_toolbar/panels/sql.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/sql.html
@@ -46,6 +46,11 @@
           </td>
           <td>
             <div class="djDebugSql">{{ query.sql|safe }}</div>
+            {% if query.count is not None %}
+              <strong>
+                {% blocktranslate count counter=query.count %}Executed {{ counter }} time.{% plural %}Executed {{ counter }} times.{% endblocktranslate %}
+              </strong>
+            {% endif %}
             {% if query.similar_count %}
               <strong>
                 <span class="djdt-color" data-djdt-styles="backgroundColor:{{ query.similar_color }}"></span>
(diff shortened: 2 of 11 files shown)


tim-schilling:
@alliasgher please be more concise when using LLMs to generate messages. They are overly verbose and some of it is unhelpful.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -219,13 +219,27 @@ def _record(self, method, sql, params):
...
+            if many:
+                # params is a sequence of param sequences, but
+                # last_executed_query() expects a flat sequence of scalars, so
+                # it cannot interpolate this. Report the statement and how many
+                # times it ran, matching Django's own CursorDebugWrapper.
+                try:
+                    times = len(params)
+                except TypeError:
+                    times = "?"
+                display_sql = f"{times} times: {sql}"
```
matthiask: I think this should probably be internationalized.

docs/changes.rst
```diff
@@ -16,6 +16,9 @@ Pending
 * Stopped the history panel buttons from submitting their form when clicked
   before the panel script has loaded, which navigated away from the page.
 * Added support for Django 6.1.
+* Fixed ``cursor.executemany()`` raising from the SQL panel instead of being
+  recorded. The statement is now shown as ``N times: <sql>``, matching
+  Django's own debug cursor.
```
matthiask: This has to be moved up to the section covering the next release now.

tests/panels/test_sql.py
```diff
@@ -104,6 +104,51 @@ def test_recording(self):
...
+        The backend's last_executed_query() expects a flat sequence of scalar
+        params, so handing it a list of param sequences fails. It happens in a
+        finally: block after the write has already landed, so the exception
+        escapes into the caller.
+        """
+        self.assertEqual(len(self.panel._queries), 0)
+
+        with connection.cursor() as cursor:
+            cursor.executemany(
+                "INSERT INTO tests_binary (field) VALUES (%s)",
```
matthiask, line 120: Just asking: Is there a specific reason you chose `tests_binary` for this?

matthiask (changes requested):

debug_toolbar/panels/sql/tracking.py
alliasgher: Done in `aead842d`, using `ngettext` so the count pluralizes, with a separate string for the `?` case when `params` has no `len()`. Added a singular test.

docs/changes.rst
alliasgher: Moved to Pending in `aead842d`. I rebased on main and it had landed inside the released 7.1.0 section.

tests/panels/test_sql.py
alliasgher: It is the only backend-agnostic model in `tests/models.py`. The other two are Postgres-only (`PostgresJSON`) and GeoDjango. `Binary` also has a single field, so the INSERT stays one column. Happy to switch if you prefer another.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -219,13 +220,32 @@ def _record(self, method, sql, params):
...
+                except TypeError:
+                    display_sql = _("? times: %(sql)s") % {"sql": sql}
+                else:
+                    display_sql = ngettext(
+                        "%(count)d time: %(sql)s",
+                        "%(count)d times: %(sql)s",
+                        times,
+                    ) % {"count": times, "sql": sql}
+            else:
+                display_sql = self._last_executed_query(sql, params)
```
tim-schilling, line 233: Something about this feels funky to me. I don't like that we're adding a "5 times: <sql>" prefix to the value that's meant to just be the SQL. I see the count as a characteristic for the SQL, not as a part of it. Similar to how the duration has a separate value. Can we remove the `count` aspects to this and move it into the `kwargs` below in place of the `many`? I think that may be a bit cleaner.

tim-schilling (changes requested):

debug_toolbar/panels/sql/tracking.py
alliasgher: Agreed, that reads better. Done in `8bf5c945`: `sql` is just the statement now, and the repeat count is a `count` kwarg alongside `duration`. The panel renders it under the query, and the Sel/Expl/Prof buttons key off `count is None` instead of the `many` flag, which also handles the `count == 0` case correctly.

debug_toolbar/panels/sql/tracking.py
```diff
@@ -219,13 +219,28 @@ def _record(self, method, sql, params):
...
+
             kwargs = {
                 "vendor": vendor,
                 "alias": alias,
-                "sql": self._last_executed_query(sql, params),
+                "sql": display_sql,
                 "duration": duration,
                 "raw_sql": sql,
                 "params": _params,
+                "count": count,
```
tim-schilling, line 243: Let's not be stingy with characters here 😁 Let's make this obvious what it's a count of. There are too many data structures for us to make an assumption.

debug_toolbar/templates/debug_toolbar/panels/sql.html
```diff
@@ -46,6 +46,11 @@
           </td>
           <td>
             <div class="djDebugSql">{{ query.sql|safe }}</div>
+            {% if query.count is not None %}
```
tim-schilling, line 49: Do we need to show this for the zero and one cases?

tim-schilling:
As a heads up, I revised the PR description to be more concise. I felt that the information that was removed either belonged in the issue report or was accessible by reading the code and CI test results. Unfortunately, LLMs are overly verbose and their default output is way too much to share without editing IMO.

I also added a screenshot of what this looks like.

tim-schilling:
@alliasgher so I pulled this down to play around with it and discovered that it seems to be possible to show the full SQL for the `executemany` flow. See what I propose we do instead in 1852db623f2e4a2bd9beb2f841d4dcd877b5e094. I was 70% of the way there and decided I may as well clean-up the rest of it for you to review.

Screenshot from 2026-09-10 18-36-42
 

I was able to get the Select button to work and I think we could even get Explain to work as well. However, I think those are out of scope for this particular issue and could be follow-up work.

PR #2438, opened 2026-08-08
Improve selenium tests reliability in CI

tim-schilling:
#### Description

This replaces the default wait time and makes the test a little more reliable. Update selenium test approach when waiting for panel switches.

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -1,6 +1,5 @@
 import os
 import re
-import time
 import unittest
 import warnings
 from unittest.mock import patch
@@ -37,7 +36,10 @@ from .views import regular_view
 
 try:
     from selenium import webdriver
-    from selenium.common.exceptions import NoSuchElementException
+    from selenium.common.exceptions import (
+        NoSuchElementException,
+        StaleElementReferenceException,
+    )
     from selenium.webdriver.common.by import By
     from selenium.webdriver.firefox.options import Options
     from selenium.webdriver.support import expected_conditions as EC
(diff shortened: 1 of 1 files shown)


tim-schilling:
@novucs do you mind taking a look at this as well?

tests/test_integration.py
```diff
@@ -804,29 +837,21 @@ def test_rerender_on_history_switch(self):
...
-            return (
-                selenium.find_element(By.ID, "djDebug").get_attribute("data-request-id")
-                != previous_request_id
-            )
-
-        self.wait.until(switch_to_oldest_snapshot)
+        # Click to switch back to the jinja page view snapshot. The rows are
+        # re-populated once the panel has rendered, so wait until one is there
+        # to click.
+        self.wait.until(lambda selenium: self.click_last(".switchHistory"))
```
novucs: It's possible we'll end up clicking `.switchHistory` before the click event for this button is registered with `history.js`. I'd popped the click & check in the `wait.until` loop so that we'll click the button again when we do not see a change. That being said, it's entirely possible selenium attempts to click the button again while it is registered and we're still awaiting a response, but we should still get the same response regardless so I wasn't too worried about that side-effect.

This change here now reverts it so selenium only clicks the button once, regardless of whether or not it was registered with `history.js`. We likely could end back up with this being a flaky test.

tests/test_integration.py
```diff
@@ -986,20 +1011,13 @@ def test_ajax_dont_refresh(self):
...
-                self.selenium.find_element(
-                    By.CSS_SELECTOR, "#djdt-HistoryPanel a.HistoryPanel small"
-                ).text
-                == "/json_view/"
-            )
-        )
+        # The toolbar debounces the ajax response before fetching the new state,
+        # then swaps in every panel synchronously, so a changed request id means
+        # the history panel has finished being replaced.
+        self.wait.until(lambda selenium: self.get_request_id() != request_id)
```
novucs, line 1010: This change looks good to me, should avoid the flake seen in: https://github.com/django-commons/django-debug-toolbar/actions/runs/31280101857/job/93159913439 👍

tests/test_integration.py
tim-schilling: My understanding is that `click_last` would be retried until it succeeds or the `wait_until` timeout is exceeded. It shouldn't only be called once.

tim-schilling: Ah, that's not your point though.

tim-schilling: You're right, let me revert this and incorporate the stale reference handler check. Thank you for the review!

PR #2444, opened 2026-08-13
Serialize the TaskResult class for Tasks panel.

### LINKED ISSUE #2443: Tasks panel doesn't appear to be populated in example app
When running the example app and using the tasks page, it appears the tasks aren't being rendered properly. It's being captured, but not rendered out properly. It appears that the serialization of the `TaskResult` may be transforming it entirely into a string rather than serializing the object.

Image

tim-schilling:
#### Description

This serializes the Tasks panel's content so it can be stored and fetched. It also resolves the issue of extra whitespace on the `kwargs` column of the tasks panel.

Fixes #2443

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/panels/tasks.py
+++ b/debug_toolbar/panels/tasks.py
@@ -52,10 +52,26 @@ class TasksPanel(Panel):
             task_enqueued.disconnect(self._record_task)
 
     def generate_stats(self, request, response):
-        # Pass TaskResult objects directly to the template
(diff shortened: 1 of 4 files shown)


matthiask (approved):
Thanks! LGTM.

codingjoe (approved):
Ah, dang, that was changed on my suggestion. I wasn't aware of that.

BTW: Maybe add a comment for a future chore once this makes it into Django: https://github.com/django/new-features/issues/182

PR #2429, opened 2026-08-01
Update the design interface of the toolbar

VeldaKiara:
#### Description

This PR will update the following:
- The newly designed logo
- Design documentation and guidelines for the documentation
- Toolbar UI colours with the new theme
- Credit to the designer for the work on docs
- Small resource change for a video I did 

Fixes #2428 

#### Checklist:

- [X] I have added the relevant tests for this change.
- [X] I have added an item to the Pending section of ``docs/changes.rst``.

#### AI/LLM Usage
- [X] The accessibility tests on the newly implemented design were run and tested with the help of an AI/LLM for compliance with axe-core 4.10.2 (WCAG 2.0/2.1 A + AA rulesets).

PR #2446, opened 2026-08-14
Updated example screenshot of the toolbar with the new design.

### LINKED ISSUE #2440: Update README screenshot with tasks panel inclusion
There should be a `make` command that will generate the screenshot.

tim-schilling:
#### Description

Updates the screenshot to reflect the design changes of #2429. Also fixes the command to capture screenshots which broke when we introduced the shadow DOM in #2266.

Note that this increases the screenshot image size because the new design is a bit "bigger" and I wanted to continue to show more of the toolbar. It's not perfect, but let me know if it should be better.

Fixes #2440

#### Checklist:

#### AI/LLM Usage


matthiask:
Thanks, this looks much better!

PR #2451, opened 2026-08-18
fix: distinguish cache miss from stored None in Cache panel

### LINKED ISSUE #2417: Cache panel incorrectly reports hits and misses for cache.get() when a default value is supplied or None is cached

aryansk:
## Summary

Fixes #2417

The Cache panel previously determined hit/miss by checking whether the return value was `None`, which caused two classes of incorrect reporting:

1. `cache.get("missing", "default")` was counted as a hit (the default is not None)
2. `cache.set("key", None); cache.get("key")` was counted as a miss (the stored value is None)

## Fix

Uses a sentinel object as the default when wrapping `get`/`get_or_set`, following the pattern recommended in Django's cache docs. The sentinel is substituted before calling the original method and swapped back afterward, so callers always see the expected value. The panel now checks for the sentinel to correctly count misses.

## Testing

Verified both scenarios:
- `cache.get("missing", "default")` → now correctly counted as miss
- `cache.set("key", None); cache.get("key")` → now correctly counted as hit

🤖 Generated with Codebuff
Co-Authored-By: Codebuff <noreply@codebuff.com>

### DIFF
--- a/debug_toolbar/panels/cache.py
+++ b/debug_toolbar/panels/cache.py
@@ -30,6 +30,13 @@ WRAPPED_CACHE_METHODS = [
 ]
 
 
+# Sentinel to distinguish a cache miss from a stored ``None`` value.
+# Django's own docs recommend this pattern; we use it inside the panel
+# so that ``cache.get(key, sentinel)`` returns the sentinel on miss,
+# letting the panel count hits/misses correctly.  (#2417)
+_CACHE_MISS_SENTINEL = object()
+
+
 def _monkey_patch_method(cache, name, alias):
     original_method = getattr(cache, name)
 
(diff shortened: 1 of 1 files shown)


tim-schilling:
This approach won't work for the toolbar as discussed in https://github.com/django-commons/django-debug-toolbar/pull/2418. Please see the referenced issue to learn about the current state of the issue and where things are at. I believe someone is working on it already.

Harsh3006:
Hi @aryansk, I've been working on the same issue, and my initial approach was also to use a sentinel. Tim raised an important concern around modifying the arguments passed to the wrapped cache call, which led us to explore the trade-offs and alternative approaches.
 
We've since discussed a best-effort heuristic that improves some of the current cases while documenting the remaining ambiguity, as well as the possibility of more accurate tracking as an opt-in feature if there's enough interest.
 
I've started a [GitHub Discussion](https://github.com/django-commons/django-debug-toolbar/discussions/2441) to gather feedback on this. Please have a look there and share your thoughts if you have any ideas or use cases that could help shape the approach.

aryansk:
Thanks @tim-schilling — understood, and apologies for missing the #2418 discussion. Given @Harsh3006 is actively working the agreed approach on this issue, I'll close my draft to keep the thread clean.

PR #2454, opened 2026-08-28
Stop a JSON key named items breaking the History panel

### LINKED ISSUE #2453: HistoryPanel crashes when AJAX JSON response has items as one of the fields

aviseth:
Fixes #2453.

`history_tr.html` iterates the recorded request body with `{% for key, value in history_context.history_stats.data.items %}`. Django resolves a template variable with a dictionary lookup *before* it tries attribute lookup, so when the body is JSON containing a key named `items`, `data.items` returns that key's value instead of the mapping's `items()` method. The panel then iterates whatever that value happens to be.

The reporter described the crash. Running the current template against a few bodies, it is a bit worse than that — one case is silently wrong rather than loud:

```
{'foo': 'bar'}           -> 'foo=bar;'                  ok
{'items': 'a string'}    -> ValueError: Need 2 values to unpack in for loop; got 1.
{'items': [1, 2, 3]}     -> ValueError: Need 2 values to unpack in for loop; got 1.
{'items': {'a': 1}}      -> ValueError: Need 2 values to unpack in for loop; got 1.
{'items': [['k', 'v']]}  -> 'k=v;'                      silently renders the wrong table
```

The last one is the reason I did not just guard against the exception: with a list of pairs there is no error to catch, and the panel presents the user's data as if it were the request's key/value mapping.

This adds a `dict_items` filter and uses it at that one call site. It calls `items()` in Python, where a key named `items` does not shadow the method, so the ambiguity disappears rather than being handled. Anything that is not a mapping yields no pairs, which also tidies up a second case: `generate_stats` does `json.loads(request.body)`, so `data` can be a list or a string, and `data.items` currently fails variable resolution silently. Those now render the existing `{% empty %}` branch.

I kept this to presentation only. Normalising in `generate_stats` instead would change the shape of the stored panel data, which the store serialises and `get_stats()["data"]` exposes, so it seemed the more invasive option for the same result — happy to go that way if you would prefer it.

The other templates using `.items` iterate headers, settings and context processors — all developer-controlled keys — so I left them alone rather than widening the diff.

Tests: a unit test for the filter, and a History panel test asserting the panel renders for `items` values that are a string, a list, a nested mapping, and a mapping where `items` sits alongside another key. Five of them fail on `main` with the `ValueError` above and pass here.

`ruff check` and `ruff format --check` are clean. The full suite is unchanged: 5 errors before and after, all in `test_csp_rendering` and the example app, and identically present on an unmodified checkout — they are a gap in my local dependency install, not related to this change.

### DIFF
--- a/debug_toolbar/templates/debug_toolbar/panels/history_tr.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/history_tr.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load i18n debug_toolbar %}
 <tr class="{% if request_id == current_request_id %}djdt-highlighted{% endif %}" id="historyMain_{{ request_id }}" data-request-id="{{ request_id }}">
     <td>
         {{ history_context.history_stats.time|escape }}
@@ -24,7 +24,7 @@
               </tr>
             </thead>
             <tbody>
-              {% for key, value in history_context.history_stats.data.items %}
+              {% for key, value in history_context.history_stats.data|dict_items %}
                 <tr>
                   <td><code>{{ key|pprint }}</code></td>
                   <td><code>{{ value|pprint }}</code></td>
--- a/debug_toolbar/templates/debug_toolbar/panels/history_tr.html
+++ b/debug_toolbar/templates/debug_toolbar/panels/history_tr.html
@@ -1,4 +1,4 @@
-{% load i18n debug_toolbar %}
+{% load debug_toolbar i18n %}
 <tr class="{% if request_id == current_request_id %}djdt-highlighted{% endif %}" id="historyMain_{{ request_id }}" data-request-id="{{ request_id }}">
     <td>
         {{ history_context.history_stats.time|escape }}
--- a/debug_toolbar/templatetags/debug_toolbar.py
+++ b/debug_toolbar/templatetags/debug_toolbar.py
@@ -17,6 +17,29 @@ def _svg_contents(path):
         return handle.read().strip()
 
 
+@register.filter
+def dict_items(value):
+    """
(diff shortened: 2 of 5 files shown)


tim-schilling:
@aviseth this PR is a bit frustrating to receive and I suspect that's not your intention. We have a PR template that we'd like people to use and either your agent or you ignored it entirely. Plus the PR body has text that's just typical verbose AI output that's not helpful for us and is just extra load.

> Tests: a unit test for the filter, and a History panel test asserting the panel renders for items values that are a string, a list, a nested mapping, and a mapping where items sits alongside another key. Five of them fail on main with the ValueError above and pass here.
>
> ruff check and ruff format --check are clean. The full suite is unchanged: 5 errors before and after, all in test_csp_rendering and the example app, and identically present on an unmodified checkout — they are a gap in my local dependency install, not related to this change.

aviseth:
@tim-schilling - don't see the template anywhere, sorry!

tim-schilling:
@aviseth it's part of GitHub's PR request flow, it happens automatically when a PR is created: https://github.com/django-commons/django-debug-toolbar/blob/main/.github/PULL_REQUEST_TEMPLATE.md

aviseth:
@tim-schilling I opened this with gh pr create --body-file, which silently ignores the template (it only prefills in interactive mode), so I never saw the checklist. If you want that checkbox to be un-skippable, the template alone can’t do it as gh and the API both bypass it.

tim-schilling:
Thank you for that information. I didn't close your PR. I was sharing my frustration with this particular interaction.

aviseth:
Oh yeah no worries, I closed the PR - will look into the template and
reopen :)

Regards
Avi Seth
(he/him/his)

PR #2461, opened 2026-09-01
Document and test the USE_SHADOW_DOM setting

alissaahodge:
#### Description

`USE_SHADOW_DOM` has been supported since 7.0.0 and is mentioned in that release's notes, but it never made it into `docs/configuration.rst` — it's the only key in `CONFIG_DEFAULTS` without an entry there, and it had no test coverage.

This adds the docs entry under Toolbar options, after `UPDATE_ON_FETCH`. The wording about custom panels relying on external styles or DOM lookups comes from the 7.0.0 changelog, and it points at the existing Theming support note rather than repeating the CSS variable change.

It also adds two integration tests, since the setting changes rendered output and nothing was asserting on it. I checked they aren't vacuous by flipping the default to `False` and confirming `test_shadow_dom_enabled_by_default` fails.

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -28,6 +28,8 @@ Pending
   in its title bar.
 * Fixed the History panel's Refresh button submitting its form when clicked
   before the panel's script had loaded.
+* Documented the ``USE_SHADOW_DOM`` setting, which was added in 7.0.0 but was
+  missing from the configuration documentation, and added test coverage for it.
 
 7.1.1 (2026-08-14)
 ------------------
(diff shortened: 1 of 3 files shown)


tim-schilling:
Thanks @alissaahodge this is a good catch. As a note, we have a PR template that coding agents tend to bypass. When contributing, it's generally polite to see how a project accepts PRs.

federicobond (approved):

alissaahodge:
@tim-schilling That's a fair callout! I updated it with the template, thanks for pointing that out. Definitely should have used the template to indicate AI assistance rather than just including it in the commit message.

tim-schilling:
I stripped the cursor co-author. Thanks for the adjustments!

PR #2462, opened 2026-09-02
Shrink the toolbar panel icons to 1.1em

federicobond:
#### Description

I think the icons sit better beside their labels in a smaller size, and don't draw so much attention to themselves.

| Before | After |
|---|---|
| Captura de pantalla 2026-09-02 a la(s) 12 24 52 p  m | Captura de pantalla 2026-09-02 a la(s) 12 24 47 p  m |

#### Checklist:

#### AI/LLM Usage

### DIFF
--- a/debug_toolbar/static/debug_toolbar/css/toolbar.css
+++ b/debug_toolbar/static/debug_toolbar/css/toolbar.css
@@ -306,15 +306,15 @@
 }
 
 #djDebug #djDebugToolbar li .djdt-nav-icon {
-    width: 1.4em;
-    height: 1.4em;
+    width: 1.1em;
+    height: 1.1em;
     margin-inline-end: 3px;
     vertical-align: text-bottom;
 }
 
 #djDebug #djDebugToolbar li#djdt-VersionsPanel .djdt-nav-icon {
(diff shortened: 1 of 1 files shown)


federicobond:
@VeldaKiara what do you think?

VeldaKiara:
The toolbar row height/spacing was adjusted to accommodate the large icons, so if we are reducing the icon size, I think we should revisit that spacing as well.

It would also be worth checking panels that have text below the label, such as the Versions panel showing the Django 6.0 version, to make sure the alignment and spacing still look right.

You could eyeball the adjustments and then share a screenshot afterwards so we can see what it looks like now.

federicobond:
I've included before/after screenshots in the PR description, which show panels that have text below the label. The current spacings look ok to me, but if you prefer I can increase the vertical padding for each item a few pixels?

VeldaKiara:
No, we can proceed with how it looks. When you eyeball the screens you shared, they seem to show the same height, which is why I commented that it would need to be adjusted as well.
I checked the code edits you made, and they change it.

This is good to go.

matthiask:
Thanks to both of you!

PR #2467, opened 2026-09-07
Fix HistoryPanel crash when JSON body has an items key

### LINKED ISSUE #2453: HistoryPanel crashes when AJAX JSON response has items as one of the fields

arbazkhan971:
#### Description

Django template variable resolution looks up dictionary keys before methods. The History panel template iterated `history_stats.data.items`, so a captured JSON body or `QueryDict` with a key named `items` returned that value instead of `dict.items()`. The `{% for key, value in ... %}` then crashed (or silently rendered the wrong table).

This records a Python-built `data_items` list of pairs in `HistoryPanel.generate_stats` (`list(data.items())` when `data` has `.items()`, else `[]`) and iterates that list in `history_tr.html`. The original `data` mapping is still stored, so existing tests that read `get_stats()["data"]` keep working.

I commented on #2453 first with this plan. #2454 previously attempted a `dict_items` template filter for the same bug but was closed without merge because it did not use this project's PR template.

Fixes #2453

#### Checklist:

#### AI/LLM Usage

Made with [Cursor](https://cursor.com)


debug_toolbar/panels/history/panel.py
```diff
@@ -89,12 +89,16 @@ def generate_stats(self, request: HttpRequest, response: HttpResponse) -> None:
             # It is not guaranteed that we may read the request data (again).
             pass
 
+        # Build pairs in Python so a key named "items" cannot shadow
+        # dict.items() when the History panel template iterates the data.
+        data_items = list(data.items()) if hasattr(data, "items") else []
```
tim-schilling: Doesn't `hasattr(data, "items")` always resolve to `True`?

debug_toolbar/templates/debug_toolbar/panels/history_tr.html
```diff
@@ -24,7 +24,7 @@
               </tr>
             </thead>
             <tbody>
-              {% for key, value in history_context.history_stats.data.items %}
+              {% for key, value in history_context.history_stats.data_items %}
```
tim-schilling, line 27: I think the better approach here is a custom template tag that iterates over the dictionary.

tests/panels/test_history.py
```diff
@@ -48,6 +48,37 @@ def test_post_json(self):
                 data = self.panel.get_stats()["data"]
                 self.assertDictEqual(data, expected_stats_data)
 
+    def test_post_json_items_key_does_not_crash_history_render(self):
+        """A JSON key named items must not shadow dict.items() in history_tr.html."""
+        self.request = rf.post(
+            "/",
+            data={"items": "should-not-shadow", "foo": "bar"},
+            content_type="application/json",
+            CONTENT_TYPE="application/json",  # Force django test client to add the content-type even if no data
```
tim-schilling, line 57: This seems suspect. Why do we need both?

tests/panels/test_history.py
```diff
@@ -209,3 +240,31 @@ def test_history_refresh(self):
 
         for val in ["spam", "eggs"]:
             self.assertIn(val, data["requests"][1]["content"])
+
+    def test_history_refresh_json_items_key(self):
```
tim-schilling, line 244: These tests look really similar. I think we can consolidate them.

tests/panels/test_history.py
```diff
@@ -48,6 +48,37 @@ def test_post_json(self):
                 data = self.panel.get_stats()["data"]
                 self.assertDictEqual(data, expected_stats_data)
 
+    def test_post_json_items_key_does_not_crash_history_render(self):
+        """A JSON key named items must not shadow dict.items() in history_tr.html."""
```
tim-schilling, line 52: Let's go ahead and mention that this is a regression with the github issue link. We've done it in a few other places in the code base and this is odd enough to call out I think.

tim-schilling (changes requested):
Thank you for the PR. I think there's a better approach that's a bit cleaner though with a template tag.

PR #2469, opened 2026-09-08
Stop a JSON key named items breaking the History panel

### LINKED ISSUE #2453: HistoryPanel crashes when AJAX JSON response has items as one of the fields

aviseth:
#### Description

Resubmission of #2454 with the template filled in, as @matthiask suggested in #2453. Rebased on current `main`.

`history_tr.html` iterates the recorded request body with `{% for key, value in history_context.history_stats.data.items %}`. Django resolves a template variable with a dictionary lookup before it tries attribute lookup, so when the body is JSON containing a key named `items`, `data.items` returns that key's value instead of the mapping's `items()` method, and the loop iterates whatever that value happens to be.

Most shapes raise `ValueError: Need 2 values to unpack in for loop`. One is worse: `{"items": [["k", "v"]]}` renders without any error and presents the user's data as if it were the request's key/value mapping. That silent case is why this adds a filter rather than guarding against the exception.

The new `dict_items` filter calls `items()` in Python, where a key named `items` cannot shadow the method. Anything that is not a mapping yields no pairs, which also tidies up a second case: `generate_stats` does `json.loads(request.body)`, so `data` can be a list or a string, and `data.items` currently fails variable resolution silently. Those now render the existing `{% empty %}` branch.

I kept the change to presentation. Normalising in `generate_stats` would change the shape of the stored panel data that the store serialises and `get_stats()["data"]` exposes, so it looked like the more invasive option for the same result — happy to go that way if you would prefer it. The other templates using `.items` iterate headers, settings and context processors, all developer-controlled keys, so I left them alone.

@arbazkhan971 said they would take this on #2453 — apologies for the overlap, this is the existing patch rather than a new one, and I am happy to step back if you have something in flight.

Fixes #2453.

#### Checklist:

#### AI/LLM Usage


tim-schilling:
Closing as duplicate of #2647. We'd appreciate your review on the other PR to improve the approach, but this issue was already being worked on by someone.

PR #2470, opened 2026-09-10
fix: strip whitespace in is_processable_html_response and add unit tests

aoright:
### Summary

- In `is_processable_html_response`, strip whitespace around `content_type` and `content_encoding` so that headers like `text/html ; charset=utf-8` (with trailing space before the delimiter) are properly recognized as valid processable HTML responses.
- Fix minor grammar typo in `get_name_from_obj` docstring (`a object` -> `an object`).
- Add `IsProcessableHtmlResponseTestCase` in `tests/test_utils.py` to cover `is_processable_html_response` across various content types, character sets with whitespace, encodings, and streaming responses.


tim-schilling:
@aoright can you please explain what bug this is fixing? We also have a PR template that we'd like people to use.

tim-schilling:
This is cleaning something that the toolbar doesn't manipulate or manage. It doesn't make sense to me for the toolbar to handle this. You're welcome to create an issue to discuss this further.

PR #2471, opened 2026-09-10
Load toolbar fonts from the light DOM

### LINKED ISSUE #2466: Fonts are not loaded
It seems that the fonts don't load properly in the shadow DOM. Despite the main font being specified as "Geist", I'm seeing the system default UI font shown:

Image

Seems like this might be a known problem with some browser engines (see e.g. https://issues.chromium.org/issues/41085401 and https://github.com/mdn/interactive-examples/issues/887 for related discussion), nonetheless, it should probably be fixed.

gyanu2507:
#### Description

Fixes #2466.

`@font-face` for Geist and Alef lived in `toolbar.css`, which loads inside the open shadow tree. Chrome does not register those faces, so the toolbar fell back to the system UI font. Safari already worked.

`fonts.css` is linked on `#djDebugRoot` before the `<template shadowrootmode="open">`, so the faces are document-scoped. `toolbar.css` still sets `font-family` inside the shadow tree.

This does not add a Geist TTF. That was a separate request on the issue.

#### Checklist:

- [x] I have added an item to the Pending section of `docs/changes.rst`.

#### AI/LLM Usage


debug_toolbar/static/debug_toolbar/css/fonts.css
```diff
@@ -0,0 +1,18 @@
+/* @font-face must live in the light DOM. Chrome ignores it inside a shadow tree. */
```
tim-schilling:
```suggestion
/* @font-face must live in the regular DOM. Chrome ignores it inside a shadow DOM. */
```
I honestly thought light DOM was something light-weight, not light as in the opposite of shadow. While it looks like light is used elsewhere, maybe we can use "regular" instead, the [same as MDN](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_shadow_DOM#:~:text=regular%20DOM,-node)?

debug_toolbar/static/debug_toolbar/css/fonts.css
```diff
@@ -0,0 +1,18 @@
...
+@font-face {
+    font-family: "Alef";
```
tim-schilling:
```suggestion
    font-family: "djdt-Alef";
```

Let's prefix these two font family values so that we don't collide with an applications own definition and change the other styles.

We'll also have to track down where Alef and Geist are used to prefix them as well.

tim-schilling (commented):
Thanks @gyanu2507 for working on this. I've got some real minor nitpicks.

debug_toolbar/static/debug_toolbar/css/fonts.css
gyanu2507: Updated both comments to say regular DOM / shadow DOM, matching the MDN wording.

debug_toolbar/static/debug_toolbar/css/fonts.css
gyanu2507: Prefixed both faces: djdt-Alef for panel titles and djdt-Geist for --djdt-font-family-primary. File URLs still point at Alef-Bold / Geist.woff2.

tim-schilling:
@gyanu2507 did you want to add a change to the changes.rst file? Also, we have a PR template that would guide you through these things.

gyanu2507:
Added a Pending line in `docs/changes.rst` and filled in the PR template.

The changelog entry is the Chrome `@font-face` / shadow-tree fix. The font-link test was already on the branch.

tim-schilling (approved):
Thank you @gyanu2507

